PR #1, opened 2025-07-21
Add initial examples

ElliottKasoar:
Adds example benchmarks for OC157 and Li diffusion barriers (NEB).

To do:

- [x] Fix callback for Li diffusion atom visualisation
- [x] Fix sorting/order of models/storage of data (use dictionary rather than list)
- [x] Add benchmark with multiple tables


mlip_testing/analysis/surfaces/S24/analyse_S24.py
```diff
@@ -0,0 +1,188 @@
...
+    Returns
+    -------
+    dict[str, list]
+        Dictionary of all reference and predicted adsorption energies.
+    """
+    results = {"ref": []} | {mlip: [] for mlip in MODELS}
+    ref_stored = False
+
+    for model_name in MODELS:
+        for i, system_path in enumerate(sorted((CALC_PATH / model_name).glob("*.xyz"))):
```
ElliottKasoar, line 91:
```suggestion
        for system_path in sorted((CALC_PATH / model_name).glob("*.xyz")):
```
I don't think `i` is used so it (and the enumerate) can probably be removed?

Sorting is a good idea though, we should probably do this everywhere, as the glob order isnt' guaranteed.

mlip_testing/calcs/surfaces/cat_surfaces.py
```diff
@@ -0,0 +1,81 @@
...
+            s24 = S24Benchmark(
+                model=model,
+                model_name=model_name,
+            )
+            oc157_benchmarks.append(oc157)
+            s24_benchmarks.append(s24)
+
+    # Create surface category
+    with project.group("surface"):
+        surface_category = SurfaceCategory(
```
ElliottKasoar, line 63:
```suggestion
        SurfaceCategory(
```

mlip_testing/calcs/surfaces/cat_surfaces.py
```diff
@@ -0,0 +1,81 @@
...
+    Category node for surface-related benchmarks.
+
+    Aggregates OC157 and S24 benchmarks into a single category.
+    """
+
+    oc157_benchmarks: list[OC157Benchmark] = zntrack.deps()
+    s24_benchmarks: list[S24Benchmark] = zntrack.deps()
+
+    def run(self):
+        """Categories don't run calculations - they just organize the DAG."""
```
ElliottKasoar, line 24:
```suggestion
        """Organise the DAG (categories don't run calculations)."""
```
For docstring consistency ("Summary must start with infinitive verb")

mlip_testing/calcs/surfaces/cat_surfaces.py
```diff
@@ -0,0 +1,81 @@
+"""Surface benchmark category node."""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+import zntrack
+
+from mlip_testing.calcs.OC157.calc_OC157 import OC157Benchmark
+from mlip_testing.calcs.S24.calc_S24 import S24Benchmark
```
ElliottKasoar, line 10:
```suggestion
from mlip_testing.calcs.surfaces.OC157.calc_OC157 import OC157Benchmark
from mlip_testing.calcs.surfaces.S24.calc_S24 import S24Benchmark
```

mlip_testing/calcs/surfaces/cat_surfaces.py
ElliottKasoar, line 1: I think we want to use the `chdir` function I set up here.

We want to be able to run this script from anywhere, but that it currently fails in most places since there's no `dvc.lock` file. (`chdir` is a context manager to temporarily change directory, so we can change to the directory which contains this, then change back after running)

ElliottKasoar, line 1: We also need to add guidance on setting up dvc within a git-tracked repo.

I think we can just run `dvc init --subdir` (it failed without the `--subdir` for me)? There may be other/better ways though

.gitignore
```diff
@@ -1,5 +1,8 @@
+*.json
+*.html
 *.dat
 *.lock
+!dvc.lock
```
ElliottKasoar, line 5: Do we need this? If I delete it OC157 etc. it doesn't seem to make a difference to running things?

ThomasWarford:
Hi, 

the bug I noticed was that points in the parity plots for S24 are given the wrong labels

The following XYZ (found at `mlip_testing/app/data/surfaces/S24/mace_mp_0b3/C48H24O26Zn8-H2..csv`) has ref adsorption energy of -0.1.

NOTE: I save it as csv because GitHub won't let me upload xyz, change the file extension if you want.
image

I fixed this on my branch by changing the `get_system_names` function in `mlip_testing/analysis/surfaces/S24/analyse_S24.py`

```
def system_names() -> list:
    """
    Get list of system names.

    Returns
    -------
    list
        List of all system names.
    """
    # NOTE: The parity plot associates hover data with points in the
    # order the data arrays are provided. In `adsorption_energies` the
    # systems are iterated with `sorted((CALC_PATH / model_name).glob("*.xyz"))`.
    # Originally this function used the unsorted `.glob()` ordering,
    # which can differ from the sorted order, leading to mismatched
    # (mislabelled) hover labels (system names) in the output JSON.
    # We therefore reproduce the exact ordering used in the data
    # collection step by applying the same `sorted(...)` here.
    system_names_list = []
    for model_name in MODELS:
        for system_path in sorted((CALC_PATH / model_name).glob("*.xyz")):
            system_names_list.append(system_path.stem)
        break  # only need the first model to list available systems
    return system_names_list
    ```

ThomasWarford:
I have not confirmed this fix for S24, but I did for the branch I made, which copied this code, maybe it should be moved into another file without the global variable dependence

PR #33, opened 2025-09-12
Update PR templates

### LINKED ISSUE #32: Add documentation tickbox for new model/benchmark PR tempaltes
### Summary

I overlooked this since we need to do it in #6, but we should check that people properly document their contributions.

### Example

_No response_

### Alternatives

_No response_

### Additional context

_No response_

ElliottKasoar:

- [x] I've confirmed the [contribution guidelines](https://github.com/ddmms/mlip-testing/blob/main/contributing.md).

## Summary

Adds checkbox for documentation when adding models and benchmarks, and other minor fixes.

## Linked issue

Resolves #32

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/new_model.md
+++ b/.github/PULL_REQUEST_TEMPLATE/new_model.md
@@ -15,6 +15,7 @@ PR author **must** check the checkboxes below when creating the PR.
 ## Summary
 
 <!-- Describe the new model. This can be brief, as most information can be in the linked issue. -->
+
 ## Model Reference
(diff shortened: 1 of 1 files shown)


alinelena (dismissed):

joehart2001 (commented):
add reference to the added model, maybe also add instrucutions on how to run the model or default params? e.g. float64

ElliottKasoar:
> add reference to the added model, maybe also add instrucutions on how to run the model or default params? e.g. float64

Thanks. Maybe let's see how it works in practice. I think a lot of that can be defined in the config itself, but we'll need to revisit this once we start actually accepting new models.

joehart2001 (approved):

PR #57, opened 2025-09-19
Add category name and simplify test names

### LINKED ISSUE #43: Set tab/category name
### Summary

Currently in #1, the tab name is simply the directory name, but it should be settable somewhere.

### Example

_No response_

### Alternatives

_No response_

### Additional context

_No response_

ElliottKasoar:

- [x] I've confirmed the [contribution guidelines](https://github.com/ddmms/mlip-testing/blob/main/contributing.md).

## Summary

Adds yaml files for category title, used for the tab, title, etc., and description.

Also combined individual test names/titles into a single name parameter, as we don't need it to be extracted from the file name, and this allows more human-readable text.

## Linked issue

Resolves #43

## Testing

None

Could use testing from @joehart2001, and improvements to the category descriptions are welcome!

### DIFF
--- a/ml_peg/app/build_app.py
+++ b/ml_peg/app/build_app.py
@@ -103,10 +103,14 @@ def build_category(
 
     for category in all_layouts:
         # Get category name and description
-        with open(APP_ROOT / category / f"{category}.yml") as file:
-            category_info = safe_load(file)
-            category_title = category_info["title"]
-            category_descrip = category_info["description"]
+        try:
+            with open(APP_ROOT / category / f"{category}.yml") as file:
+                category_info = safe_load(file)
+                category_title = category_info.get("title", category)
+                category_descrip = category_info.get("description", "")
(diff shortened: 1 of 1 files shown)


ElliottKasoar:
> Looks good, will give it a test. I'm also thinking we should add an extended description which is collapsable (think we've discussed briefly before). can add it to the yaml file

Thanks!

And/or link to the documentation where we describe the category in detail (in addition to each test, as we have at the moment).

joehart2001:
> > Looks good, will give it a test. I'm also thinking we should add an extended description which is collapsable (think we've discussed briefly before). can add it to the yaml file
> 
> Thanks!
> 
> And/or link to the documentation where we describe the category in detail (in addition to each test, as we have at the moment).

yeah i think would be good to have both, as people may not want to open loads of links and just have a quick glance. will barely take up any space when colapsed and should be simple to add

PR #75, opened 2025-10-02
add metric weights ability and implement for current benchmarks

### LINKED ISSUE #65: Add weights to individual test tables
### Summary

In addition to the summary and category weights, we will add weights for all tables.

### Example

_No response_

### Alternatives

_No response_

### Additional context

_No response_

joehart2001:

## Summary

Implemented weights for metrics within each benchmark, which updates the benchmark score and propagates through to the category and overall scores

## Linked issue

Resolves #65 

## Testing

 launching the app and checking for bugs

### DIFF
--- a/ml_peg/app/build_app.py
+++ b/ml_peg/app/build_app.py
@@ -126,9 +126,7 @@ def build_category(
         # Build weight components for category summary table
         weight_components = build_weight_components(
             header="Benchmark weights",
-            columns=list(all_tables[category].keys()),
-            input_ids=list(all_tables[category].keys()),
-            table_id=f"{category_title}-summary-table",
+            table=summary_table,
         )
 
         # Build full layout with summary table, weight controls, and test layouts
@@ -282,9 +280,7 @@ def build_full_app(full_app: Dash) -> None:
     summary_table = build_summary_table(category_tables)
     weight_components = build_weight_components(
         header="Benchmark weights",
-        columns=list(category_tables.keys()),
-        input_ids=list(category_tables.keys()),
-        table_id="summary-table",
+        table=summary_table,
     )
     # Build summary and category tabs
     build_tabs(full_app, category_layouts, summary_table, weight_components)
--- a/ml_peg/app/utils/build_components.py
+++ b/ml_peg/app/utils/build_components.py
@@ -71,9 +71,7 @@ def build_slider(
 
 def build_weight_components(
     header: str,
-    columns: list[str],
-    input_ids: list[str],
-    table_id: str,
+    table: DataTable,
 ) -> Div:
     """
     Build weight sliders, text boxes and reset button.
@@ -82,12 +80,8 @@ def build_weight_components(
     ----------
     header
         Header for above sliders.
-    columns
-        Column headers to look up stored values, and label sliders and input boxes.
-    input_ids
-        ID prefixes for sliders and input boxes.
-    table_id
-        ID for table. Also used to identify reset button and weight store.
+    table
+        DataTable to build weight components for.
 
     Returns
     -------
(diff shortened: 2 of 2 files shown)


ml_peg/app/utils/build_components.py
```diff
@@ -205,3 +205,57 @@ def build_test_layout(
...
+    reserved = {"MLIP", "Score", "Rank", "id"}
+    metric_columns = [
+        col["id"] for col in table.columns if col.get("id") not in reserved
+    ]
+
+    if not metric_columns:
+        return Div()
+
+    # Use table id to generate unique input ids
+    def _safe(col: str) -> str:
```
ElliottKasoar, line 238: Is this actually an issue? We have a bunch of IDs with spaces at this point I think. If it can be a problem, we may want to use this more widely.

ml_peg/app/utils/build_components.py
```diff
@@ -205,3 +205,57 @@ def build_test_layout(
         layout_contents.extend(extra_components)
 
     return Div(layout_contents)
+
+
+def build_metric_weight_components(
```
ElliottKasoar, line 210: I'm a little confused by this function. It looks sensible, but I'm also not sure if it doesn't anything that we don't need for the general weight building, which it calls at the end? Can/should these be merged, or am I missing something?

ml_peg/app/utils/register_callbacks.py
```diff
@@ -138,6 +141,84 @@ def update_scores_store(
         return scores_data
 
 
+def register_benchmark_to_category_callback(
```
ElliottKasoar, line 144: I need to look at this more carefully, but my gut question is to check why this needs to be different to the existing callback we have?

In either case, we have a list of tables (from categories, or from benchmarks in a category), and these are used to populate a summary table.

Is anything different because they're all in the same tab?

ml_peg/app/build_app.py
```diff
@@ -137,6 +142,21 @@ def build_category(
...
+        for test_name, benchmark_table in all_tables[category].items():
+            try:
+                register_benchmark_to_category_callback(
+                    benchmark_table_id=benchmark_table.id,
+                    category_table_id=f"{category_title}-summary-table",
+                    benchmark_column=test_name,
+                )
+            except Exception as err:
+                msg = (
+                    f"Unable to wire benchmark table '{test_name}' to category "
```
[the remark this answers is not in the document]
joehart2001:
```suggestion
        # Register benchmark table -> category table callbacks
        for test_name, benchmark_table in all_tables[category].items():
            register_benchmark_to_category_callback(
                benchmark_table_id=benchmark_table.id,
                category_table_id=f"{category_title}-summary-table",
                benchmark_column=test_name,
            )
```

ml_peg/app/build_app.py
```diff
@@ -137,6 +142,21 @@ def build_category(
...
 
+        # Wire child test tables to update this category summary table
+        for test_name, benchmark_table in all_tables[category].items():
+            try:
+                register_benchmark_to_category_callback(
+                    benchmark_table_id=benchmark_table.id,
+                    category_table_id=f"{category_title}-summary-table",
+                    benchmark_column=test_name,
+                )
+            except Exception as err:
```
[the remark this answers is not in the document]
joehart2001: was meant to clean this up, don't think theres any reason to have a try except

ml_peg/app/base_app.py
```diff
@@ -74,12 +77,18 @@ def build_layout(self) -> Div:
             Div component with list all components for app.
         """
         # Define all components/placeholders
+        # Auto-append metric-weight controls for this benchmark table
+        metric_weights = build_metric_weight_components(self.table)
+        extra_components = [metric_weights]
+        if self.extra_components:
+            extra_components.extend(self.extra_components)
```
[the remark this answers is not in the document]
joehart2001: thats a good spot, i've moved it as we want it for every benchmark anyway

ml_peg/app/utils/build_components.py
joehart2001: agreed

ml_peg/app/utils/build_components.py
joehart2001: lets discuss this? i guess we could add this to build_weight_components but i think its useful to break up the code

Issue #12, opened 2025-08-20
Add option to swap between MAE, RMSE, etc.


ElliottKasoar:
May make use of Table `hidden_columns` feature?

For example, drop down for error type -> change hidden columns, and update scores.

joehart2001:
sounds like a great idea!

ElliottKasoar:
We also need to think about how we generate the different error values.

The simplest option would be to create multiple functions each time we have an error, and simply treat them as independent results, but it should be something we can do in a completely general way.

PR #84, opened 2025-10-17
Add score normalisation

### LINKED ISSUE #37: Add normalisation
### Summary

Our current live benchmarks are normalised to mace_mp_0a, but this is not currently implemented here.

I think we also need to consider how scores are combined in a way that makes more sense, but this may be helped a little by the thresholds from #4, as perhaps they could also control a maximum score?

### Example

See: http://mlip-testing.stfc.ac.uk:8050

### Alternatives

_No response_

### Additional context

_No response_

ElliottKasoar:

## Summary

Draft changes (primarily from @joehart2001) to add normalisation to combine scores in a unitless way. 

## Linked issue

Resolves #37

## Testing

TBC

### DIFF
--- a/ml_peg/app/build_app.py
+++ b/ml_peg/app/build_app.py
@@ -15,7 +15,7 @@ from ml_peg.analysis.utils.utils import calc_ranks, calc_table_scores, get_table
 from ml_peg.app import APP_ROOT
 from ml_peg.app.utils.build_components import build_weight_components
 from ml_peg.app.utils.register_callbacks import register_benchmark_to_category_callback
-from ml_peg.app.utils.utils import calculate_column_widths
+from ml_peg.app.utils.utils import calculate_column_widths, rank_format, sig_fig_format
 from ml_peg.models.get_models import get_model_names
 from ml_peg.models.models import current_models
 
@@ -197,7 +197,16 @@ def build_summary_table(
     data = calc_ranks(data)
 
     columns_headers = ("MLIP",) + tuple(tables.keys()) + ("Score", "Rank")
+
     columns = [{"name": headers, "id": headers} for headers in columns_headers]
+    for column in columns:
+        column_id = column["id"]
+        if column_id == "Rank":
+            column["type"] = "numeric"
+            column["format"] = rank_format()
+        elif column_id != "MLIP":
+            column["type"] = "numeric"
+            column["format"] = sig_fig_format()
 
     style = get_table_style(data)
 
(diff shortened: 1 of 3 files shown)


ml_peg/analysis/utils/decorators.py
```diff
@@ -356,3 +356,189 @@ def build_table_wrapper(*args, **kwargs) -> dict[str, Any]:
...
+            from ml_peg.analysis.utils import utils as _utils
+            from ml_peg.analysis.utils.utils import calc_ranks
+
+            if normalizer is None:
+                _normalizer = _utils.normalize_metric
+            else:
+                _normalizer = normalizer
+
+            # Local scorer using the chosen normalizer
+            def _calc_normalized_scores(rows, ranges):
```
ElliottKasoar: Is this not something we want as a general utility function that will be called later?

ElliottKasoar: These could use type hints, especially ranges

ml_peg/analysis/utils/utils.py
```diff
@@ -178,3 +178,119 @@ def rgba_from_val(val: float, vmin: float, vmax: float, cmap: Colormap) -> str:
...
+        return 1.0 if value == x_threshold else 0.0
+
+    # Linear map: Y -> 0, X -> 1
+    t = (value - y_threshold) / (x_threshold - y_threshold)
+    # Clip to [0, 1]
+    if t < 0.0:
+        return 0.0
+    if t > 1.0:
+        return 1.0
+    return float(t)
```
ElliottKasoar: This could probably be condensed e.g.  `max(min(1, t), 0)` (check!)

ml_peg/analysis/utils/utils.py
```diff
@@ -178,3 +178,119 @@ def rgba_from_val(val: float, vmin: float, vmax: float, cmap: Colormap) -> str:
...
+        for key, value in row.items():
+            if (
+                key not in ("MLIP", "Score", "Rank", "id")
+                and key in normalization_ranges
+            ):
+                try:
+                    x_threshold, y_threshold = normalization_ranges[key]
+                    x_threshold = float(x_threshold)
+                    y_threshold = float(y_threshold)
+                except (TypeError, ValueError):
```
ElliottKasoar: What are the `TypeError` and `ValueError`s that we're getting? Continuing on those doesn't sound good...

ElliottKasoar (commented):
The alignment stuff is a lot worse than I expected, so I haven't really touched it yet...

It is making me reconsider whether we can do any other way e.g. integrated into the tables still, which would also solve issues that we don't see here for #12, if we want to swap out columns etc.

PR #109, opened 2025-10-28
Add GMTKN55 benchmark

### LINKED ISSUE #81: Investigate adding dispersion correction for specific tests
### Summary

Currently, the expectation is that models with added D3 correction are defined in the models.yml file, but in reality, we may want to add it only for specific tests.

I believe this should be possible, since the models have a `add_d3` function, since #78.

We also want to consider other changes, such as setting the precision on-the-fly, but this may be more complex, as the model may already have been loaded, and how precision is set may vary more, and may not even be possible.

### LINKED ISSUE #97: GMTKN55
### Summary

Gas-phase chemical accuracy for various chemical domains:

- Basic Properties
- Reaction Energies
- Barrier Heights
- Intramolecular Noncovalent
- Intermolecular Noncovalent

### Interactive features

- Parity plot of energies
- Parity to structure

- Plots by category/subset?

### Category

Molecular

### Data availability

GMTKN55 dataset: 55 individual test sets totalling 1,505 relative energies

CCSD(T)/ωB97M-D3BJ/PBE

### Computational cost

Low-medium

### Additional context

_No response_

### LINKED ISSUE #106: Add default weights
### Summary

Similar to thresholds (#4), we should be able to set default weights for different metrics, rather than using 1 for everything.

### Example

This would be useful for #97, for example, since we may want to show the category WTMADs, but only use the combined WTMAD by default (weights 0 and 1 respectively).

ElliottKasoar:

## Summary

GMTKN55 benchmark for Molecular category, with weighted mean absolute deviations (WTMAD) for each category (weights set to 0) and a combined WTMAD.

WTMADs exclude those labelled "excluded", in addition to open shell and charged structures.

This also includes:

- Setting default weights for a metric table (**new feature**)
- Adding D3 dispersion for a specific test (**first implementation**)

## Linked issue

Resolves #97

## Progress

- [x] Calculations
- [x] Analysis
- [x] Application
- [x] Documentation

## Testing

Tested on several MACE models and Orb.

## New decorators/callbacks

None initially. We may want to consider ways to visualise different categories, and/or hide excluded data etc.


joehart2001 (approved):
all seems to work well on all models

PR #112, opened 2025-10-28
Add units to cols and thresholds

### LINKED ISSUE #111: Make units more explicit
### Summary

Unitful columns and good/bad thresholds should have units, so these are clear, rather than through tooltips.

It should also be clear in the analysis scripts that these thresholds are unitful, make sure the unit propagates through, and document this all.

### Example

_No response_

### Alternatives

_No response_

### Additional context

_No response_

joehart2001:

## Summary
Added units to table columns and good/bad thresholds. units are added alongside the specified default thresholds in the analyse script for each benchmark

## Linked issue

Resolves #111 

## Testing

mace models

### DIFF
--- a/ml_peg/analysis/molecular_crystal/DMC_ICE13/metrics.yml
+++ b/ml_peg/analysis/molecular_crystal/DMC_ICE13/metrics.yml
@@ -1,0 +1,7 @@
+metrics:
+  MAE:
+    good: 0.0
+    bad: 100.0
+    unit: meV
+    tooltip: "Mean Absolute Error for all polymorphs"
+    level_of_theory: PBE
--- a/ml_peg/analysis/molecular_crystal/X23/metrics.yml
+++ b/ml_peg/analysis/molecular_crystal/X23/metrics.yml
@@ -1,0 +1,7 @@
+metrics:
+  MAE:
+    good: 0.0
+    bad: 100.0
+    unit: kJ/mol
+    tooltip: "Mean Absolute Error for all systems"
+    level_of_theory: PBE
--- a/ml_peg/analysis/physicality/extensivity/metrics.yml
+++ b/ml_peg/analysis/physicality/extensivity/metrics.yml
@@ -1,0 +1,7 @@
+metrics:
+  ΔE:
+    good: 0.0
+    bad: 10.0
+    unit: meV
+    tooltip: "Absolute energy difference between isolated and combined slab systems"
+    level_of_theory: null
--- a/ml_peg/analysis/physicality/locality/metrics.yml
+++ b/ml_peg/analysis/physicality/locality/metrics.yml
@@ -1,0 +1,19 @@
+metrics:
+  Ghost atoms max ΔF:
+    good: 0.0
+    bad: 5.0
+    unit: meV/Å
+    tooltip: "aximum force difference on solute atoms due to ghost atoms"
+    level_of_theory: null
+  Random hydrogen mean ΔF:
+    good: 0.0
+    bad: 5.0
+    unit: meV/Å
+    tooltip: "Mean force difference on solute atoms due to random hydrogen atoms"
+    level_of_theory: null
+  Random hydrogen std ΔF:
+    good: 0.0
+    bad: 5.0
+    unit: meV/Å
+    tooltip: "Standard deviation of force difference on solute atoms due to random hydrogen atoms"
+    level_of_theory: null


ElliottKasoar:
Something else to consider is that it might be nice if the normalised version of the column doesn't show the units, both as a way to differentiate, and since it is unitless?

It might end up being complicated since it might change the column width if we're not careful though.

ml_peg/analysis/utils/decorators.py
```diff
@@ -251,7 +253,7 @@ def build_table(
     filename: str = "table.json",
     metric_tooltips: dict[str, str] | None = None,
     normalize: bool = True,
-    thresholds: dict[str, tuple[float, float]] | None = None,
+    thresholds: dict[str, Any] | None = None,
```
ElliottKasoar: Why would thresholds not be a tuple of floats?

ml_peg/analysis/utils/utils.py
```diff
@@ -3,17 +3,86 @@
...
+        raise TypeError(msg)
+    thresholds: dict[str, Any] = {}
+    tooltips: dict[str, str] = {}
+
+    for metric_name, metric_config in metrics_data.items():
+        if not isinstance(metric_config, dict):
+            msg = f"Metric configuration for '{metric_name}' must be a mapping."
+            raise TypeError(msg)
+
+        unit = metric_config.get("unit", metric_config.get("units"))
```
ElliottKasoar: It might be better to make s sure things are input consistently and raising errors for incorrect keys, rather than being too flexible?

It feels like the sort of thing that might get missed if we ever access the same data outside of this function.

ml_peg/analysis/utils/utils.py
```diff
@@ -3,17 +3,86 @@
...
+        level = metric_config.get("level_of_theory", metric_config.get("level"))
```
ElliottKasoar: Same as above. Maybe just `theory`? I don't think `level` is particularly clear on its own

ml_peg/analysis/utils/utils.py
```diff
@@ -3,17 +3,86 @@
...
+            msg = (
+                f"Metric '{metric_name}' in '{config_path}' must define a non-empty "
+                "'tooltip' entry."
+            )
+            raise ValueError(msg)
+        if level:
+            tooltip = f"{tooltip} (Level of theory: {level})"
+        tooltips[metric_name] = tooltip
+
+    extra_tooltips = data.get("extra_tooltips") or {}
```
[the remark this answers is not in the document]
joehart2001: something i tried initially but dont use in the end so removed

.gitignore
```diff
@@ -11,7 +11,7 @@
 *.cif
 *.xyz
 *.extxyz
-*.yml
+.yml
```
ElliottKasoar: Maybe we just remove it?

ml_peg/analysis/utils/utils.py
```diff
@@ -322,8 +395,8 @@ def update_score_rank_style(
     weights
         Weight for each metric. Default is `None`.
     thresholds
-        Normalisation thresholds keyed by metric name, where each value is
-        a (good_threshold, bad_threshold) tuple. Default is `None`.
+        Normalisation thresholds keyed by metric name. Each value may be a numeric
+        pair or a mapping with ``good``/``bad`` entries. Default is `None`.
```
ElliottKasoar: Also includes unit (unused)?

ml_peg/app/utils/utils.py
```diff
@@ -201,3 +212,170 @@ def get_scores(
...
+
+    Returns
+    -------
+    str
+        Tooltip string with updated unit descriptor.
+    """
+    if not text or not unit:
+        return text
+
+    pattern = re.compile(rf"\s*[\(\[]\s*{re.escape(unit)}\s*[\)\]]")
```
ElliottKasoar: It feels a little dangerous using the two different methods to extract the unit/metric name. We had `.split` previously. Perhaps we should pick one, and use it in both places?

ml_peg/app/utils/utils.py
```diff
@@ -201,3 +212,170 @@ def get_scores(
         return raw_rows
 
     return scored_rows
+
+
+def _base_column_label(column: dict[str, Any]) -> str:
```
ElliottKasoar: Not especially important, but in terms of consistency, is there a particular reason behind the `_` for some of these functions, but not others?

Generally it's useful to indicate private functions, but none of this is really user-facing, so I'm not sure it's necessary.

ml_peg/app/utils/load.py
```diff
@@ -57,11 +67,20 @@ def rebuild_table(filename: str | Path, id: str) -> DataTable:
         elif column.get("type") == "numeric" or is_numeric_column(data, column_id):
             column["type"] = "numeric"
             column.setdefault("format", sig_fig_format())
+
+        base_name = column_name if isinstance(column_name, str) else None
```
ElliottKasoar: In what situations are column_name/column_id not strings, or at minimum not able to be converted into strings (e.g. if for some reason they were numerical)?

Robustness is good, but maybe we should just raise an error, as otherwise the logic ends up more and more complex.

ml_peg/analysis/utils/utils.py
```diff
@@ -15,8 +15,12 @@
 
 from ml_peg.app.utils.utils import clean_weights
 
+ThresholdEntry = dict[str, float | str | None]
+Thresholds = dict[str, ThresholdEntry]
+MetricRow = dict[str, float | int | str | None]
```
ElliottKasoar: Not a bad idea. One thing to consider is `TypedDict` for these e.g. https://github.com/stfc/janus-core/blob/main/janus_core/helpers/janus_types.py#L171

Probably most appropriate for `ThresholdEntry`, since that has specific keys

ml_peg/analysis/utils/decorators.py
```diff
@@ -251,7 +251,7 @@ def build_table(
     filename: str = "table.json",
     metric_tooltips: dict[str, str] | None = None,
     normalize: bool = True,
-    thresholds: dict[str, tuple[float, float]] | None = None,
+    thresholds: dict[str, dict[str, Any]] | None = None,
```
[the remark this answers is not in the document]
ElliottKasoar: Can use `Thresholds` now?

ml_peg/app/utils/build_components.py
```diff
@@ -368,8 +405,9 @@ def build_test_layout(
     metric_weights = build_weight_components(
         header="Metric Weights",
         table=table,
-        use_thresholds=True,
+        use_thresholds=thresholds is not None,
```
ElliottKasoar: These are for the metric tables, so should always have thresholds, shouldn't they?

ml_peg/app/utils/build_components.py
```diff
@@ -18,6 +19,27 @@
...
+        Raw threshold data which may be a tuple, list, or dictionary.
+
+    Returns
+    -------
+    tuple[float | None, float | None, str | None]
+        ``(good, bad, unit)`` values where unit may be ``None``.
+    """
+    if not bounds:
+        raise KeyError("Threshold definition is required for each metric.")
+    return bounds["good"], bounds["bad"], bounds["unit"]
```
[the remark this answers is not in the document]
joehart2001: yep definitely is now, removed

ml_peg/analysis/utils/utils.py
joehart2001: ok nice, i wasnt familiar with them before but i think what i've added makes sense now. definitely a more clear way of structuring things

ml_peg/analysis/utils/utils.py
```diff
@@ -83,20 +146,30 @@ def calc_metric_scores(
...
+                entry = thresholds[key]
+                if not isinstance(entry, dict):
                     row[key] = value
+                    continue
+                try:
+                    good_threshold = float(entry["good"])
+                    bad_threshold = float(entry["bad"])
+                except (KeyError, TypeError, ValueError):
+                    row[key] = value
+                    continue
```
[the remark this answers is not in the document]
joehart2001: yep was meant to address last time

alinelena (approved):

PR #115, opened 2025-10-30
Add locality benchmark

### LINKED ISSUE #103: Locality
### Summary

"Ghost atoms" - adding hydrogens far away from an aluminium slab, which should produce net force of 0.

### Interactive features

- Forces at difference separations?
- Visualisation of structures

### Category

Physicality

### Data availability

No external data required

### Computational cost

_No response_

### Additional context

_No response_

ElliottKasoar:

## Summary

Adds first section of physicality tests from mlip-testing.stfc.ac.uk:8050, measuring the difference in force on acetone when  Ne atoms or a hydrogen are placed far away.

I think there are other locality tests we may want to add that are similar, but subtly different, such as placing hydrogens near an aluminium slab.

Perhaps I should also rename the benchmark locality, rather than ghost atoms?

## Linked issue

Resolves #103

## Progress

- [x] Calculations
- [x] Analysis
- [x] Application
- [x] Documentation

## Testing

Tested on several MACE models, Orb, and UMA.

## New decorators/callbacks

Adds a new callback to view a structure based on table column. This mostly seems to work, although the second ghost atoms structure doesn't seem to display?

Ideally we'd maybe  show the different forces from the randomly placed hydrogens, and link each of those to the specific structure, but the hydrogens are also very hard to see.


ml_peg/analysis/physicality/ghost_atoms/analyse_ghost_atoms.py
```diff
@@ -0,0 +1,149 @@
...
+        "Random hydrogen std ΔF": "Standard deviation of force difference on solute "
+        "atoms due to random hydrogen atoms (meV/Å)",
+    },
+    thresholds=DEFAULT_THRESHOLDS,
+)
+def metrics(
+    ghost_force: dict[str, float], hydrogen_force: dict[str, tuple[float, float]]
+) -> dict[str, dict]:
+    """
+    Get all Li diffusion metrics.
```
ElliottKasoar:
```suggestion
    Get all ghost atom metrics.
```

ElliottKasoar:
Note: this now also adds 'first implementation', similar to #109, of setting the default data type for a specific benchmark. This will also be used in #117.

joehart2001 (approved):
only thing is that maybe good thresholds should be lower but i guess in multihead papper table XX they are similar.

PR #122, opened 2025-11-01
Add level of theory flags

### LINKED ISSUE #121: Flag when there is a mismatch between a model's level of theory and the reference data
### Summary

also improve tooltips for mlips to view what the mismatch it

### Example

_No response_

### Alternatives

_No response_

### Additional context

_No response_

joehart2001:

## Summary
- add level of theory to metric and model yaml
- add level of theory mismatch flag and MLIP tooltip for model info
- add level of theory to each metric tooltip

## Linked issue

Resolves #121 

## Testing
mace models

### DIFF
--- a/ml_peg/analysis/utils/decorators.py
+++ b/ml_peg/analysis/utils/decorators.py
@@ -15,6 +15,7 @@ import plotly.graph_objects as go
 
 from ml_peg.analysis.utils.utils import calc_table_scores
 from ml_peg.app.utils.utils import Thresholds
+from ml_peg.models.get_models import load_model_configs
 
 
 def plot_parity(
@@ -669,9 +670,6 @@ def build_table(
                 id="metrics",
                 tooltip_header=tooltip_header,
             )
-
-            # Load model configurations and extract level of theory metadata
-            from ml_peg.models.get_models import load_model_configs
 
             model_configs, model_levels = load_model_configs(mlips)
 


ml_peg/models/models.yml
```diff
@@ -2,34 +2,39 @@ mace_mp_0a:
...
   device: "auto"
+  level_of_theory: PBE
   kwargs:
     model: "medium-omat-0"
 
 mace_matpes_r2scan:
   module: mace.calculators
   class_name: mace_mp
   device: "auto"
+  level_of_theory: PBE
```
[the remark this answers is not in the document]
joehart2001: yep! thanks!

joehart2001:
```suggestion
  level_of_theory: r2SCAN
```

ElliottKasoar:
Rebased

I get some strange spacings for some of the warnings:

image

I also don't seem to get any hover info for the theory on several of the benchmarks e.g. LNCI16, even though the `level_of_theory` is present in the metrics.yml?

ElliottKasoar:
I think we also want some sort of list of known levels of theory (e.g. via a set of `Literal` strings. Currently we have clashes e.g. 
image

Obviously we could do things like compare lowercase strings, but there will be less trivial alternatives to how things are written, and we also don't really want these inconsistencies, even if we catch them.

ml_peg/analysis/utils/decorators.py
```diff
@@ -665,6 +673,35 @@ def build_table_wrapper(*args, **kwargs) -> dict[str, Any]:
                 tooltip_header=tooltip_header,
             )
 
+            with open(MODELS_ROOT / "models.yml", encoding="utf8") as model_file:
```
ElliottKasoar: Maybe this code block should go in models.py, alongside `load_models` etc?

That has bits such as

```
    # Load models from registry YAML: models.yml
    with open(MODELS_ROOT / "models.yml") as file:
        all_models = yaml.safe_load(file)

    for name, cfg in get_subset(all_models, models).items():
```
which we may be able to use here too ?

ml_peg/analysis/utils/decorators.py
```diff
@@ -665,6 +673,35 @@ def build_table_wrapper(*args, **kwargs) -> dict[str, Any]:
...
+                model_configs[mlip] = cfg
+                model_levels[mlip] = cfg.get("level_of_theory")
+            model_levels: dict[str, str | None] = {}
+            model_configs: dict[str, Any] = {}
+            for mlip in mlips:
+                cfg = deepcopy(all_models.get(mlip) or {})
+                if not isinstance(cfg, dict):
+                    cfg = {}
+                model_configs[mlip] = cfg
+                model_levels[mlip] = cfg.get("level_of_theory")
```
ElliottKasoar: This is duplicated?

ml_peg/app/utils/register_callbacks.py
```diff
@@ -29,18 +30,21 @@ def register_summary_table_callbacks() -> None:
     @callback(
         Output("summary-table", "data"),
         Output("summary-table", "style_data_conditional"),
+        Output("summary-table", "tooltip_data"),
```
ElliottKasoar: Why is this changing when it didn't before?

ml_peg/models/models.py
```diff
@@ -17,6 +19,42 @@
 current_models = None
 
 
+def load_model_configs(
```
ElliottKasoar: Sorry I probably said to put it here, can it be moved to get_models.py? This module only really defines model classes, whereas that actually loads them in different was, as does this

ml_peg/app/utils/utils.py
```diff
@@ -3,9 +3,14 @@
 from __future__ import annotations
 
 from collections.abc import Mapping, MutableMapping, Sequence
```
ElliottKasoar: This wasn't newly introduced, but how many of the type hints are actually only `Mapping`s, and not always a `dict` in reality?

It's not something I've encountered a huge amount, but I'm pretty sure `Mapping` as an abstract base class doesn't require the implementation of quite a few things that we may depend on, such as `pop`, `update`, `__setitem__` (I think `MutableMapping` does), `__delitem__`, or `setdefault`.

I think in most cases these probably should just be `dict`0s

PR #204, opened 2025-12-01
Label D3 in tables and add trained on D3 model parameter 

### LINKED ISSUE #139: Add check when building D3 SumCalculator
### Summary

If a calculator has setting that means it always uses D3, running a test which adds D3 may break/create duplicate the D3 correction.

We can add a check in the function to make sure this doesn't happen.

### Example

_No response_

### Alternatives

_No response_

### Additional context

_No response_

### LINKED ISSUE #202: label models with D3
### Summary

add d3 labels to model names where relevant. some models are trained on d3 so dont require adding d3

### Example

_No response_

### Alternatives

_No response_

### Additional context

_No response_

joehart2001:

## Summary

1. add "-D3" to model names in tests in which D3 correction is added
2. new model arg in models.yml: trained_on_d3
     - if this is true, then add_d3_calculator returns the original calculator during the calculation, and "-D3" is not               
         added to the mode name during analysis

## Linked issue

Resolves #139, resolves #202

## Testing
all models + set trained_on_d3 to true for a random mace model to test if it actually didnt apply the D3 dispersion during the calculation as well as in the model table

### DIFF
--- a/ml_peg/analysis/utils/decorators.py
+++ b/ml_peg/analysis/utils/decorators.py
@@ -627,7 +627,7 @@ def build_table(
             # Use MLIP keys from first (any) metric keys
             mlips = tuple(next(iter(results.values())).keys())
 
-            name_map = dict(mlip_name_map or {})
+            name_map = mlip_name_map if mlip_name_map else {}
             display_names = {mlip: name_map.get(mlip, mlip) for mlip in mlips}
             display_values = list(display_names.values())
             if len(display_values) != len(set(display_values)):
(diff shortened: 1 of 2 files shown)


ml_peg/analysis/utils/decorators.py
```diff
@@ -622,12 +627,22 @@ def build_table_wrapper(*args, **kwargs) -> dict[str, Any]:
             # Use MLIP keys from first (any) metric keys
             mlips = tuple(next(iter(results.values())).keys())
 
+            name_map = dict(mlip_name_map or {})
```
ElliottKasoar:
```suggestion
            name_map = mlip_name_map if mlip_name_map else {} 
```
Probably don't need to make a copy of the dict?

PR #247, opened 2026-01-07
Fix D3 correction for r2SCAN model

ElliottKasoar:

## Summary

When adding the `TorchDFTD3Calculator`, the functional should match that of the model. In all other cases where we add dispersion, this is PBE, but for r2SCAN, it should be D4, which is set via `xc=r2scan` (see: https://github.com/pfnet-research/torch-dftd/pull/29)

## Testing

Tested for mace-matpes-r2scan

### DIFF
--- a/ml_peg/models/models.yml
+++ b/ml_peg/models/models.yml
@@ -48,6 +48,8 @@ mace-matpes-r2scan:
   kwargs:
     model: "mace-matpes-r2scan-0"
     head: "matpes_pbe"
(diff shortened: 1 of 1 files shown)


joehart2001:
> Ah, so the table will still state `mace-matpes-r2scan-D3`, so perhaps we need to fix that alongside this, since at least the current label is correct, even if the correction isn't the best?

yeah agreed

alinelena (approved):

PR #315, opened 2026-01-29
add qmof test

alinelena:
Please click the `Preview` tab and select a PR template:

- [New benchmark for existing models](?expand=1&template=new_benchmark.md&labels=new+benchmark)
- [Benchmark a new model](?expand=1&template=new_model.md&labels=new+model)
- [All other proposed changes](?expand=1&template=custom.md)


joehart2001:
Screenshot 2026-08-14 at 15 06 30

joehart2001:
@alinelena is working well. can you check this: i changed the metric to eV/atom as in the mace mp paper, is this correct? would you be able to take a look over the calc and analysis to check everything is correct. the app is looking good now

ElliottKasoar:
For future reference, I think this will be in the `porous_materials` category, and QMOF benchmark name (which we should use for the data re-upload too)

ml_peg/calcs/mofs/qmof/calc_qmof.py
```diff
@@ -0,0 +1,59 @@
...
+
+    qmof_energy_dir = (
+        download_s3_data(
+            key="inputs/mofs/qmof/qmof.zip",
+            filename="qmof.zip",
+        )
+        / "qmof"
+    )
+    input_file = "qmof_valid_structures.traj"
+    mofs = read(qmof_energy_dir / input_file, index=":100")
```
joehart2001:
```suggestion
    mofs = read(qmof_energy_dir / input_file, index=":")
```

forgot to change this back

alinelena:
@joehart2001 what is left to do on this?

docs/source/user_guide/benchmarks/mofs.yml
joehart2001, line 1: @alinelena docs

ml_peg/app/mofs/mofs.yml
```diff
@@ -0,0 +1,2 @@
+    title: Metal Organic Frameworks
+    description: various MOFs tests
```
joehart2001: change to porous materials?

joehart2001:
otherwise i think its pretty ready and @ElliottKasoar can take a look. should be quick

ml_peg/calcs/porous_materials/QMOF/calc_QMOF.py
```diff
@@ -0,0 +1,59 @@
...
+        download_s3_data(
+            key="inputs/porous_materials/QMOF/QMOF.zip",
+            filename="QMOF.zip",
+        )
+        / "QMOF"
+    )
+    input_file = "qmof_valid_structures.traj"
+    mofs = read(qmof_energy_dir / input_file, index=":")
+    for mof in tqdm(mofs, desc=model_name):
+        mof.calc = copy(calc)
```
ElliottKasoar, line 53: We need a spin multiplicity and charge here

ml_peg/calcs/porous_materials/QMOF/calc_QMOF.py
```diff
@@ -0,0 +1,59 @@
...
+        sp = SinglePoint(struct=mof)
```
ElliottKasoar, line 54: We should wrap this in a try...except so some can fail without the whole test failing

ml_peg/analysis/porous_materials/QMOF/analyse_QMOF.py
```diff
@@ -0,0 +1,202 @@
...
+
+        struct_file = "qmof_valid_structures.traj"
+        mofs = read(model_dir / struct_file, index=":")
+        for mof in mofs:
+            mof_energy = mof.get_potential_energy() / len(mof)
+
+            results[model_name].append(mof_energy)
+
+            # Store reference energies (only once)
+            if not ref_stored:
```
ElliottKasoar, line 93: We may need to be a little careful with this. Particularly once some structures are allowed to fail, the reference values stored may not be the full set. The way we mostly deal with this elsewhere is to check that the length of `results["ref"]` matches the length of one of the `INFO` keys, since that comes from the mock calculator that should work for all structures, and only then change the value of `ref_stored`.

ElliottKasoar (commented):
We can revert if I'm mistaken, but I don't think we want porous_materials/mofs/qmof - we don't have a way of dealing with nested categories at the moment, and (subject to change) our current convention is for test names to match the standard form (QMOF), so I've pushed these changes.

I'm testing one or two models as well, but otherwise just a few things to tidy up I hope

docs/source/user_guide/benchmarks/porous_materials.yml
ElliottKasoar, line 1: (This is still to do)

ml_peg/analysis/porous_materials/QMOF/analyse_QMOF.py
```diff
@@ -0,0 +1,202 @@
...
+    dict[str, list]
+        Dictionary of reference and predicted energies per atom.
+    """
+    results = {"ref": []} | {mlip: [] for mlip in MODELS}
+    ref_stored = False
+
+    for model_name in MODELS:
+        model_dir = CALC_PATH / model_name
+
+        if not model_dir.exists():
```
ElliottKasoar, line 82: Depending on how you deal with errors in the calculation,, we may need extra checks to make sure scores are `NaN` if any failed.

If we still write every structure, but set the energy to NaN before writing, this is probably fine though.

ml_peg/analysis/porous_materials/QMOF/analyse_QMOF.py
```diff
@@ -0,0 +1,202 @@
...
+            continue
+
+        struct_file = "qmof_valid_structures.traj"
+        mofs = read(model_dir / struct_file, index=":")
```
ElliottKasoar, line 86:
```suggestion
        struct_path = CALC_PATH / model_name / "qmof_valid_structures.traj"

        if not struct_path.exists():
            continue

        mofs = read(struct_path, index=":")
```
I think this is safer since the directory may exist but the file may not have been written successfully, which would cause this to error out.

joehart2001 (changes requested):

PR #321, opened 2026-01-29
Extend elasticity benchmark to full elastic constant matrix

### LINKED ISSUE #298: Extending elasticity benchmark to full tensors
### Summary

Right now the elasticity benchmark only compares the Voigt-Reuss-Hill average bulk modulus and shear strain. A better and more informative benchmark would compare the full elasticity tensors.

### Interactive features

An entry to the elasticity table with a mean-average-error over all symmetry-reduced elastic constant components. In the case where the measured elasticity tensor does not seem to align with the `pymatgen` predicted crystal symmetry, fall back to doing a direct comparison of all components in the upper right triangle.

### Category

Bulk crystals / elasticity.

### Data availability

Full elastic constant tensor materials project reference data is available at a PBE level of theory.

### Computational cost

The same amount of time as the elasticity benchmark - the full test is likely to take hours-days to run on GPU.

### Additional context

_No response_

Fraser-Birks:

## Summary

Adds an additional test to the elasticity benchmark which compares the mean average error over the symmetrically independent components of the full elasticity tensors. 

## Linked issue

Resolves #298

## Progress

- [x] Calculations
- [x] Analysis
- [x] Application
- [x] Documentation

## Testing

mace-mpa-small

## New decorators/callbacks

n/a


ElliottKasoar (commented):
Thanks for this, @Fraser-Birks!

I think you'd discussed this with @joehart2001, but it would be good to save the structures post-initial relaxation, for potential future visualisation.

Is there any other visualisation of the elastic tensor MAE that we may want to add (probably as a new PR at this point)? I seem to recall seeing some cool visualisations e.g. via https://github.com/coudertlab/elate

Also a minor note that will probably not come up again, and not something that needs changing, but the types in the parameters section of docstrings aren't necessary, as I think the real type annotations should be used when we create the API docs anyway.

ml_peg/analysis/bulk_crystal/elasticity/analyse_elasticity.py
```diff
@@ -33,6 +34,221 @@
...
+    rtol : float, optional
+        Relative tolerance for consistency checks (default: 0.10).
+
+    Returns
+    -------
+    tuple of ndarray
+        Independent elastic constants for reference and comparison tensors.
+    """
+
+    def _allclose(values):
```
ElliottKasoar:
```suggestion
    def _allclose(values) -> bool:
```
Something like [ArrayLike](https://numpy.org/devdocs/reference/typing.html#numpy.typing.ArrayLike) for values too?

Fraser-Birks:
Have now:

- Addressed comments above (np.isclose, test inputs, type labelling). 
- Added structure saving + violin plots for each of the elasticity quantities. Seems to work fine when tested locally on 100 unrelaxed structures with mpa-medium, but full set will need to be rerun.

joehart2001:
> Have now:
> 
> * Addressed comments above (np.isclose, test inputs, type labelling).
> * Added structure saving + violin plots for each of the elasticity quantities. Seems to work fine when tested locally on 100 unrelaxed structures with mpa-medium, but full set will need to be rerun.

Great! i've just pushed a few fixes and should be good to go after @ElliottKasoar takes a look

ElliottKasoar (approved):

joehart2001 (approved):

PR #327, opened 2026-01-30
EOS test as in section A13 of mace mp0 paper

### LINKED ISSUE #310: Equation of state for BCC (A13) and other elemental structures
### Summary

In section A13 we compared EOS for BCC and FCC phases for W, Mo and Nb. Together with comparing lattice constants and bulk moduli this served as a test of relative stability between these two phases.  For BCC metals adding A15 phase is also of interest and can be challenging for interatomic potentials. 

### Interactive features

Since this is really related to lattice constants and bulk modulus calculations. A plot of EOS can be shown when a user clicks on the plot for the lattice constants next to the equilibrium structure. 

Image

Otherwise, a separate section can be added. Maybe grouping the results by element and/or equilibrium structure is possible.

### Category

Bulk crystals / elasticity. Physicality.

### Data availability

For W, M, Nb the data was collected from different sources. In current lattice constants test there are 16 elemental structures. For the extension of the test to other elements one has to look for data.

For equilibrium structures and EOS we could look for the results of 

[Reproducibility in density functional theory calculations of solids](https://www.science.org/doi/10.1126/science.aad3000?ijkey=teUZMpwU49vhY&keytype=ref&siteid=sci)

Moreover, the propose a single value metric for comparing EOS for equilibrium structure. 

### Computational cost

Low

### Additional context

I will start with adding the test for three BCC metals and then we can discuss how to generalize it to other elemental structures.

pgrigorev:

## Summary

EOS test for three phases as in A13 of mace-mp0 paper. W, Mo and Nb are checked with one extra phase A15. I have not yest come up with a metric, hopefully we can think about it together. 

To help to think about the metric and integration to the interactive app here is a plot that can be done with the results.
eos_comparison_all_models

The _performance_  varies quite a lot between the models, should be interesting to put a number on it. I attach also the reference data: [equation_of_state.zip](https://github.com/user-attachments/files/24965762/equation_of_state.zip)
I should be able to get similar data for Ta,  non BCC Zr, Ti and maybe a couple of more metals. 

## Linked issue

Resolves #310 

## Progress

- [x] Calculations
- [x] Analysis
- [x] Application
- [x] Documentation

## Testing

'mace-matpes-r2scan',
 'mace-mp-0a',
 'mace-mp-0b3',
 'mace-mpa-0',
 'mace-omat-0',
 'orb-v3-consv-inf-omat',
 'pet-mad'
'GRACE-2L-OMAT'

## New decorators/callbacks

- added zmin, zmax parameters to `plot_periodic_table()` in order to synchronize with the limits of the summary table. 
- added `show_eos_curve()` callback locally to the `app_equation_of_state.py`. Will be happy to discuss if that was actually necessary and if it is possible to avoid adding a new decorator.


ml_peg/calcs/bulk_crystal/equation_of_state/calc_equation_of_state.py
```diff
@@ -0,0 +1,152 @@
...
+import numpy as np
+import pandas as pd
+import pytest
+
+from ml_peg.models.get_models import load_models
+from ml_peg.models.models import current_models
+
+MODELS = load_models(current_models)
+
+DATA_PATH = Path(__file__).parent / "../../../../inputs/bulk_crystal/equation_of_state/"
```
ElliottKasoar: We can replace this with an S3 download later

ml_peg/calcs/bulk_crystal/equation_of_state/calc_equation_of_state.py
```diff
@@ -0,0 +1,152 @@
...
+
+        phases = [col.split("_")[1] for col in dft_data.columns if "Delta" in col]
+
+        for phase in phases:
+            assert phase in lattices, f"Lattice {phase} not implemented for EOS test."
+            lattice = lattices[phase]
+            """'
+            start_time = datetime.now()
+            print(f"Start time for {phase} @ {model_name}: {start_time}")
+            """
```
ElliottKasoar: We can probably delete this?

ElliottKasoar (commented):
Thanks for this @pgrigorev, it's looking really good so far!

This relates to a one of the comments I've left, but how general do you expect this to be? I'm wondering if we may want to rename `equation_of_state` as a test slightly to make the scope clearer, since otherwise it's completely unqualified, if that makes sense.

From testing, things mostly seem to work, but I think the final part of the interactivity doesn't. I can definitely help fix it, but I suggest simplifying the data slightly as I've commented first before we deal with that.

ml_peg/calcs/bulk_crystal/equation_of_state/calc_equation_of_state.py
```diff
@@ -5,7 +5,10 @@
 from pathlib import Path
 from typing import Any
 
+from ase.filters import UnitCellFilter
```
ElliottKasoar: Generally the `ExpCellFilter` is recommended as more performant, are you sure you want this?

https://ase-lib.org/ase/filters.html#the-unitcellfilter-class

docs/source/user_guide/benchmarks/bulk_crystal.rst
```diff
@@ -115,3 +115,57 @@ Reference data:
...
+   volume point. This metric is practically important: it indicates whether a model predicts spurious phase transitions under extreme conditions, such as the high tensile stresses at a crack tip.
+
+Computational cost
+------------------
+Low. Every eos curve requires 50 calls on unit cells.
+
+Data availability
+-----------------
+
+The reference data for W, Nb, Mo is taken from: `Čák, M., Hammerschmidt, T., Rogal, J., Vitek, V., & Drautz, R. (2014). Analytic bond-order potentials for the bcc refractory metals Nb, Ta, Mo and W. Journal of Physics Condensed Matter, 26(19), 195501. <https://doi.org/10.1088/0953-8984/26/19/195501>`_
```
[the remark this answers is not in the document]
pgrigorev: I put in the summary that those are PBE calculations. But I can add it once more on the reference list. Here I construct the cells from the name of the phases since I do not have the DFT calculation results. This can be an option to run this calculations in a controlled manner and have them as an input. This will simplify the calculation script.

ml_peg/calcs/bulk_crystal/energy_volume_curves_metals/calc_energy_volume_curves_metals.py
```diff
@@ -0,0 +1,152 @@
...
+    # dummy call to have calc_num_atoms available
+    lattice(symbol="W", latticeconstant=3.16)
+    lattice_constants = (volumes_per_atoms * lattice.calc_num_atoms()) ** (1 / 3)
+
+    structures = [
+        lattice(latticeconstant=lc, size=size, symbol=symbol)
+        for lc in lattice_constants
+    ]
+    for structure in structures:
+        structure.calc = calc
```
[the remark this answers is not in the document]
pgrigorev: yes I could do it. Especially for magnetic metals it should change the calculation for such models. I will think about the best way to do it.

ml_peg/app/bulk_crystal/equation_of_state/app_equation_of_state.py
```diff
@@ -0,0 +1,205 @@
...
+
+MODELS = get_model_names(current_models)
+BENCHMARK_NAME = "Equation of State (metals)"
+DOCS_URL = (
+    "https://ddmms.github.io/ml-peg/user_guide/"
+    "benchmarks/bulk_crystal.html#equation-of-state"
+)
+DATA_PATH = APP_ROOT / "data" / "bulk_crystal" / "equation_of_state"
+CALC_PATH = CALCS_ROOT / "bulk_crystal" / "equation_of_state" / "outputs"
+INPUT_PATH = Path(__file__).parents[4] / "inputs" / "bulk_crystal" / "equation_of_state"
```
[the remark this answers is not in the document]
pgrigorev: so instead of `CALC_PATH` I should write the outputs to `DATA_PATH\outputs` or somtehing like this?

ml_peg/analysis/bulk_crystal/energy_volume_curves_metals/analyse_energy_volume_curves_metals.py
```diff
@@ -0,0 +1,518 @@
...
+                else f"{metric_name} (%)"
+            )
+            colorbar_title = title
+            file_suffix = file_suffixes[metric_name]
+            values = get_metric_per_element(model, eos_stats, metric_name)
+            if (
+                DEFAULT_THRESHOLDS[metric_name]["good"]
+                > DEFAULT_THRESHOLDS[metric_name]["bad"]
+            ):
+                plot_periodic_table(
```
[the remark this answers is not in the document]
pgrigorev: So for now we have 9: (W, Nb, Mo, Ta, Ti, Zr, Cr, Fe) - find the updated zipfile attached:
[equation_of_state.zip](https://github.com/user-attachments/files/26717350/equation_of_state.zip). I could add couple more, but I have to look for the data. 

> This relates to a one of the comments I've left, but how general do you expect this to be? I'm wondering if we may want to rename `equation_of_state` as a test slightly to make the scope clearer, since otherwise it's completely unqualified, if that makes sense.

I agree that we could rename this to something like `energy_volume_curves_metals` since we are looking on a couple of more things than just the ground state. 

To make the test more general I could look at:
- [https://github.com/molmod/DeltaCodesDFT](https://github.com/molmod/DeltaCodesDFT) which has the `.cif` files for 71 element: H He Li Be B C N O F Ne Na Mg Al Si P S Cl Ar K Ca Sc Ti V Cr Mn Fe Co Ni Cu Zn Ga Ge As Se Br Kr Rb Sr Y Zr Nb Mo Tc Ru Rh Pd Ag Cd In Sn Sb Te I Xe Cs Ba Lu Hf Ta W Re Os Ir Pt Au Hg Tl Pb Bi Po Rn. Going through `.cif` files the calculation can be automated easily and we have also the metric already imlemented. There are reference values for few DFT codes including vasp.
- Continuation of that work has not only ground states for 91 elements and also data for oxides. [https://acwf-verification.materialscloud.org/](https://acwf-verification.materialscloud.org/). It uses aiida workflows and I am not sure how easy it would be to adapt the workflow for us or extract the data. Both works use **non magentic** calculations. 

I am open to discussion. Using the set for 71 materials for general EOS test seems to be easy to do. Exploring the second dataset/workflows seems a bit more demanding. My original idea was a simple test for metals community. This is essentially one of the first tests during material specific potential development. I think it is wise to separate it from a general purpose EOS test across the periodic table. Especially for systems where magnetism is important (Fe, Ni).

ml_peg/calcs/bulk_crystal/equation_of_state/calc_equation_of_state.py
```diff
@@ -0,0 +1,152 @@
...
+            print(duration)
+            """
+            results[f"{phase}_a"] = lattice_constants
+            results[f"{phase}_E"] = energies
+
+        write_dir = OUT_PATH / model_name
+        df = pd.DataFrame(results)
+        output_file = write_dir / f"{element}_eos_results.csv"
+        write_dir.mkdir(parents=True, exist_ok=True)
+        df.to_csv(output_file, index=False)
```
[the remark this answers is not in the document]
pgrigorev: This should be easy to do :)

ml_peg/calcs/bulk_crystal/equation_of_state/calc_equation_of_state.py
```diff
@@ -0,0 +1,152 @@
...
+    ----------
+    mlip
+        Tuple of (model_name, model) as provided by pytest parametrize.
+    """
+    model_name, model = mlip
+    calc = model.get_calculator()
+
+    fns = list(DATA_PATH.glob("*DFT*"))
+
+    for fn in fns:
```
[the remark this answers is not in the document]
pgrigorev: it is short for `filenames`. My old habit and I will be happy to refactor.

docs/source/user_guide/benchmarks/bulk_crystal.rst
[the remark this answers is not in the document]
ElliottKasoar: Good point, thanks for that, but yes if you can add it here as well, that would be great, just so someone always knows where to look.

I'm not sure I fully follow the change you're suggesting for the new option? If you think it makes things simpler, that's great though

ml_peg/calcs/bulk_crystal/equation_of_state/calc_equation_of_state.py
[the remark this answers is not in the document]
ElliottKasoar: Ah I see! Yes, something like `filename in filenames` would be great

ml_peg/analysis/bulk_crystal/energy_volume_curves_metals/analyse_energy_volume_curves_metals.py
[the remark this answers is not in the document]
ElliottKasoar: I see, thanks!

To a large extent it's up to you, and how you think it would be useful to break this down.

My inclination would be to keep this as a (relatively) simple metal test, which as you say is a useful starting point, and then we can potentially reuse/refactor parts of this for a more general test, which also sounds really interesting.

Assuming we do keep this as a limited number of metals to start with, I'd consider whether we need the full periodic table, or whether there's a more focused option (maybe even just a dropdown), but ultimately it's up to you how best to visualise your results.

ml_peg/analysis/bulk_crystal/equation_of_state/metrics.yml
```diff
@@ -0,0 +1,19 @@
...
+        level_of_theory: PBE
+  Phase energy:
+        good: 0.0
+        bad: 200.0
+        unit: meV/atom
+        tooltip: Mean absolute error of the phase energy differences relative to the reference phase between model and DFT.
+        level_of_theory: PBE
+  Phase stability:
+        good: 100.0
+        bad: 0.0
```
ElliottKasoar: How much have you considered and/or understand the motivation behind the good and bad thresholds?

To me, for example, I would probably have expected a non-0% bad metric for the phase stability, since say getting only 10% right is probably not really useful?

But ultimately these are your choices!

ml_peg/app/bulk_crystal/energy_volume_curves_metals/app_energy_volume_curves_metals.py
```diff
@@ -0,0 +1,137 @@
...
+from dash.exceptions import PreventUpdate
+from dash.html import Div
+from plotly.io import read_json
+
+from ml_peg.app import APP_ROOT
+from ml_peg.app.base_app import BaseApp
+from ml_peg.app.utils.build_callbacks import plot_from_table_cell
+from ml_peg.models.get_models import get_model_names
+from ml_peg.models.models import current_models
+
```
ElliottKasoar, line 16:
```suggestion
from ml_peg.models import current_models
from ml_peg.models.get_models import get_model_names

```
This has moved, sorry! You may need to rebase for it to work

ml_peg/calcs/bulk_crystal/equation_of_state/calc_equation_of_state.py
```diff
@@ -0,0 +1,289 @@
...
+    ----------
+    mlip
+        Tuple of (model_name, model) as provided by pytest parametrize.
+    """
+    model_name, model = mlip
+
+    write_dir = OUT_PATH / model_name
+    write_dir.mkdir(parents=True, exist_ok=True)
+
+    calc = model.get_calculator()
```
ElliottKasoar:
```suggestion
    calc = model.get_calculator(precision="high")
```
You may need to rebase, but this now takes a precision parameter that should convert to float64 generally, unless you want lower precision

ml_peg/calcs/bulk_crystal/equation_of_state/calc_equation_of_state.py
```diff
@@ -0,0 +1,289 @@
...
+import pandas as pd
+import pytest
+from tqdm import tqdm
+
+from ml_peg.models import current_models
+from ml_peg.models.get_models import load_models
+
+MODELS = load_models(current_models)
+
+DATA_PATH = Path(__file__).parent / "../../../../inputs/bulk_crystal/equation_of_state/"
```
ElliottKasoar:
```suggestion
from ml_peg.calcs.utils.utils import download_s3_data
from ml_peg.models import current_models
from ml_peg.models.get_models import load_models

MODELS = load_models(current_models)
```
In combination with below

ml_peg/calcs/bulk_crystal/equation_of_state/calc_equation_of_state.py
```diff
@@ -0,0 +1,289 @@
...
+    mlip
+        Tuple of (model_name, model) as provided by pytest parametrize.
+    """
+    model_name, model = mlip
+
+    write_dir = OUT_PATH / model_name
+    write_dir.mkdir(parents=True, exist_ok=True)
+
+    calc = model.get_calculator()
+    filenames = list(DATA_PATH.glob("*DFT*"))
```
ElliottKasoar:
```suggestion

    data_path = (
        download_s3_data(
            filename="equation_of_state.zip",
            key="inputs/bulk_crystal/equation_of_state/equation_of_state.zip",
        )
        / "equation_of_state"
    )
    filenames = list(data_path.glob("*DFT*"))
```

ml_peg/calcs/bulk_crystal/equation_of_state/calc_equation_of_state.py
```diff
@@ -0,0 +1,289 @@
...
+        if symbol in magnetic_moments and lattice_name in magnetic_moments[symbol]:
+            structure.info["spin"] = magnetic_moments[symbol][lattice_name] * len(
+                structure
+            )
+        else:
+            structure.info["spin"] = 0
+        structure.calc = calc
+
+    for structure in structures:
+        structure.get_potential_energy()
```
ElliottKasoar: This isn't something we enforce yet strictly, but it would be useful to do a try/except around this, and perhaps if it fails set the energy to `NaN`, so it's easier for us to take partial results (e.g. if one of the elements is unsupported/fails for other reasons) later.

ml_peg/app/bulk_crystal/equation_of_state/app_equation_of_state.py
```diff
@@ -0,0 +1,137 @@
...
+from ml_peg.app.base_app import BaseApp
+from ml_peg.app.utils.build_callbacks import plot_from_table_cell
+from ml_peg.models.get_models import get_model_names
+from ml_peg.models.models import current_models
+
+MODELS = get_model_names(current_models)
+BENCHMARK_NAME = "Equation of State (metals)"
+DOCS_URL = (
+    "https://ddmms.github.io/ml-peg/user_guide/"
+    "benchmarks/bulk_crystal.html#equation-of-state"
```
ElliottKasoar:
```suggestion
    "benchmarks/bulk_crystal.html#equation-of-state-metals"
```
Since it corresponds to the .rst header

ml_peg/analysis/bulk_crystal/equation_of_state/analyse_equation_of_state.py
```diff
@@ -0,0 +1,518 @@
...
+from ml_peg.app import APP_ROOT
+from ml_peg.calcs import CALCS_ROOT
+from ml_peg.models.get_models import get_model_names
+from ml_peg.models.models import current_models
+
+MODELS = get_model_names(current_models)
+CALC_PATH = CALCS_ROOT / "bulk_crystal" / "equation_of_state" / "outputs"
+OUT_PATH = APP_ROOT / "data" / "bulk_crystal" / "equation_of_state"
+
+DATA_PATH = Path(__file__).parent / "../../../../inputs/bulk_crystal/equation_of_state/"
```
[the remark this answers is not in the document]
ElliottKasoar: With the changes I suggested below to download from S3, I think we need to update this, ideally to use data from within `CALC_PATH`, which may involve minor changes to things that are saved during the calc

ElliottKasoar (commented):
Thanks again for all the work that's gone into this! I've left a few more comments, but generally I think it looks amazing!

Another small question: is it expected that the lines here don't seem to span the same volume range as the DFT references?

image

Would you also be able to resolve the merge conflict with the docs please?

ml_peg/calcs/bulk_crystal/equation_of_state/calc_equation_of_state.py
ElliottKasoar: (I've uploaded the newer file, so this should work I believe!)

pgrigorev:
> Thanks again for all the work that's gone into this! I've left a few more comments, but generally I think it looks amazing!
> 
> Another small question: is it expected that the lines here don't seem to span the same volume range as the DFT references?
> 
> image
> Would you also be able to resolve the merge conflict with the docs please?

There min and max values were detected only from the most stable phase and in case in Ni it lead to smaller scanning interval. I have updated the calculation function to detect vmin and vmax over all phases in the data file.

ml_peg/analysis/bulk_crystal/equation_of_state/metrics.yml
pgrigorev: I did not put too much thought to it to be honest. By agree that even 25 % of correct stability is not very useful. I updated the value.

pgrigorev:
I think I have resolved all the conflicts and conversation. Two main things that are left to decide before it can be merged:
- Drop down vs periodic table. I personally prefer periodic table. Hopefully empty slots will encourage people to add reference data :). If that is technically possible we could shoe only metals part of the periodic table
- The name of the test. Since I would like to add test for ground state equation of state as we discussed before, I think we could rename it to `metals_energy_volume_curves`. This would require renaming the files and paths, but I think it should be manageable.

ml_peg/calcs/bulk_crystal/equation_of_state/calc_equation_of_state.py
```diff
@@ -0,0 +1,324 @@
...
+    ]
+
+    for structure in structures:
+        structure.info["charge"] = 0
+        if symbol in magnetic_moments and lattice_name in magnetic_moments[symbol]:
+            structure.info["spin"] = magnetic_moments[symbol][lattice_name] * len(
+                structure
+            )
+        else:
+            structure.info["spin"] = 0
```
ElliottKasoar: I think normally the default is 1 since it's spin multiplicity. I'm not sure if the above is also correct given that?

ElliottKasoar (commented):
Thanks for all the updates, @pgrigorev!

I think something may have gone wrong when you updated the branch, there seem to be a bunch of extra files related to the last few commits to be merged into main

ElliottKasoar:
> I think I have resolved all the conflicts and conversation. Two main things that are left to decide before it can be merged:
> 
> * Drop down vs periodic table. I personally prefer periodic table. Hopefully empty slots will encourage people to add reference data :). If that is technically possible we could shoe only metals part of the periodic table
> * The name of the test. Since I would like to add test for ground state equation of state as we discussed before, I think we could rename it to `metals_energy_volume_curves`. This would require renaming the files and paths, but I think it should be manageable.

I think the periodic table is quite nice, especially now the number of metals covered has grown, so I'd be inclined to leave it.

I'm also pretty open with respect to the name. We've had some tests that cover multiple categories (GSCDB138). In that case, we had slight variations of it e.g. `GSCDB138_NCIs`.

Keeping it as is would probably be fine, so I don't mind for now, but something like `equation_of_state_metals` or your suggestion of`metals_energy_volume_curves` might be more future-proof.

ml_peg/calcs/bulk_crystal/equation_of_state/calc_equation_of_state.py
```diff
@@ -0,0 +1,324 @@
...
+                / (np.sqrt(2))
+            ) ** (1 / 3)
+            lattice_name = "HCP"
+        try:
+            unit_cell = lattice(symbol=symbol, latticeconstant=(a0, a0 * ideal_ratio))
+            unit_cell.info["charge"] = 0
+            if symbol in magnetic_moments and lattice_name in magnetic_moments[symbol]:
+                unit_cell.info["spin"] = magnetic_moments[symbol][lattice_name] * len(
+                    unit_cell
+                )
```
ElliottKasoar:
```suggestion
                unit_cell.info["spin"] = int(
                    round(magnetic_moments[symbol][lattice_name] * len(unit_cell))
                )
```
UMA requires this to be an int, and we normally just round first

ml_peg/calcs/bulk_crystal/equation_of_state/calc_equation_of_state.py
```diff
@@ -0,0 +1,324 @@
...
+        lattice(latticeconstant=lc, size=size, symbol=symbol)
+        for lc in lattice_constants
+    ]
+
+    for structure in structures:
+        structure.info["charge"] = 0
+        if symbol in magnetic_moments and lattice_name in magnetic_moments[symbol]:
+            structure.info["spin"] = magnetic_moments[symbol][lattice_name] * len(
+                structure
+            )
```
ElliottKasoar:
```suggestion
            structure.info["spin"] = int(
                round(magnetic_moments[symbol][lattice_name] * len(structure))
            )
```
UMA requires an integer spin

ElliottKasoar:
This would almost certainly be an extension after merging this, but another neat thing would be in the periodic table colour bar/colours updated to match the current thresholds, and colour schemes.

If you'd be interested in looking into this at some point down the line I can open a separate issue to track it?

pgrigorev:
> Thanks for all the updates, @pgrigorev!
> 
> I think something may have gone wrong when you updated the branch, there seem to be a bunch of extra files related to the last few commits to be merged into main

Thanks a lot for having a look! I think I used 'merge' to sync with upstream and then tried 'rebase'. Think is not a good idea because it then treats merged commits as part of this branch and PR. I should stick to 'merge' for this PR to have more clean commit history. I will try to fix it asap.

ml_peg/calcs/bulk_crystal/equation_of_state/calc_equation_of_state.py
pgrigorev: Yes I think I should add 1 to the sum over magnetic moments. Thanks for catching that, I had to refresh my understanding about what spin multiplicity is.

pgrigorev:
> This would almost certainly be an extension after merging this, but another neat thing would be in the periodic table colour bar/colours updated to match the current thresholds, and colour schemes.
> 
> If you'd be interested in looking into this at some point down the line I can open a separate issue to track it?

yes sure!

ml_peg/calcs/bulk_crystal/energy_volume_curves_metals/calc_energy_volume_curves_metals.py
```diff
@@ -0,0 +1,325 @@
...
+
+    write_dir = OUT_PATH / model_name
+    write_dir.mkdir(parents=True, exist_ok=True)
+
+    calc = model.get_calculator(precision="high")
+
+    data_path = (
+        download_s3_data(
+            filename="energy_volume_curves_metals.zip",
+            key="inputs/bulk_crystal/energy_volume_curves_metals/energy_volume_curves_metals.zip",
```
pgrigorev, line 281: I think you will have to update the file on the s3. I have attached the renamed file to this comment. 

[energy_volume_curves_metals.zip](https://github.com/user-attachments/files/28175778/energy_volume_curves_metals.zip)

pgrigorev:
I had to cherry pick the commits on top of the current state of the main branch to avoid git picking extra commits and files. This makes a clean commit history, but might break the logic of previous review and comments. Sorry about that. 

I think now it should be ready to be merged :crossed_fingers:. Thank you!

ml_peg/calcs/bulk_crystal/energy_volume_curves_metals/calc_energy_volume_curves_metals.py
ElliottKasoar: Done, thanks!

ElliottKasoar (approved):
Thanks again for all the work, @pgrigorev!

I'll open a new issue to add a couple of the suggestions we discussed, but in the meantime I'll start running this for all the models, and hopefully can make those results live this week.

joehart2001:
Thanks for your work on this @pgrigorev !

PR #333, opened 2026-01-31
Graphene wetting benchmark

### LINKED ISSUE #292: Graphene Wetting Under Strain
### Summary

Single-point calculations of adsorption energy of single water molecule on graphene, with various orientations, distances, and tensile strain conditions.

### Interactive features

MAEs; possibly (with greater weights) also binding energy & equilibrium distance as function of strain

### Category

Surfaces

### Data availability

Data is described in upcoming publication ("Revealing Strain Effects on the Graphene-Water Contact Angle Using a Machine Learning Potential"). Calculations made at the revPBE-D3 level of theory using FHI-aims.

### Computational cost

Very low

### Additional context

_No response_

dwl38:

## Summary

Added a new benchmark based on the adsorption energy curves of a single water molecule (at various orientations) on a sheet of graphene (under various strain conditions), which is useful for understanding nanoscale wetting. Reference calculations based on PBE functional, calculated using FHI-aims on "intermediate" settings.

Three metrics of equal weight:

1. MAE of all single-point calculations (across all orientations, strains, and distances)
2. MAE of binding energies (across all orientations and strains), by comparing fitted adsorption energy curves
3. MAE of binding lengths (across all orientations and strains), from the same method

## Linked issue

Resolves #292

## Progress

- [x] Calculations
- [x] Analysis
- [x] Application
- [x] Documentation

## Testing

Benchmark tested on all currently-implemented models, i.e.:

- mace-mp-0a
- mace-mp-0b3
- mace-mpa-0
- mace-omat-0
- mace-matpes-r2scan
- orb-v3-consv-inf-omat
- pet-mad

with no issues. At current time of writing, mace-mp-0a performs the best on all metrics, whereas pet-mad completely fails to produce physically-reasonable adsorption energy curves.

## New decorators/callbacks

Added a new "plot_from_scatter" callback, which implements almost identical functionality to "struct_from_scatter" except that it renders a Plotly Graph object instead. (EDIT 14 Mar 2026: this is now unused.)

(EDIT 14 Mar 2026) Added a new "struct_from_multi_scatters" callback, which implements almost identical functionality to "struct_from_scatter" except that, instead of accepting a single traj file or a single list of struct files which is rendered based on the clicked point index, this callback accepts a list of traj files or a list of list of struct files which is rendered based on the clicked curve number and point index.


ml_peg/calcs/surfaces/graphene_wetting_under_strain/calc_graphene_wetting_under_strain.py
```diff
@@ -0,0 +1,94 @@
...
+            yaml.safe_dump(database_info, target_fp, sort_keys=False)
+        DATABASE_INFO_SAVED = True
+
+    # Calculate energy of single water molecule
+    atoms = ase.io.read(structs_dir / "ref_water.xyz", format="extxyz")
+    atoms.calc = calc
+    water_energy = atoms.get_potential_energy()
+
+    # Iterate through strain conditions
+    for strain in strains:
```
ElliottKasoar: It's fairly minor since it's so short, but it might be nice to use tqdm for this, just so we can see progress 

For slower models, it can take a minute or more, and since it's very reasonable to run this locally, that might be concerning

ml_peg/analysis/surfaces/graphene_wetting_under_strain/analyse_graphene_wetting_under_strain.py
```diff
@@ -0,0 +1,460 @@
...
+                    return {
+                        "ref": (
+                            results["distances"],
+                            results["ref"][orientation][strain]["energies"],
+                        ),
+                        model: (
+                            results["distances"],
+                            results[model][orientation][strain]["energies"],
+                        ),
+                    }
```
ElliottKasoar: I would probably move this and the other plots out of this function, probably just to a top-level function that is called here.

You may still need to wrap it in a function to parameterise the `plot_scatter` decorators, but it makes `processed_data` a bit less unwieldy

ElliottKasoar (commented):
Thanks for this, it's looking really nice! I've left a few mostly minor comments.

My main suggestions/questions revolve around the app callbacks. I think what do have works really well, but if possible it would be great to integrate any modifications you need into the existing helper functions we have.

Also, in terms of visualisation, I wonder how tricky it would be to view the structure as a trajectory vs distance, a bit like our live NEBs example - so if you click on a point, it shows the structure at that distance, but you can also play the trajectory as the distance changes.

ml_peg/calcs/surfaces/graphene_wetting_under_strain/calc_graphene_wetting_under_strain.py
dwl38: Yep, tqdm works like a charm

ml_peg/app/surfaces/graphene_wetting_under_strain/app_graphene_wetting_under_strain.py
```diff
@@ -0,0 +1,372 @@
...
+            Message explaining interactivity.
+        Div
+            Message explaining interactivity.
+        """
+        if not active_cell:
+            return (
+                Div("Click on a metric to view plot."),
+                Div("Click on a metric to view plot."),
+                Div("Click on a metric to view the structure."),
+            )
```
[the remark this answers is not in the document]
dwl38: Yeah, this was a bit weird, I've decided that there's no need to simultaneously view the parity plot & binding energy curves at the same time. See next comment.

I think the original logic was that, there are three different energy curves (0-leg, 1-leg, 2-leg) for each model, hence I wanted each data point in the parity plot to lead to a binding energy curve. And so both plots needed to be visible at the same time. I'm rethinking this now...

ml_peg/app/surfaces/graphene_wetting_under_strain/app_graphene_wetting_under_strain.py
```diff
@@ -0,0 +1,372 @@
...
+        return (
+            Div("Click on a metric to view plot."),
+            Div("Click on a metric to view plot."),
+            Div("Click on a metric to view the structure."),
+        )
+
+
+def plot_and_struct_from_scatter(scatter_id, plot_id, plots_list, struct_id, structs):
```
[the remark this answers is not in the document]
dwl38: In the original logic, I wanted two plots + one struct to all be visible at once when a data point was clicked (see comments below, I think this is a bit clunky so I'm rethinking it now). The other callbacks that I added had a problem where, because of the suppression of duplicate callbacks, it wasn't possible to simultaneously open a plot and a struct in one click.

ml_peg/app/utils/build_callbacks.py
```diff
@@ -130,6 +130,52 @@ def show_plot(active_cell, current_table_data) -> Div:
         return Div("Click on a metric to view plot.")
 
 
+def plot_from_scatter(
```
ElliottKasoar: Is this now unused?

ElliottKasoar:
Thanks for addressing so many comments, @dwl38! I've left a few more but they're mostly relatively minor, it's looking great!

I think it may have got lost somewhere but can you add a default spin (multiplicity) and charge to the structures. If you rebase, you can test against `orb-v3-consv-omol` to see if it's happy.

Rebasing will also solve the new conflict we have in the `surfaces.rst` documentation. This is just because the docs look very similar between tests, so git struggles to resolve this automatically. I can help if needed, but often I just end up copy and pasting the old file, and then adding the new contents below, since there's no real conflict.

ml_peg/calcs/surfaces/graphene_wetting_under_strain/calc_graphene_wetting_under_strain.py
```diff
@@ -0,0 +1,91 @@
...
+
+        # Iterate through orientations
+        for orientation in orientations:
+            systems = ase.io.iread(
+                structs_dir / f"{orientation}_{strain}.xyz", index=":", format="extxyz"
+            )
+            write_file = write_dir / f"{orientation}_{strain}.xyz"
+            if os.path.isfile(write_file):
+                os.remove(write_file)
+            for atoms in tqdm(systems):
```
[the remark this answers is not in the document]
dwl38: Fair enough, this was premature optimization at play -- in my own code I have a habit of using `iread` instead of `read` to save on memory when analyzing a 10k+ frames trajectory -- but completely irrelevant in this context of course.

Changed `iread` to `read`, which now returns a list of Atoms (and more importantly allows `tqdm` to figure out the count automatically). Also added a description for the orientation and strain of each calculation.

ml_peg/app/utils/build_callbacks.py
dwl38: Err, yes. Decided to leave it in, just in case anybody else might want to use it. (I think @joehart2001 said someone else does. Or did I misunderstand?)

ml_peg/app/utils/build_callbacks.py
```diff
@@ -195,6 +241,73 @@ def show_struct(click_data):
...
 
 
+def struct_from_multi_scatters(
+    scatter_id: str,
+    struct_id: str,
+    structs: list[str] | list[list[str]],
+    mode: Literal["struct", "traj"] = "struct",
+) -> None:
+    """
+    Attach callback to show a structure when a scatter point is clicked.
```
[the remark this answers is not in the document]
dwl38: Ah, didn't realise that! Indeed there used to be a much better documentation, but the linter annoyed me a bit too much with the "one-line" rule and I didn't know you could add a second paragraph.

Updated docstring accordingly.

ml_peg/app/utils/build_callbacks.py
ElliottKasoar: Yes, in general it's always a little risky adding code with intended future use, but I think there are other active PRs we could use this, so happy to keep it, thanks!

ml_peg/calcs/surfaces/graphene_wetting_under_strain/calc_graphene_wetting_under_strain.py
```diff
@@ -0,0 +1,94 @@
...
+        for orientation in orientations:
+            systems = ase.io.read(
+                structs_dir / f"{orientation}_{strain}.xyz", index=":", format="extxyz"
+            )
+            write_file = write_dir / f"{orientation}_{strain}.xyz"
+            if os.path.isfile(write_file):
+                os.remove(write_file)
+            desc = f"{orientation} orientation with {strain[1:5]}% strain"
+            for atoms in tqdm(systems, desc=desc, unit="configurations"):
+                atoms.calc = calc
```
ElliottKasoar, line 88:
```suggestion
                atoms.calc = calc
                atoms.info.setdefault("charge", 0)
                atoms.info.setdefault("spin", 1)
```
See my previous comment - these keys need to be set before we call `get_potential_energy` else we get an error from Orb-omol:

```
ValueError: atoms.info must contain both 'charge' and 'spin'
```

ElliottKasoar (commented):
Thanks for the updates, @dwl38!

There are just a couple of things to resolve and I think we're good to merge!

One is the issue that prevents us running Orb-omol, which I've added a code suggestion for now too, so hopefully is very easy to resolve.

The final thing is the question regarding the thresholds. I think you have sufficient context, so ultimately it's up to you. If you're happy with them, please resolve those comment threads.

ElliottKasoar:
> Oh, on a side note, I am planning to redo the reference calculations in future using a series of different DFT functionals, and perhaps also QMC; would that be of interest, to add to the dataset for you?

Sorry I'm not sure we replied to this, but yes, definitely!

In general, higher level theory comparisons are always welcome.

In the current form of the code, this would just be via extra columns (we have a few examples of this already) but in future we'd like multiple references to be a more standard thing.

This can of course come through future PRs, so we don't hold this one up.

ml_peg/calcs/surfaces/graphene_wetting_under_strain/calc_graphene_wetting_under_strain.py
dwl38: Thanks! I admit I didn't quite understand the error previously, so thanks for the fix

dwl38:
Regarding thresholds, I took Joe's advice and left the good/bad thresholds at 15/60 meV respectively, which as mentioned gives a score of 0 for all current models (fair enough). I'm happy to consider this resolved now.

(Sorry, I hadn't realized I actually need to click the "resolve" button to let the review go through!)

As for further reference calculations -- got distracted on this one, WIP -- for a future PR in that case :)

ElliottKasoar (approved):

PR #337, opened 2026-02-04
Add new category containing interstitial benchmarks

zhonganr:
Add a new category `interstitial` to assess the models' predictive performance for interstitial defect properties. Two benchmarks `FE1SIA` and `Relastab` are included:
- `FE1SIA` evaluates the formation energy of a single self-interstitial atom (SIA) in a host lattice for distinct configurations.
- `Relastab` evaluates the ability of models to correctly rank the stability of different interstitial configurations.
Related to #339


ml_peg/calcs/interstitial/Relastab/calc_Relastab.py
```diff
@@ -0,0 +1,67 @@
...
+    Parameters
+    ----------
+    mlip
+        Name of model use and model to get calculator.
+    """
+    model_name, model = mlip
+    calc = model.get_calculator()
+
+    data_path = download_s3_data(
+        key="inputs/interstitial/relative_stability/DB.zip",
```
ElliottKasoar: See above

ElliottKasoar:
Thanks for adding this and sharing the data! It's looking great so far!

.gitignore
```diff
@@ -41,3 +41,5 @@ __pycache__/
 ml_peg/app/data/*
 !ml_peg/app/data/onboarding/
 certs/
+DB/
+DB.zip
```
ElliottKasoar: Do we still need these?

ml_peg/analysis/defect/Defectstab/analyse_Defectstab.py
```diff
@@ -0,0 +1,338 @@
...
+
+from ml_peg.analysis.utils.decorators import build_table, plot_parity
+from ml_peg.analysis.utils.utils import load_metrics_config, rmse
+from ml_peg.app import APP_ROOT
+from ml_peg.calcs import CALCS_ROOT
+from ml_peg.models.get_models import get_model_names
+from ml_peg.models.models import current_models
+
+MODELS = get_model_names(current_models)
+D3_MODEL_NAMES = {m: m for m in MODELS}
```
ElliottKasoar:
```suggestion
```
I don't think this is needed since there's no dispersion correction

ElliottKasoar (commented):
Thanks for addressing so much, @zhonganr, it's looking great!

I've left a few more minor questions/suggestions.

Also note that there are a few conflicts to be resolved with the main branch. Can you try rebasing? We are happy to help with this if you need it.

ml_peg/analysis/defect/Defectstab/metrics.yml
```diff
@@ -0,0 +1,7 @@
+metrics:
+    RMSD:
+        good: 0.0
+        bad: 10.0
+        unit: eV
```
ElliottKasoar: Apologies if I already asked this elsewhere, but just to check, are these definitely sensible thresholds (the same question for Relastab)?

Generally the "good" threshold should take into account any inherent errors e.g. uncertainty if you used the reference method, while the "bad" threshold should be a value where the model is essentially useless. 

This doesn't have to give a good spread of scores between 0 and 1 - it's perfectly valid for all models to get a score of 0, if they're all in fact currently unable to perform a task in useful way!

10 eV seems like quite a lot, but if something on this order is still of some use, that's fine. This is just to check!

ml_peg/app/defect/Defectstab/app_Defectstab.py
```diff
@@ -0,0 +1,165 @@
...
+            tuple[list[dict], list[dict]]
+                Table data and conditional styles.
+            """
+            if subset_value == "total":
+                filename = "defectstab_metrics_table.json"
+            else:
+                filename = f"defectstab_metrics_table_{subset_value}.json"
+
+            try:
+                new_table = rebuild_table(
```
ElliottKasoar: This is really cool, but unfortunately does not propagate to the category table. Either @joehart2001 or I can try to have a look at this to get it working if you need a hand.

ml_peg/analysis/defect/Relastab/analyse_Relastab.py
```diff
@@ -0,0 +1,342 @@
...
+@plot_parity(
+    filename=OUT_PATH / "figure_energy.json",
+    title="Relastab Energies (Shifted per subset)",
+    x_label="Predicted Energy (Shifted) / eV",
+    y_label="Reference Energy (Shifted) / eV",
+    hoverdata={
+        "System": get_system_names(),
+    },
+)
+def stability_energies(grouped_data) -> dict[str, list]:
```
ElliottKasoar, line 135: I think this needs to be used similarly to `formation_energies` in Defectstab, else the plot doesn't appear, see my code suggestion below.

ml_peg/app/defect/Defectstab/app_Defectstab.py
```diff
@@ -0,0 +1,165 @@
...
+            DATA_PATH / "figure_energy.json",
+            id=f"{BENCHMARK_NAME}-figure",
+        )
+
+        # Assets dir will be parent directory - individual files for each system
+        # Assuming the first model has all structures
+        structs_dir = DATA_PATH / MODELS[0]
+        # Sort files to match the order in the scatter plot
+        structs = [
+            f"assets/defect/Defectstab/{MODELS[0]}/{struct_file.name}"
```
ElliottKasoar:
```suggestion
            f"/assets/defect/Defectstab/{MODELS[0]}/{struct_file.name}"
```
Due to some internal code changes

ml_peg/app/defect/Relastab/app_Relastab.py
```diff
@@ -0,0 +1,168 @@
...
+        scatter = read_plot(
+            DATA_PATH / "figure_energy.json",
+            id=f"{BENCHMARK_NAME}-figure",
+        )
+
+        # Assuming first model has structure files
+        structs_dir = DATA_PATH / MODELS[0]
+        # Sort glob to match analysis order
+        structs = [
+            f"assets/defect/Relastab/{MODELS[0]}/{struct_path.name}"
```
ElliottKasoar:
```suggestion
            f"/assets/defect/Relastab/{MODELS[0]}/{struct_path.name}"
```
Due to some internal code changes

ml_peg/analysis/defect/Relastab/analyse_Relastab.py
```diff
@@ -0,0 +1,342 @@
...
+
+
+@pytest.fixture
+@plot_parity(
+    filename=OUT_PATH / "figure_energy.json",
+    title="Relastab Energies (Shifted per subset)",
+    x_label="Predicted Energy (Shifted) / eV",
+    y_label="Reference Energy (Shifted) / eV",
+    hoverdata={
+        "System": get_system_names(),
```
ElliottKasoar, line 131: This is not essential, but given that you have filtering by subset, perhaps the points could be labeled with their subset?

ml_peg/analysis/defect/Defectstab/analyse_Defectstab.py
```diff
@@ -0,0 +1,340 @@
...
+
+
+@pytest.fixture
+@plot_parity(
+    filename=OUT_PATH / "figure_energy.json",
+    title="Defectstab Formation Energies",
+    x_label="Predicted Formation Energy / eV",
+    y_label="Reference Formation Energy / eV",
+    hoverdata={
+        "System": get_system_names(),
```
ElliottKasoar, line 198: This is not essential, but given that you have filtering by subset, perhaps the points could be labeled with their subset?

ElliottKasoar (commented):
Thanks for responding to all the comments! It's looking really good!

Just a few more minor suggestions/questions and I think we'll be good to merge.

ml_peg/analysis/defect/Defectstab/metrics.yml
zhonganr: Thanks for this comment! You're right, 10 eV is quite large. However, the variation in RMSD across different subsets is also very large. For example, in boroncarbide_stoichiometry, the RMSD can reach around 17 eV. 

The issue is that RMSD values strongly depend on the formation energies themselves, which vary significantly between systems (subsets). Because of this, it's difficult to define a universally meaningful threshold for what should be considered "bad". Since the "bad" value is currently defined globally in metrics.yml, I initially avoided introducing subset-dependent thresholds to keep the code structure simple. 

My thinking was that users could adjust this value directly in the UI, especially since you've made it easy to do so :) I'm still unsure whether it would be better to keep it this way or move to subset-specific values. Do you have any suggestions on this?

ElliottKasoar: I see, thanks for the explanation (and kind words)!

So, the simplest option would be to have individual columns per subset, and then you can define a different threshold for each of those in the `metrics.yml`, along the lines of what we have for GMTKN55 if you take a look at the table there: https://ml-peg.stfc.ac.uk/category/molecular-systems (note: in these cases we typically only have one non-zero weight, usually the total, by default).

Multiple columns would also allow you to, for example, include several subsets, but not all of them, through adjusting the different weights that would become available.

This would also remove the need to resolve the need to fix the score propagation I mentioned.

Given all of these benefits, I think this would be my suggestion. It does add more complexity to the table, especially Relastab, where you'd have two sets for everything, but it is at least simple and flexible.

The alternative, if you wanted to keep the drop down and column that swaps out, is probably still defining multiple columns with unique thresholds in `metrics.yml`, and finding a way to apply these as part of the dropdown callback to the corresponding input value for the thresholds.

@joehart2001 any others thoughts?

ElliottKasoar (commented):
I think other than the discussion around the metrics/thresholds, this is pretty much good to go!

One other note: I think your documentation was deleted during one of your commits. If you could add back this back in (and update it following whatever decision we make regarding metrics) that would be great!

ml_peg/analysis/defect/Defectstab/metrics.yml
joehart2001: I agree with this. Fixing propagation issues is often quite a tedious job so its probably best to keep it simple here. I think its quite nice to be able to see all the data in the table at a glance too

zhonganr: Thank you so much for your helpful suggestions! 

I've created per-subset "bad" thresholds for Defectstab and different metric columns in the app tables . 

For Relastab, I think we can keep 0 and 1 as the "bad" and "good" thresholds. For GlobalMin, it's simply a matter of whether a model is able to find the most stable configuration or not. For Top5_Spearman, it's an ordering problem, so 1 means perfect ordering and 0 means random ordering (while we could get negative values for inverse ordering, I think random is already bad enough :))

docs/source/user_guide/benchmarks/defect.rst
```diff
@@ -0,0 +1,229 @@
...
+
+* Subset ``cawo4`` of the ``Relastab`` dataset (:math:`\mathrm{CaWO_4}` interstitial configurations).
+
+  * G. Soum-Sidikov, A. Boisard, D. L010, M. Loidl, O. Stézowski, A. Music, G. Music, Y. Zeng, R. Cong, T. Yang, A. Echeverria, L. Music, D. Music, V. Music,
+    Calculation of crystal defects induced in :math:`\mathrm{CaWO_4}` cryogenic detector by 100 eV displacement cascades using a data driven force field,
+    Phys. Rev. D 111, 085021 (2025).
+
+Reference data:
+
+* Computed from the DFT total energies provided with the input structures.
```
ElliottKasoar, line 224: I know you said it above but it would be good to reiterate PBE is the functional here, so it's easy for people to know the exact level of theory we're comparing to at a glance, including potential functional mismatches with training data

ml_peg/analysis/defect/Defectstab/metrics.yml
```diff
@@ -0,0 +1,36 @@
...
+        level_of_theory: DFT
+        weight: 0
+    boroncarbide_defects:
+        good: 0.0
+        bad: 0.57
+        unit: eV
+        tooltip: "RMSD of formation energies for boron carbide defects subset"
+        level_of_theory: DFT
+        weight: 0
+    mapi_tetragonal:
```
ElliottKasoar: These names could be slightly more human-friendly e.g. spaces are perfectly valid here

ml_peg/analysis/defect/Defectstab/metrics.yml
```diff
@@ -0,0 +1,36 @@
...
+        good: 0.0
+        bad: 0.39
+        unit: eV
+        tooltip: "RMSD of formation energies for MAPI tetragonal subset"
+        level_of_theory: DFT
```
ElliottKasoar:
```suggestion
        level_of_theory: PBE
        weight: 1
    fe_sia:
        good: 0.0
        bad: 1.32
        unit: eV
        tooltip: "RMSD of formation energies for Fe SIA subset"
        level_of_theory: PBE
        weight: 0
    boroncarbide_stoichiometry:
        good: 0.0
        bad: 11.09
        unit: eV
        tooltip: "RMSD of formation energies for boron carbide stoichiometry subset"
        level_of_theory: PBE
        weight: 0
    boroncarbide_defects:
        good: 0.0
        bad: 0.57
        unit: eV
        tooltip: "RMSD of formation energies for boron carbide defects subset"
        level_of_theory: PBE
        weight: 0
    mapi_tetragonal:
        good: 0.0
        bad: 0.39
        unit: eV
        tooltip: "RMSD of formation energies for MAPI tetragonal subset"
        level_of_theory: PBE
```
Generally we try to give the functional, since it's useful to know when a model has been trained on data using the same functional or not

ml_peg/app/defect/Defectstab/app_Defectstab.py
```diff
@@ -0,0 +1,90 @@
...
+        # Sort files to match the order in the scatter plot
+        structs = [
+            f"/assets/defect/Defectstab/{MODELS[0]}/{struct_file.name}"
+            for struct_file in sorted(structs_dir.glob("*.xyz"))
+        ]
+
+        plot_from_table_column(
+            table_id=self.table_id,
+            plot_id=f"{BENCHMARK_NAME}-figure-placeholder",
+            column_to_plot={"RMSD": scatter},
```
ElliottKasoar: It's not strictly necessary, but it may be worth adding the other columns: scatter in this dict as well, so any of the cells will show the plot?

ml_peg/analysis/defect/Relastab/metrics.yml
```diff
@@ -0,0 +1,43 @@
...
+        unit: null
+        tooltip: "GlobalMin for CaWO4 subset"
+        level_of_theory: DFT
+        weight: 0
+    Top5_Spearman_cawo4:
+        good: 1.0
+        bad: 0.0
+        unit: null
+        tooltip: "Top5_Spearman for CaWO4 subset"
+        level_of_theory: DFT
```
ElliottKasoar:
```suggestion
        level_of_theory: PBE
        weight: 1
    Top5_Spearman:
        good: 1.0
        bad: 0.0
        unit: null
        tooltip: "Spearman rank correlation of the 5 highest (least stable) energy configurations"
        level_of_theory: PBE
        weight: 1
    GlobalMin_fe:
        good: 1.0
        bad: 0.0
        unit: null
        tooltip: "GlobalMin for Fe subset"
        level_of_theory: PBE
        weight: 0
    Top5_Spearman_fe:
        good: 1.0
        bad: 0.0
        unit: null
        tooltip: "Top5_Spearman for Fe subset"
        level_of_theory: PBE
        weight: 0
    GlobalMin_cawo4:
        good: 1.0
        bad: 0.0
        unit: null
        tooltip: "GlobalMin for CaWO4 subset"
        level_of_theory: PBE
        weight: 0
    Top5_Spearman_cawo4:
        good: 1.0
        bad: 0.0
        unit: null
        tooltip: "Top5_Spearman for CaWO4 subset"
        level_of_theory: PBE
```
As above

ml_peg/analysis/defect/Relastab/metrics.yml
ElliottKasoar, line 1: As above, I'd suggest these metric names could be more human readable e.g. with spaces

ml_peg/app/defect/Relastab/app_Relastab.py
```diff
@@ -0,0 +1,88 @@
...
+        # Sort glob to match analysis order
+        structs = [
+            f"/assets/defect/Relastab/{MODELS[0]}/{struct_path.name}"
+            for struct_path in sorted(structs_dir.glob("*.xyz"))
+        ]
+
+        plot_from_table_column(
+            table_id=self.table_id,
+            plot_id=f"{BENCHMARK_NAME}-figure-placeholder",
+            column_to_plot={"GlobalMin": scatter, "Top5_Spearman": scatter},
```
ElliottKasoar: As above, you may want to add the other columns as valid options to show the scatter too

ElliottKasoar (commented):
Thanks for addressing the suggestions so quickly, @zhonganr!

I just spotted a few final minor things we may want to tidy, but otherwise I think we'll probably be good to merge!

ml_peg/analysis/defect/Defectstab/metrics.yml
zhonganr: Thanks for the remark, indeed it's better.
The functionals used in Defectstab are not all PBE, but I've specified the corresponding functional for each subset.

ElliottKasoar: Ah I see, sorry I missed that, and thanks for fixing it!

Do you think it's still meaningful to combine them as an average then?

docs/source/user_guide/benchmarks/defect.rst
```diff
@@ -0,0 +1,231 @@
...
+    Challenges in predicting positron annihilation lifetimes in lead halide perovskites: correlation functionals and polymorphism,
+    arXiv:2511.06926 (2025).
+
+  * K. Madaan,
+    Phases and vacancy defects in methylammonium lead iodide perovskite: an ab initio study,
+    PhD thesis, Université Paris-Saclay (2023).
+
+Reference data:
+
+* Computed from the DFT total energies provided with the input structures.
```
ElliottKasoar, line 157: Could you add a note here either restating the functionals again or noting that the specific functionals are specified with the metrics above?

docs/source/user_guide/benchmarks/defect.rst
```diff
@@ -0,0 +1,231 @@
...
+
+* Subset ``cawo4`` of the ``Relastab`` dataset (:math:`\mathrm{CaWO_4}` interstitial configurations).
+
+  * G. Soum-Sidikov, A. Boisard, D. L010, M. Loidl, O. Stézowski, A. Music, G. Music, Y. Zeng, R. Cong, T. Yang, A. Echeverria, L. Music, D. Music, V. Music,
+    Calculation of crystal defects induced in :math:`\mathrm{CaWO_4}` cryogenic detector by 100 eV displacement cascades using a data driven force field,
+    Phys. Rev. D 111, 085021 (2025).
+
+Reference data:
+
+* Computed from the DFT total energies provided with the input structures.
```
ElliottKasoar, line 224: Could you add a note here either restating the different functionals again or noting that the specific functionals are specified with the metrics above?

ml_peg/analysis/defect/Defectstab/metrics.yml
zhonganr: You're right, I've removed the overall metrics values and just keep the final scores.

ElliottKasoar (approved):

ElliottKasoar:
Thank you so much @zhonganr for all the work! I think we're good to merge.

I'll start running these for the live site with all of the models and let you know if anything else pops up

PR #325, opened 2026-01-30
Added 2D TMDs NEB benchmark and docs updates

Jad-J3far:
## Pre-review checklist for PR author

- [x] I've confirmed the [contribution guidelines](https://github.com/ddmms/ml-peg/blob/main/contributing.md).

## Summary

Adds a new NEB benchmark for oxygen adatom diffusion on 2D TMDs (MoS2, MoSe2, MoTe2, WS2, WSe2, WTe2).  
Includes calculation workflows, analysis metrics/tables, and the Dash app entry, plus user guide documentation and model registry updates.

## Linked issue

Resolves #303

## Progress

- [x] Calculations
- [x] Analysis
- [x] Application
- [x] Documentation

## Testing

Tested on CUDA with: `mace-mp-0b3`, `orb-v3-consv-inf-omat`, `uma-s-1p1-omat`

## New decorators/callbacks

None.

PR #350, opened 2026-02-06
NEB surface reaction

### LINKED ISSUE #293: NEB for surface reactions with OC20NEB Dataset
### Summary

Benchmark for surface reactions using NEB using subset of OC20NEB dataset 

Procedure : 
1. Select reactions from three reaction classes (adsorption/desorption, dissociation, transfer)
2. Relax Initial and final geometries
3. Generate interpolation (Same number of images + Linear interpolation + IDPP relaxation)
4. Run NEB 

(tentative) Metrics : 
- Convergence
- Error with respect to the DFT reference 
- Sanity of NEB result in terms of connectivity 

### Interactive features

NEB profile 

### Category

NEB 

### Data availability

[OC20NEB dataset](https://pubs.acs.org/doi/full/10.1021/acscatal.4c04272?casa_token=vNXIpVKIeSgAAAAA%3A6QTbniklLX3giFAzjkQD-djpmIGsqYrph6JOJah_T3VZSmPEwogoSW9xa0YQiBMJlRZ47P91k5EBejpO) 

DFT functional : RPBE 

### Computational cost

medium or long 

### Additional context

_No response_

jungsdao:

## Summary

Add NEB benchmark for three surface reactions from OC20NEB dataset comparing barrier height errors.  

## Linked issue

Resolves #293

## Progress

- [x] Calculations
- [x] Analysis
- [x] Application
- [x] Documentation

## Testing
Test on GPU with
- mace-mp-0a
- mace-mp-0b3
- mace-mpa-0
- mace-omat-0
- mace-matpes-r2scan
- orb-v3-consv-inf-omat
- pet-mad

## New decorators/callbacks
None


ElliottKasoar (commented):
Thanks for this, @jungsdao!

Apologies for the slow review.

I'd be inclined to rename the test something slightly more specific e.g. OC20NEB, since this name can't clash with any other test.

I'd also like to explore replacing a lot of the calculation with `janus-core`'s NEB calculation, as it's a lot of duplication, and I think the only thing missing is a way of checking convergence/continuing, both of which should be minor additions.

ml_peg/calcs/nebs/surface_reaction/calc_surface_reaction.py
```diff
@@ -0,0 +1,143 @@
...
+    for reaction in REACTIONS:
+        print(f"{reaction} {model_name}, start NEB ...")
+        neb = NEB(
+            init_struct=relaxed_structs[f"{reaction}-{model_name}"][0],
+            final_struct=relaxed_structs[f"{reaction}-{model_name}"][-1],
+			neb_class="DyNEB",
+            n_images=8,
+            neb_kwargs = {"climb": False, "method": "eb", "scale_fmax": 1.0},
+            interpolator="ase",
+            minimize=False,
```
ElliottKasoar: Is there a benefit to doing the minimisation separately to this? I think in the Li diffusion example, we do it because one of the endpoints is shared between two different tests, but since that's not the case here, I think it may overcomplicate things?

ml_peg/calcs/nebs/surface_reaction/calc_surface_reaction.py
```diff
@@ -0,0 +1,143 @@
...
+        neb.interpolator = None
+        neb.run(fmax=0.45, steps=200)
+
+        neb.neb.climb = True
+        neb.plot_band = True
+        neb.write_band = True
+        neb.run(fmax=0.05, steps=300)
+
+        forces = neb.neb.get_forces()
+#        print(f"{np.sqrt((forces ** 2).sum(axis=1).max()) = }")
```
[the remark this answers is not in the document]
jungsdao: I'll remove this in the final version currently It's not cleaned up yet.

ml_peg/calcs/nebs/OC20NEB/calc_OC20NEB.py
```diff
@@ -0,0 +1,143 @@
...
+        neb.write_band = True
+        neb.run(fmax=0.05, steps=300)
+
+        forces = neb.neb.get_forces()
+#        print(f"{np.sqrt((forces ** 2).sum(axis=1).max()) = }")
+        neb.results["max_force"] = np.sqrt((forces ** 2).sum(axis=1).max())
+        if neb.write_results:
+            with open(neb.results_file, "w", encoding="utf8") as out:
+                print("#Barrier [eV] | delta E [eV] | Max force [eV/Å] ", file=out)
+                print(*neb.results.values(), file=out)
```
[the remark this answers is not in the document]
ElliottKasoar: Ah good point, I will fix this in a future version of `janus-core`, thanks for spotting this!

ElliottKasoar (commented):
Hi @jungsdao, thanks again for all the work that's going into this!

I hope you don't mind that I've rebased and pushed a few suggested changes:

- Separate model directories for the calc outputs - the Li diffusion NEB is a bad example as generally it makes combining data simpler if we keep things in their own directories
- `oc20neb` -> `OC20NEB` for consistency with other tests e.g. `OC157`, and fixed the calculation file, which was still unchanged
-  Update the `janus-core` dependency and a couple of fixes around this, as a couple of the things you use from the `NEB` were only added recently
- Added spin/charge defaults, as these are required by Orb-omol
- Updated the analysis to use the new model directories, and (I think) fixed a couple of issues with paths that were wrong 
- Re-saved the DFT reference data, so we don't need to refer to the original source. This is some duplication, but it keeps things isolated

I also "removed" the docs in a [commit](https://github.com/ddmms/ml-peg/pull/350/changes/507bc46d6eb7553e06310c49e79749f3d097661c). This was to be consistent with your code changes, but I'm not sure if you'd removed it unintentionally? It would definitely be good to add an updated form back in!

Please feel free to revert/ask about any of these changes. The only other question is whether you're happy with the current visualisation, or if it might be preferable to have some sort of structure visualisation, closer to the Li diffusion, as well as #325 and #326?

I'm not sure if we currently have any examples where we include the NEB of the model and reference together, but I think we probably could, and that would save us loading quite a large image, and allow us to link those to their actual structures. What do you think?

jungsdao:
Hi @ElliottKasoar, Thanks a lot for helping with improving the code. I'm not disagreeing on your changes and thanks for pointing out the issues I have missed. 
Also I think in current S3bucket it has only three reactions but I shared more reactions via slack chat.

I think it would be nice if we can visualize geometry along the NEB trajectory like other nebs benchmarks, but I'm not sure how it can be done.  

```
I'm not sure if we currently have any examples where we include the NEB of the model and reference together, but I think we probably could, and that would save us loading quite a large image, and allow us to link those to their actual structures. What do you think?
```
I'm not quite understood what you meant with this comment. Do you mean showing RPBE reference NEB profile with model NEB profile? And showing structure altogether?

joehart2001:
> Hi @ElliottKasoar, Thanks a lot for helping with improving the code. I'm not disagreeing on your changes and thanks for pointing out the issues I have missed. Also I think in current S3bucket it has only three reactions but I shared more reactions via slack chat.
> 
> I think it would be nice if we can visualize geometry along the NEB trajectory like other nebs benchmarks, but I'm not sure how it can be done.
> 
> ```
> I'm not sure if we currently have any examples where we include the NEB of the model and reference together, but I think we probably could, and that would save us loading quite a large image, and allow us to link those to their actual structures. What do you think?
> ```
> 
> I'm not quite understood what you meant with this comment. Do you mean showing RPBE reference NEB profile with model NEB profile? And showing structure altogether?

Hi @jungsdao, (@ElliottKasoar  correct me if im misunderstanding), but i think what he means is we use the plotly decorators to show an interactive NEB plot with both the reference NEB profile and the MLIP NEB profile, instead of plotting these before and loading an image. I think the idea here could be to click on the NEB profile to visualise the structures at each step (as we do in the other NEB benchmark), but also be able to view the reference structures too, so one could compare the evolution of structures of the reference and MLIP. I think this would be pretty cool.

jungsdao:
Sorry for belated reply. I understand your suggestion and it should be definitely cool to show structures together. But I don't know what would be suitable way to implement it. It would be cascade of interactive calling: 1) scatter 2) NEB profile 3) structure for each point in NEB profile. Do you have any suggestion how to make it work? There seems to be many functions in `ml_peg/app/utils/build_callbacks.py` but should I define a new one or do you think it can be done by using what's already there?

ElliottKasoar:
> Sorry for belated reply. I understand your suggestion and it should be definitely cool to show structures together. But I don't know what would be suitable way to implement it. It would be cascade of interactive calling: 1) scatter 2) NEB profile 3) structure for each point in NEB profile. Do you have any suggestion how to make it work? There seems to be many functions in `ml_peg/app/utils/build_callbacks.py` but should I define a new one or do you think it can be done by using what's already there?

Hi @jungsdao, apologies for the very delayed reply on our end too! I've been away for all of June, and we've recently introduced a few changes that I wanted to be live before following up too much, to make sure everything is in-sync.

Can you also rebase and make sure the tests/pre-commit etc. is passing?

I'll do another proper review of everything this week.

jungsdao:
@ElliottKasoar Thanks for reviewing! I'll rebase and push again. Also I think it would be better to change and add some more reactions than what is already in S3 link. How can I share the reactions again?

ElliottKasoar:
> @ElliottKasoar Thanks for reviewing! I'll rebase and push again. Also I think it would be better to change and add some more reactions than what is already in S3 link. How can I share the reactions again?

Great, thanks! However is easiest, with the exception of as files in the PR itself ideally. You can upload the zip in this PR as part of a comment, on Slack, Dropbox/OneDrive, or email me (ek440@cam.ac.uk).

ml_peg/calcs/nebs/OC20NEB/calc_OC20NEB.py
```diff
@@ -0,0 +1,80 @@
...
+    for reaction in reactions:
+        print(f"{reaction} {model_name}, start NEB ...")
+        dft_traj = read(data_path / f"{reaction}.xyz", ":")
+        initial, final = dft_traj[0], dft_traj[-1]
+        for struct in [initial, final]:
+            struct.calc = calc.get_calculator()
+            struct.info.setdefault("spin", 1)
+            struct.info.setdefault("charge", 0)
+
+        neb = NEB(
```
ElliottKasoar, line 54: You may want to consider tightening the max steps for the minimisation as well, just in case a model explodes and doesn't converge. I think janus_core's default is 1000 for geom_opt (which gets called by the NEB minimise), which is much more reasonable than ASE's, but still quite large.

It's fine to leave it too, just a thought

ml_peg/analysis/nebs/OC20NEB/metrics.yml
ElliottKasoar, line 1: Apologies if I've already asked this, but how much have you thought about these thresholds/do you understand what they should represent?

If you're happy with them, great! If not, we're happy to discuss further.

ml_peg/app/nebs/OC20NEB/app_OC20NEB.py
```diff
@@ -0,0 +1,299 @@
...
+        metric_handler = partial(
+            build_serialized_scatter_content,
+            models_data=models_data,
+            label_map=label_to_key,
+            scatter_id=SCATTER_GRAPH_ID,
+            instructions=refresh_msg,
+        )
+
+        # ── Table cell → DFT-vs-MLIP scatter plot ────────────────────────────
+        scatter_and_assets_from_table(
```
ElliottKasoar, line 102: @joehart2001 may also be more familiar with this, since he wrote this function, but do we actually want/need (based on the structure visualisation) individual scatter plots for each model?

If we were doing all 932 NEBS from OC20NEB we'd probably want individual plots, but currently it's not even a handful of points.

In terms of the structure visualisation, this also seems to be using data from the calculations, similar to the phonon plotting, but in most cases we want these structures to be saved in the app data, which makes the process more modular.

ml_peg/app/nebs/OC20NEB/interactive_helpers.py
ElliottKasoar, line 1: Generally we don't want to be constructing these at the point we run the app, but instead build them ahead of time during the analysis, and just load them as part of the app.

ElliottKasoar (commented):
Hi @jungsdao, a couple of other quick questions in addition to the above comments:

I know you said you wanted to expand the number of reactions, but what is the reason to limit this number in general?

For the visualisation, I'd think we can simplify things a bit. In general there are already helper functions that will render plots from column/cell clicks (`plot_from_table_column`/`plot_from_table_cell`), plots from scatters (`plot_from_scatter`), and structures from scatters (`struct_from_scatter`, although perhaps not as readily the dual structures you currently have), so I think we reuse a lot of this, and reduce the custom changes here.

Please do make sure you look at the filtering mentioned above too. We need to write an info.json file during analysis to enable element filtering.

Can you also ensure the CI/pre-commit etc all passes, and the conflicting file is resolved?

jungsdao:
> > @ElliottKasoar Thanks for reviewing! I'll rebase and push again. Also I think it would be better to change and add some more reactions than what is already in S3 link. How can I share the reactions again?
> 
> Great, thanks! However is easiest, with the exception of as files in the PR itself ideally. You can upload the zip in this PR as part of a comment, on Slack, Dropbox/OneDrive, or email me ([ek440@cam.ac.uk](mailto:ek440@cam.ac.uk)).

Hi Sorry for also delayed reply. This is compressed file for 927 DFT NEB trajectories. If computational resource is available, whole trajectories can be used. 

What do you think about having separate `interactive_helpers.py`? It was inspired by bulk_crystal/phonons and I had no idea how to enable cascade of plotting (scatter -> NEB profile -> Two geometries from MLIP and DFT) 
Let me know whether it makes sense to keep current API. 

[oc20neb.tar.gz](https://github.com/user-attachments/files/29967343/oc20neb.tar.gz)

joehart2001:
Hi, just a note, you seem to have committed changes to many files by accident

jungsdao:
> Hi, just a note, you seem to have committed changes to many files by accident

Sorry for messy commits. I think I made mistake before but hopefully it's cleaned up by now.

jungsdao:
I think I have addressed most of comments and benchmark seems to work in app as well so I'll be happy to have your review on this PR.  
Probably some details need to be changed if there's larger number of NEB calculations.

ml_peg/analysis/nebs/OC20NEB/analyse_OC20NEB.py
```diff
@@ -0,0 +1,429 @@
...
+
+    for reaction in ref_data.keys():
+        ref_traj_path = CALC_PATH / f"{reaction}-dft.xyz"
+        ref_profile = ref_data[reaction]["energy"]
+
+        write(dft_assets_dir / ref_traj_path.name, ref_data[reaction]["traj"])
+        ref_cache[reaction] = {
+            "profile": ref_profile,
+            "traj_path": ref_traj_path,
+            "traj_url": f"{ASSETS_URL_PREFIX}/DFT/{ref_traj_path.name}",
```
ElliottKasoar: We don't normally need to use `/assets` in analysis, because when the app is launched, we set the `assets_folder` to the `ml_peg/app/data` directory

ml_peg/app/nebs/OC20NEB/interactive_helpers.py
```diff
@@ -0,0 +1,191 @@
...
+    ref_path = click_context.get("ref_profile")
+    pred_path = click_context.get("pred_profile")
+    image_index = click_context.get("image_index", 0)
+    system_label = click_context.get("system_label", "")
+
+    if not ref_path or not pred_path:
+        return None
+
+    iframe_style = {
+        "height": "450px",
```
joehart2001:
```suggestion
        "height": "550px",
```

joehart2001: benchmark is looking great and this is my only commnet. currently when you click on a neb image point the weas box renders with a scroll bar and you cant see that its a traj initiallly. this larger height mostly fixes this

jungsdao:
Hi is there any other changes that need to be done for this benchmark? Please let me know if there is anything :)

ml_peg/analysis/nebs/OC20NEB/analyse_OC20NEB.py
```diff
@@ -0,0 +1,420 @@
...
+DEFAULT_THRESHOLDS, DEFAULT_TOOLTIPS, DEFAULT_WEIGHTS = load_metrics_config(
+    METRICS_CONFIG_PATH
+)
+
+METRIC_LABELS = {
+    "delta_E": "Reaction Energy MAE",
+    "barrier": "Barrier MAE",
+    "fmax": "Unconverged Percentage",
+}
+FMAX = 0.05
```
ElliottKasoar: Can I suggest we set this as a constant (and use it) in the calc, and import it here? Then if we ever change it, we remain consistent

ml_peg/analysis/nebs/OC20NEB/analyse_OC20NEB.py
```diff
@@ -0,0 +1,420 @@
...
+    Run OC20NEB test.
+
+    Parameters
+    ----------
+    metrics
+        All OC20NEB metrics.
+    interactive_dataset
+        Scatter metadata produced by the ``interactive_dataset`` fixture.
+    """
+    write_struct_info(
```
ElliottKasoar: In this case I'd consider `get_struct_info`, since this writes a single set for all elements, and in principle it would be useful in future to be able to filter out specific NEBs, which `get_struct_info` would enable, since it saves the elements per structure.

ml_peg/analysis/nebs/OC20NEB/analyse_OC20NEB.py
ElliottKasoar, line 1: Can you make sure this works if data is missing/incomplete? For example, if I haven't run mace-mp-0a, but run `ml_peg analyse --test OC20NEB` for all models, I get


```
    @pytest.fixture
    def oc20neb_stats() -> dict[str, dict[str, float]]:
        """
        Get error in energy barrier for all reactions.
    
        Returns
        -------
        dict[str, float]
            Dictionary of predicted barrier errors for all models.
        """
        ref_data = _get_ref_data()
        ref_cache: dict[str, dict[str, Any]] = {}
    
        dft_assets_dir = OUT_PATH / "DFT"
        dft_assets_dir.mkdir(parents=True, exist_ok=True)
    
        for reaction in ref_data.keys():
            ref_traj_path = CALC_PATH / f"{reaction}-dft.xyz"
            ref_profile = ref_data[reaction]["energy"]
    
            write(dft_assets_dir / ref_traj_path.name, ref_data[reaction]["traj"])
            ref_cache[reaction] = {
                "profile": ref_profile,
                "traj_path": ref_traj_path,
            }
    
        stats: dict[str, dict[str, Any]] = {}
    
        for model_name in MODELS:
            metrics_data: dict[str, dict[str, Any]] = {
                key: {"points": [], "ref": [], "pred": [], "mae": None}
                for key in METRIC_LABELS.keys()
            }
            model_assets_dir = OUT_PATH / model_name
            model_assets_dir.mkdir(parents=True, exist_ok=True)
    
            for reaction in ref_data.keys():
>               with open(
                    CALC_PATH / model_name / f"{reaction}-neb-results.dat", encoding="utf8"
                ) as f:
E               FileNotFoundError: [Errno 2] No such file or directory: '/Users/elliottkasoar/Documents/PSDI/ml-peg-copy/ml_peg/calcs/nebs/OC20NEB/outputs/mace-mp-0a/transfer_id_601_1482_1_211-5-neb-results.dat'
```

ml_peg/calcs/nebs/OC20NEB/calc_OC20NEB.py
```diff
@@ -0,0 +1,96 @@
...
+from ml_peg.models.get_models import load_models
+
+MODELS = load_models(current_models)
+
+OUT_PATH = Path(__file__).parent / "outputs"
+S3_KEY = "inputs/nebs/OC20NEB/OC20NEB.zip"
+S3_FILENAME = "OC20NEB.zip"
+
+
+@pytest.mark.slow
```
ElliottKasoar:
```suggestion
@pytest.mark.very_slow
```
With the new data, I think this will be pretty slow

ElliottKasoar (commented):
Hi @jungsdao, sorry for another very slow reply, thanks for addressing so much! I've added a few more comments, but mostly pretty minor.

I think the `interactive_helpers.py` look reasonable, but it might be useful for @joehart2001 to take a look since there may be some commonality with his implementation for the phonons and others since.

If you would also be able to resolve the new conflict with the docs (sorry), and confirm you're happy with the thresholds (https://github.com/ddmms/ml-peg/pull/350#discussion_r3530531035)?

jungsdao:
> Hi @jungsdao, sorry for another very slow reply, thanks for addressing so much! I've added a few more comments, but mostly pretty minor.
> 
> I think the `interactive_helpers.py` look reasonable, but it might be useful for @joehart2001 to take a look since there may be some commonality with his implementation for the phonons and others since.
> 
> If you would also be able to resolve the new conflict with the docs (sorry), and confirm you're happy with the thresholds ([#350 (comment)](https://github.com/ddmms/ml-peg/pull/350#discussion_r3530531035))?

Regarding the good/bad threshold of this benchmark: I think perhaps it's better to evaluate MAE for only those cases that are converged with NEB. One example as image here is not converged and this makes MAE very large. What do you think about this? Do you agree that only converged NEBs should be used for MAE, then I'll modify accordingly. 

image

Otherwise I'll need to increase good/bad threshold for Barrier MAE to for example good 0.5 eV / bad 1.0 eV

jungsdao:
Also when can I see scatter plot for 900 NEBs? Will calculation start once the PR is merged or is calculation done already and can I download the data?

joehart2001:
> > Hi @jungsdao, sorry for another very slow reply, thanks for addressing so much! I've added a few more comments, but mostly pretty minor.
> > I think the `interactive_helpers.py` look reasonable, but it might be useful for @joehart2001 to take a look since there may be some commonality with his implementation for the phonons and others since.
> > If you would also be able to resolve the new conflict with the docs (sorry), and confirm you're happy with the thresholds ([#350 (comment)](https://github.com/ddmms/ml-peg/pull/350#discussion_r3530531035))?
> 
> Regarding the good/bad threshold of this benchmark: I think perhaps it's better to evaluate MAE for only those cases that are converged with NEB. One example as image here is not converged and this makes MAE very large. What do you think about this? Do you agree that only converged NEBs should be used for MAE, then I'll modify accordingly.
> 
> image
> Otherwise I'll need to increase good/bad threshold for Barrier MAE to for example good 0.5 eV / bad 1.0 eV

Hi @jungsdao, this is tricky, but i think the best approach would be to have two metrics:
- all MAE
- only converged MAE (default 0 weight)

jungsdao:
Hi Is there anything else that I need to do for this PR? The pipeline failed for some reason but seems like it's not the problem of code itself. Please let me know if there's anything needed! I'll be happy to see scatter plot later and finish analysis on the result.

ElliottKasoar:
Hi @jungsdao, I just rebase to resolve some conflicts that'd we'd introduced with changes elsewhere (and to re-trigger the CI).

Apologies, yes we'll start running the calculations now. In principle we do sometimes start running these before merging if we're confident the calculation side of things is finished, which has been true here for a while, but we've had quite a few benchmarks to run, so wouldn't have been in a position to do so before now anyway.

I still think I'd like to revisit the visualisation, but I think let's merge this and come back to it once we have a lot of the data

@joehart2001 can you approve, if you're happy?

Thanks for all the hard work, @jungsdao!

ElliottKasoar (approved):

jungsdao:
Thanks for the review, I'll be curious to see the scatter plot later!

joehart2001 (approved):
thanks for all the effort on this @jungsdao!

PR #353, opened 2026-02-07
ethanol-water density curves

### LINKED ISSUE #352: ethanol-water density curves
### Summary

Benchmark of the density of water-ethanol mixtures using NPT MD, compares to experiment.
Metrics:
rms of the density difference.
rms of the difference after subtracting linear interpolation between pure water and ethanol.
Location of the minimal density after subtracting linear interpolation between pure water and ethanol.

### Interactive features

parity plot of density difference with reference

### Category

Physicality?
I am considering adding a liquids category but it probably fits an existing one.

### Data availability

Experimental data added to repository.
From: M. Southard and D. Green, Perry’s Chemical Engineers’ Handbook, 9th Edition. McGraw-Hill Education, 2018.

### Computational cost

High - about 10-100 ps of NPT MD on about 120 water/ethanol molecules, 7 times per model.

### Additional context
Similar to section A.17 in the mace-mp0 paper.

May do some more metrics if time permits and I can find reference data.
Possibly not related to the density if that is ok.

arnon-1:

## Summary
Benchmark of the density of water-ethanol mixtures using NPT MD, compare to experiment.

## Linked issue

Resolves #352

## Progress

- [x] Calculations (may need cleanup)
- [x] Analysis
- [x] Application
- [x] Documentation

## Testing

Tested on mace-mp-0b3 using lower cost simulation settings.
(reduced dispersion cutoff, shorter time, etc.)

## New decorators/callbacks

None (yet)


ml_peg/calcs/liquids/ethanol_water_density/calc_ethanol_water_density.py
```diff
@@ -0,0 +1,308 @@
...
+from ase.md import Langevin
+from ase.md.nose_hoover_chain import IsotropicMTKNPT
+from ase.md.velocitydistribution import (
+    MaxwellBoltzmannDistribution,
+    Stationary,
+    ZeroRotation,
+)
+import pytest
+
+from ml_peg.models.get_models import load_models
```
joehart2001:
```suggestion
from ml_peg.calcs.utils.utils import download_s3_data
from ml_peg.models.get_models import load_models
```

ml_peg/calcs/liquids/ethanol_water_density/calc_ethanol_water_density.py
```diff
@@ -0,0 +1,308 @@
...
+from ml_peg.models.models import current_models
+
+BENCH_ROOT = Path(__file__).resolve().parent
+OUT_PATH = BENCH_ROOT / "outputs"
```
joehart2001:
```suggestion
OUT_PATH = Path(__file__).parent / "outputs"
```

ml_peg/calcs/liquids/ethanol_water_density/calc_ethanol_water_density.py
```diff
@@ -0,0 +1,308 @@
...
+
+    model_out = OUT_PATH / model_name
+    model_out.mkdir(parents=True, exist_ok=True)
+
+    calc = model.get_calculator()
+    calc = model.add_d3_calculator(calc)
+
+    # TODO: the data downloading thing here
+
+    struct_path = DATA_PATH / case.filename
```
joehart2001:
```suggestion
    data_dir = (
        download_s3_data(
            key="inputs/liquids/ethanol_water_density/ethanol_water_density.zip",
            filename="ethanol_water_density.zip",
        )
        / "ethanol_water_density"
    )

    struct_path = data_dir / case.filename
```

ml_peg/calcs/liquids/ethanol_water_density/calc_ethanol_water_density.py
```diff
@@ -0,0 +1,227 @@
...
+        damping=model.d3_kwargs.get("damping", "bj"),
+        xc=model.d3_kwargs.get("xc", "pbe"),
+        dtype=getattr(torch, model.d3_kwargs.get("dtype", "float32")),
+        cutoff=model.d3_kwargs.get(
+            "cutoff", 25.0 * units.Bohr
+        ),  # shortened to make run more manageable.
+    )
+    calcs.append(d3_calc)
+
+    return SumCalculator(calcs)
```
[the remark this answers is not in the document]
ElliottKasoar: It looks like you removed this now, @arnon-1?

If you did want this, I'm not sure I agree entirely with @joehart2001 on changing the `models.yml`, since that changes it in all cases, whereas I think the idea of this is we may want a different dispersion for this specific test?

What I think we'd actually want is to set `model.dispersion_kwargs`, in the same way that some tests set specific `model.default_dtype`, before calling `calc = model.add_d3_calculator(calc)`.

This does all of the `TorchDFTD3Calculator` building for you already, and would know how to parse the test-specific `dispersion_kwargs`.

If the standard parameters are in fact ok, as I think you have now, that's also fine!

ElliottKasoar (commented):
Thanks for this, @arnon-1, it's looking really nice!

We may end up renaming it closer to this, but we have some liquid and water density tests already (#388 and #423) which are in the `molecular_dynamics`  category. Can I suggest we also place this in that category for now for consistency, and once all the tests are in we can reevaluate the most appropriate names?

A couple of other things that may be of interest/to consider:

- The examples above have custom command line inputs. You may find something like that useful here, to make it easier to run the different "cases", assuming they're independent? This is useful for submitting the full set of tests in parallel on HPC
- It's not a requirement but I'd encourage you to look at `janus_core`. This already does a lot of the steps like saving the density, energy, temperature during MD, removing the momentum/rotation, etc. There are some examples [here](https://stfc.github.io/janus-core/tutorials/python/md.html) and [here](https://stfc.github.io/janus-core/tutorials/cli/md.html) (the latter is for the command-line, but the options are the same so it's just to give a broader idea of how it works)

ml_peg/analysis/liquids/ethanol_water_density/metrics.yml
```diff
@@ -0,0 +1,22 @@
...
+    unit: cm^3 mol^-1
+    tooltip: "RMSE of excess (non-ideal) volume"
+    level_of_theory: experiment
+
+  Peak x error:
+    good: 0.026  # approximate error of the reference on the coarse, sampled grid
+    bad: 0.3
+    unit: mole fraction
+    tooltip: "Absolute error in the mole-fraction location of minimal excess volume"
+    level_of_theory: experiment
```
ElliottKasoar:
```suggestion
    level_of_theory: Experimental

  RMSE excess volume:
    good: 0.02  # approximate error of the reference on the coarse, sampled grid
    bad: 1.5
    unit: cm^3 mol^-1
    tooltip: "RMSE of excess (non-ideal) volume"
    level_of_theory: Experimental

  Peak x error:
    good: 0.026  # approximate error of the reference on the coarse, sampled grid
    bad: 0.3
    unit: mole fraction
    tooltip: "Absolute error in the mole-fraction location of minimal excess volume"
    level_of_theory: Experimental
```
This isn't well documented yet, sorry, but this is what we check against

ml_peg/calcs/liquids/ethanol_water_density/calc_ethanol_water_density.py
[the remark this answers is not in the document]
arnon-1: Ok so I am indeed using the standard parameters now. I think these are definitely safe and correct. But I do believe that it should be possible to get more or less correct results while reducing the duration of the calculations (because a water + ethanol system is probably easier to converge than many other of the systems that use the standard parameters). 
So I think it depends a bit on the priorities for ml-peg, and on the computational budget.

arnon-1:
Hi @ElliottKasoar,
Thank you for reviewing my code :-).
I have refactored it to the `molecular_dynamics` category.

I have added the command line arguments for the cases back. (I had originally removed them for consistency/simplicity)
Cases can now be selected with `-k <case_idx>-` (e.g. `-k 0-`).

As for `janus_core`, I have currently already changed my logging to make it similar to #388. So I would prefer not to change it again if you are ok with that.

If there is anything else you would like me to change, please let me know.

ml_peg/calcs/liquids/ethanol_water_density/calc_ethanol_water_density.py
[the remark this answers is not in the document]
ElliottKasoar: It's up to you really. We don't want to be wasteful and run things excessively, especially since some of the models are pretty slow, but I also wouldn't cut things short if you think it'll significantly impact results. 

You're best placed to make that call.

docs/source/user_guide/benchmarks/molecular_dynamics.rst
```diff
@@ -76,3 +76,35 @@ Reference data:
 
 * Same as input data
 * Experimental
+
+
+Water ethanol density curves
+=======
```
ElliottKasoar:
```suggestion
Water ethanol density curves
============================
```
Else `sphinx` will complain

docs/source/user_guide/benchmarks/molecular_dynamics.rst
```diff
@@ -76,3 +76,35 @@ Reference data:
...
+Benchmark of the density of water-ethanol mixtures for different concentrations of ethanol, compare to experiment.
+1 ns of NPT MD on about 120 water/ethanol molecules for 6 concentrations.
+
+
+Metrics
+-------
+
+1. rms of the density difference.
+2. rms of the excess volume difference.
+3. Concentration of the minimal excess volume.
```
ElliottKasoar, line 96: Could you please add a little more detail for each of these to explain how they're practically computed?

Ideally, from these docs, someone would know how to recreate the analysis, so we want minimal ambiguity.

Since this is MD analysis, this may also include things like the choice of equilibration time etc.

docs/source/user_guide/benchmarks/molecular_dynamics.rst
```diff
@@ -76,3 +76,35 @@ Reference data:
...
+
+
+Data availability
+-----------------
+Input structures:
+Packmol generated
+
+Reference densities:
+From: M. Southard and D. Green, Perry’s Chemical Engineers’ Handbook, 9th Edition. McGraw-Hill Education, 2018.
```
ElliottKasoar: I know this is stated above, but could you add a note here that values are experimental, so people can always look at this section to immediately tell what sort of reference we're comparing?

docs/source/user_guide/benchmarks/molecular_dynamics.rst
```diff
@@ -76,3 +76,35 @@ Reference data:
...
+
+Computational cost
```
ElliottKasoar: This is trivial but could these sections be swapped? Just for consistency's sake

ml_peg/analysis/molecular_dynamics/ethanol_water_density/analyse_ethanol_water_density.py
```diff
@@ -0,0 +1,422 @@
...
+import numpy as np
+import pytest
+
+from ml_peg.analysis.utils.decorators import build_table, plot_parity
+from ml_peg.analysis.utils.utils import load_metrics_config, rmse
+from ml_peg.app import APP_ROOT
+from ml_peg.calcs import CALCS_ROOT
+from ml_peg.calcs.utils.utils import download_s3_data
+from ml_peg.models.get_models import get_model_names
+from ml_peg.models.models import current_models
```
ElliottKasoar:
```suggestion
from ml_peg.models import current_models
from ml_peg.models.get_models import get_model_names
```
This has moved, sorry (you may need to rebase for it to work)

ml_peg/app/molecular_dynamics/ethanol_water_density/app_ethanol_water_density.py
```diff
@@ -0,0 +1,78 @@
...
+from ml_peg.app.utils.load import read_plot
+
+# -----------------------------------------------------------------------------
+# Configuration
+# -----------------------------------------------------------------------------
+
+CATEGORY = "molecular_dynamics"
+BENCHMARK_NAME = "ethanol_water_density"
+
+DOCS_URL = f"https://ddmms.github.io/ml-peg/user_guide/benchmarks/{CATEGORY}.html#ethanol-water-density-curves"
```
ElliottKasoar:
```suggestion
DOCS_URL = f"https://ddmms.github.io/ml-peg/user_guide/benchmarks/{CATEGORY}.html#water-ethanol-density-curves"
```
This needs to match the subtitle you used in the rst file

ml_peg/calcs/molecular_dynamics/ethanol_water_density/calc_ethanol_water_density.py
```diff
@@ -0,0 +1,301 @@
...
+from ase.md.velocitydistribution import (
+    MaxwellBoltzmannDistribution,
+    Stationary,
+    ZeroRotation,
+)
+import pytest
+
+from ml_peg.calcs.utils.utils import download_s3_data
+from ml_peg.models.get_models import load_models
+from ml_peg.models.models import current_models
```
ElliottKasoar:
```suggestion
from ml_peg.models import current_models
from ml_peg.models.get_models import load_models
```
This has moved, sorry!

ml_peg/calcs/molecular_dynamics/ethanol_water_density/calc_ethanol_water_density.py
```diff
@@ -0,0 +1,301 @@
...
+    ----------
+    atoms : ase.Atoms
+        Atomic configuration with periodic cell volume.
+
+    Returns
+    -------
+    float
+        Density in g/cm^3.
+    """
+    amu_to_kg = 1.66053906660e-27
```
ElliottKasoar:
```suggestion
    amu_to_kg = 1 / units.kg
```
Generally we want to avoid arbitrary numbers. I think this should be the same

ml_peg/calcs/liquids/ethanol_water_density/calc_ethanol_water_density.py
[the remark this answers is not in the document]
arnon-1: The reason I was asking is because these calculations are dispersion bottlenecked (because the neighbour list in torch-dftd is calculated on cpu).
So they will take a very long time.

I have run the simulations up to 200 ps and with a shorter dispersion length (for mp0), it didn't seem like the densities were moving much anymore but it is hard to tell because of the correlation. The densities I obtained were all approximately within 0.03 g/cm^-3 from the results in the original mp0 paper.
I don't know how much the remaining convergence error compares to the difference in models so the safest is to leave it like this.

So I think I will leave it like this.

docs/source/user_guide/benchmarks/molecular_dynamics.rst
arnon-1: Ok I added info consistent with the other similar benchmarks in the same category (and a bit more for the extra metrics).

arnon-1:
I forgot that I hadn't answered all comments here after committing last time.

I think I have adressed all comments, but I have left two comments open in case I misunderstood the requests. Please feel free to resolve them if they look OK, or let me know if anything still needs changing.

ElliottKasoar:
Hi @arnon-1, thanks for addressing so much, and apologies for the delay on our end, I've been away all of this month, and we've recently introduced a few changes that I wanted to be live before following up too much, to make sure everything is in-sync.

I'll do another review of everything else this week.

arnon-1:
No problem, I think this works now (I had Claude do most of the work but I double checked and tested). So if I understood everything correctly, it should be fine.

As it seemed optional, I am currently not writing structures in get_struct_info as it doesn't work with the way I am currently saving my data (all .traj files have a name depending on model only but are in different directories). I can change this if that would be needed.

ml_peg/app/molecular_dynamics/ethanol_water_density/app_ethanol_water_density.py
```diff
@@ -0,0 +1,80 @@
...
+        """Register callbacks to app."""
+        parity = read_plot(
+            DATA_PATH / "density_parity.json", id=f"{BENCHMARK_NAME}-figure"
+        )
+
+        plot_from_table_column(
+            table_id=self.table_id,
+            plot_id=f"{BENCHMARK_NAME}-figure-placeholder",
+            column_to_plot={
+                "density": parity,
```
ElliottKasoar:
```suggestion
                "RMSE density": parity,
```
This needs to match at least one of the table columns. In principle you could show extend this and just show the same scatter for each of the columns, but I'd imagine the other two would have different plots ideally, so maybe not showing any plot for those is fine for now?

ml_peg/app/molecular_dynamics/ethanol_water_density/app_ethanol_water_density.py
```diff
@@ -0,0 +1,80 @@
...
+# -----------------------------------------------------------------------------
+# Configuration
+# -----------------------------------------------------------------------------
+
+CATEGORY = "molecular_dynamics"
+BENCHMARK_NAME = "ethanol_water_density"
+
+DOCS_URL = f"https://ddmms.github.io/ml-peg/user_guide/benchmarks/{CATEGORY}.html#water-ethanol-density-curves"
+
+DATA_PATH = APP_ROOT / "data" / CATEGORY / BENCHMARK_NAME
```
ElliottKasoar:
```suggestion
BENCHMARK_NAME = "Ethanol-water densities"

DOCS_URL = f"https://ddmms.github.io/ml-peg/user_guide/benchmarks/{CATEGORY}.html#water-ethanol-density-curves"

DATA_PATH = APP_ROOT / "data" / CATEGORY / "ethanol_water_density"
```
The `name` that is passed to `EthanolWaterDecompositionCurvesApp` should be a human-friendly name, as this is what's used for the benchmark subheading and category table column label.

It's also generally fine to use this for Dash IDs, and since this is often done elsewhere, my suggestion is to change `BENCHMARK_NAME` here, in which case the only fix is to revert the `DATA_PATH`.

The name I've chosen is just a suggestion, please do feel free to choose another!

ElliottKasoar (commented):
A final couple of suggestions, but otherwise I think this is probably ready to merge, thanks!

ElliottKasoar (approved):
This looks great, thanks @arnon-1 for getting all the plots working so quickly!

I'll start running this for all the models soon. It'll be a little while until it's live given the complexity of the tests, but I'll let you know when it does go live.

PR #368, opened 2026-02-16
Element filtering

### LINKED ISSUE #255: Filter by elements
### Summary

Ideas:
- global setting e.g. periodic table with tick boxes for elements
- deselcting an element would filter all tests to only include selected elements
- would recalcualte MAEs
- could be done by included element information in scatter hover info for all structures in all tests

### Example

_No response_

### Alternatives

_No response_

### Additional context

_No response_

joehart2001:

## Summary

Element filtering
- Element sets are defined in `element_sets.yml` for example all, hcno).
- filtering occurs at analysis time: 
   - each structures element symbols are collected
   - a boolean mask is built per element set and applied 
   - compute metrics + plots on filtered data

- also recompute a set of category and summary table scores for each element set

To add:
- periodic table showing which elements are shown

## Linked issue

Resolves #255 

## Testing
LNCI16 for now

### DIFF
--- a/ml_peg/analysis/element_sets.yml
+++ b/ml_peg/analysis/element_sets.yml
@@ -1,0 +1,9 @@
+element_sets:
+  all:
+    name: All elements
+    description: Include every structure regardless of composition.
+    elements: null
+  hcno:
+    name: HCNO
+    description: Structures composed exclusively of hydrogen, carbon, nitrogen, or oxygen.
+    elements: [H, C, N, O]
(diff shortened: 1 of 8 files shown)


ElliottKasoar:
From testing, one bug to fix is that weights/thresholds for benchmarks are reset when switching tabs

alinelena:
is superseeded by #512 ?

ElliottKasoar:
Yep, superseded by #512, which may itself be superseded by a simpler form initially. We'll keep the branch since it has some useful bits to borrow.

PR #388, opened 2026-02-21
Liquid densities

kuryla:

## Summary

Benchmark of 61 organic liquid densities. Densities are computed from a NPT molecular dynamics run and errors are obtained with respect to the experimental densities.

## Linked issue

Resolves #

## Progress

- [x] Calculations
- [x] Analysis
- [x] Application
- [x] Documentation

## Testing

MACE-OMOL, MACE-POLAR-1-M, MACE-POLAR-1-L, UMA-S-1P1-OMOL


ml_peg/analysis/molecular_dynamics/liquid_densities/analyse_liquid_densities.py
```diff
@@ -0,0 +1,180 @@
...
+DEFAULT_THRESHOLDS, DEFAULT_TOOLTIPS, DEFAULT_WEIGHTS = load_metrics_config(
+    METRICS_CONFIG_PATH
+)
+
+DATA_PATH = (
+    download_s3_data(
+        filename="liquid_densities.zip",
+        key="inputs/molecular_dynamics/liquid_densities/liquid_densities.zip",
+    )
+    / "liquid_densities"
```
ElliottKasoar: Is this supposed to be here/has this been uploaded?

I think we probably want this in `outputs/...` or something, since this is more set up as calculation inputs, not analysis inputs, but I also don't know if we want to do the analysis download as part of the script. I think I'd rather check for the data in calcs/.../outputs, and if it's not there, suggest downloading is using `ml_peg download`

ElliottKasoar (approved):

PR #391, opened 2026-02-24
Add Ti64 and diamond phonon benchmarks

### LINKED ISSUE #297: Add Ti64 and diamond phonon benchmarks
### Summary

Compute phonon dispersion for carbon diamond and Ti64, and phonon DOS/PDOS and Helmholtz free energies for Ti64, and compare to DFT references.

### Interactive features

Interactive metric tables, predicted vs reference frequency scatter plots, and per-model dispersion preview panels.

### Category

bulk_crystal

### Data availability

Ti64 (PBE): input structures https://github.com/7radians/ml-peg-data/tree/main/ti64_data

Full details on the data and benchmark:

* Allen, C. S. & Bartók, A. P. Multi-phase dataset for Ti and Ti-6Al-4V. Preprint at https://arxiv.org/abs/2501.06116 (2025).

* Radova, M., Stark, W. G., Allen, C. S. et al. Fine-tuning foundation models of materials interatomic potentials with frozen transfer learning. npj Comput Mater 11, 237 (2025). https://doi.org/10.1038/s41524-025-01727-x

Carbon diamond (RSCAN): https://github.com/7radians/ml-peg-data/tree/main/diamond_data

### Computational cost

Medium (typically take minutes on CPU)

### Additional context

Ti64 phonon calculations were run as a single process on CPU on an x86_64 machine (11th Gen Intel(R) Core(TM) i5-1145G7; 4 cores / 8 threads). No explicit parallel execution (MPI or multiprocessing) was used in the benchmark driver.

7radians:

## Summary

This PR adds two new bulk-crystal phonon benchmarks with Dash app integration:

### Ti-6Al-4V (Ti64) phonons (`bulk_crystal/ti64_phonons`)
- **Calc**: Runs a CASTEP-referenced phonon suite across 10 Ti64 configurations (hcp/hex + bcc variants). For each model/case it relaxes the structure (LBFGS) then computes:
  - dispersion on a high-symmetry k-path,
  - DOS/PDOS on a dense mesh,
  - optional thermo/free-energy curves for a defined subset of cases.
  Raw outputs are written to `outputs/<model>/` as `<case>.npz` + `<case>.json`.
- **Analysis**: Produces per-model `metrics.json` plus:
  - `ti64_phonons_metrics_table.json` (Dash table)
  - `ti64_phonons_interactive.json` (interactive scatter)
  Metrics include dispersion RMSE (mean/max), ω_avg MAE, and ΔF per-atom at 0 K / 2000 K for TP-enabled cases.
- **App**: Adds a Dash app that reuses shared callback builders to provide:
  - table → ω_avg scatter (ref vs pred),
  - scatter selection → rendered dispersion + DOS preview from calc artifacts.

### Diamond phonons (bands-only) (`bulk_crystal/diamond_phonons`)
- **Calc**: Computes phonopy force constants and band dispersions for diamond for each model and writes `FORCE_CONSTANTS` + `band.yaml` under `outputs/<model>/`. The phonon band path is taken directly from the DFT reference NPZ (`dft_band.npz`) so predictions match the exact reference q-path.
- **Analysis**: Loads the DFT reference bands (cm⁻¹ → THz), loads each model’s `band.yaml`, sorts bands per q-point and reports Band MAE + Band RMSE (THz). Writes:
  - `diamond_phonons_bands_table.json`
  - `diamond_phonons_bands_interactive.json`
- **App**: Adds a Dash app that shows metric scatter plots and a per-model dispersion preview panel rendered from `band.yaml` with optional DFT overlay (RSCAN).

## Linked issue

Resolves #297 

## Progress

- [x] Calculations
- [x] Analysis
- [x] Application
- [x] Documentation

## Testing

Tested both benchmarks on:
- mace-mp-0a
- mace-mp-0b3
- mace-mpa-0
- mace-omat-0
- mace-matpes-r2scan
- orb-v3-consv-inf-omat
- pet-mad

## New decorators/callbacks

- [x] No new callbacks/decorators required (reuses existing table/scatter decorators and shared callback builders).


joehart2001:
Hey @7radians, have you had a look at adding the thermal conductivity and Grüneisen parameter? would be good to get those in! For the calc, analysis etc, i think we will have a look at writing some more general stuff that you can use, as many people want to look at phonons in some shape or form. will update you on the progress of that

joehart2001:
Hey @7radians ive now made a PR for the phonon calc in a simpler way than what i showed you originally. it would be great to match this, but for your application we won't need the multiple workers. let me know if you need any help!

7radians:
Hey @joehart2001, all done, phonons refactored and Gruneisen + thermal conductivity added.

ml_peg/models/models.py
ElliottKasoar, line 1: I think this is reverting us to an older form of the file?

ml_peg/calcs/bulk_crystal/diamond_phonons/calc_diamond_phonons.py
```diff
@@ -0,0 +1,136 @@
...
+from ml_peg.models.get_models import get_model_names, load_models
+
+GITHUB_BASE = "https://raw.githubusercontent.com/7radians/ml-peg-data/main"
+
+EXTRACTED_ROOT = Path(
+    download_github_data(
+        filename="diamond_data/data.zip",
+        github_uri=GITHUB_BASE,
+    )
+)
```
ElliottKasoar: Are you happy for these to stay here, or would you prefer them to be in our S3 bucket?

docs/source/user_guide/benchmarks/bulk_crystal.rst
```diff
@@ -514,3 +514,278 @@ Reference data:
...
+
+
+
+Data availability
+-----------------
+
+Input structures (https://github.com/7radians/ml-peg-data/tree/main/diamond_data):
+
+* ``diamond.yaml``: conventional cubic cell (8 atoms, a = 3.56 Å) × 4×4×4 = 512-atom
+  supercell, used for all MLIP band and thermal calculations.
```
ElliottKasoar: Are these structures from the reference? It's worth clarifying, and giving more explanation if not.

ml_peg/analysis/bulk_crystal/phonons_ti64/metrics.yml
ElliottKasoar, line 1: Very general question, but are you aware of/have you thought much about selection of these good and bad thresholds?

If you're happy with, great, but if not, we're happy to discuss/clarify!

ml_peg/calcs/bulk_crystal/ti64_phonons/calc_ti64_phonons.py
```diff
@@ -0,0 +1,558 @@
...
+        github_uri=GITHUB_BASE,
+    )
+)
+
+DATA_PATH = EXTRACTED_ROOT / "data"
+
+OUT_PATH = CALCS_ROOT / "bulk_crystal" / "ti64_phonons" / "outputs"
+
+FMAX = 0.001
+STEPS = 10000
```
ElliottKasoar: Do we actually want to optimise for 10,000 steps?

@joehart2001 may have more experience with this from the other phonon benchmark (where we use 1000, I believe), but from other benchmarks, my impression is the only models that keep going beyond a few hundred steps are ones that explode, in most cases, so this can slow things down enormously when we run it for less suited models.

ml_peg/calcs/bulk_crystal/ti64_phonons/calc_ti64_phonons.py
```diff
@@ -0,0 +1,558 @@
...
+    ph_pdos.run_total_dos()
+    pdos = ph_pdos.get_projected_dos_dict()
+    dos = ph_pdos.get_total_dos_dict()
+
+    # ML thermo
+    tp_dict: dict[str, Any] = {}
+    if do_tp:
+        ph_pdos.run_mesh(MESH_202020)
+        ph_pdos.run_thermal_properties(t_step=1, t_max=2000, t_min=0)
+        tp_dict = ph_pdos.get_thermal_properties_dict()
```
ElliottKasoar: Is this doing anything that isn't done by [janus_core](https://stfc.github.io/janus-core/tutorials/python/phonons.html) (code [here](https://github.com/stfc/janus-core/blob/main/janus_core/calculations/phonons.py)).

You don't have to use it, of course, but it's quite a lot of boiler-plate code that ideally we wouldn't need, especially if we want to do this for any other systems too.

ml_peg/calcs/bulk_crystal/diamond_phonons/calc_diamond_phonons.py
```diff
@@ -0,0 +1,136 @@
...
+    thermal_path = out_dir / "diamond_thermal.json"
+
+    need_band = not band_path.exists()
+    need_thermal = not thermal_path.exists()
+
+    if not need_band and not need_thermal:
+        return
+
+    model = load_models(model_name)[model_name]
+    calc = model.get_calculator()
```
ElliottKasoar:
```suggestion
    calc = model.get_calculator(precision="high")
```
Or "low" (~float32). The default is "high" (~float64), so this suggestion doesn't change the current behaviour, but it may have been equivalent to "low" when you first wrote this, and either way, I think it's nice to be explicit.

ml_peg/calcs/bulk_crystal/ti64_phonons/calc_ti64_phonons.py
```diff
@@ -0,0 +1,558 @@
...
+    model_name
+        Model identifier used for output paths.
+    """
+    print(f"{case_name} | {model_name}")
+
+    # DFT dispersion along k-path (raw reference)
+    pfc = PhononFromCastep(castep_file=structure_file, kpath_in=kpath)
+
+    # Relax
+    atoms = ase.io.read(structure_file)
```
ElliottKasoar: In the general phonon benchmark, we fix the symmetry before relaxing. Do we not need to do that here, to make sure we can compare like-for-like with the reference?

We also use the FIRE optimiser, I'm not sure if you've discussed this choice wth @joehart2001 at all?

ml_peg/calcs/bulk_crystal/ti64_phonons/calc_ti64_phonons.py
```diff
@@ -0,0 +1,558 @@
...
+            arrays["tp_temperatures"] = np.asarray(tp_dict["temperatures"], dtype=float)
+        if "free_energy" in tp_dict:
+            arrays["tp_free_energy"] = np.asarray(tp_dict["free_energy"], dtype=float)
+
+    # DFT qpoints
+    if qpoints_file is not None:
+        arrays["q_weights"] = np.asarray(q_weights, dtype=float)  # type: ignore[arg-type]
+        arrays["q_frequencies_dft"] = np.asarray(q_freq_dft, dtype=float)  # type: ignore[arg-type]
+
+    write_case_npz(case_name, model_name, **arrays)
```
ElliottKasoar: How does this compare to what we write in the general phonon benchmark?

It might be useful to be consistent (and maybe even share a function, if it makes sense )?

ml_peg/calcs/bulk_crystal/phonons/phonons_utils.py
```diff
@@ -195,7 +195,9 @@ def init_phonopy_from_ref(
     if displacement_dataset is not None:
         phonons.dataset = displacement_dataset
     elif displacement_distance is not None:
-        phonons.generate_displacements(distance=displacement_distance)
+        phonons.generate_displacements(
+            distance=displacement_distance, is_plusminus=True
```
ElliottKasoar: Is this not changing the behaviour of the current phonons benchmark? It should maybe be an extra parameter to the function?

ml_peg/calcs/bulk_crystal/phonons/thermal_utils.py
```diff
@@ -0,0 +1,310 @@
...
+from phonopy.api_phonopy import Phonopy
+
+from ml_peg.calcs.bulk_crystal.phonons.phonons_utils import (
+    _ase_atoms_to_phonopy,
+    get_fc2_and_freqs,
+)
+
+# Physical constants (SI)
+_kb = 1.380649e-23
+_hbar = 1.054571817e-34
```
ElliottKasoar: I think you should be able to import these from `ase.units`

ml_peg/app/bulk_crystal/diamond_phonons/app_diamond_phonons.py
```diff
@@ -0,0 +1,205 @@
...
+    model_asset_from_scatter,
+    scatter_and_assets_from_table,
+)
+from ml_peg.app.utils.plot_helpers import build_serialized_scatter_content
+from ml_peg.calcs import CALCS_ROOT
+from ml_peg.calcs.utils.utils import download_github_data
+
+GITHUB_BASE = "https://raw.githubusercontent.com/7radians/ml-peg-data/main"
+
+EXTRACTED_ROOT = Path(
```
ElliottKasoar: We don't normally want to be downloading things when we run the app (even with caching), can this be set up so it's accessible from the calc/analysis parts?

ml_peg/calcs/bulk_crystal/ti64_phonons/calc_ti64_phonons.py
```diff
@@ -0,0 +1,558 @@
...
+from ml_peg.models.get_models import get_model_names, load_models
+
+GITHUB_BASE = "https://raw.githubusercontent.com/7radians/ml-peg-data/main"
+
+EXTRACTED_ROOT = Path(
+    download_github_data(
+        filename="ti64_data/data.zip",
+        github_uri=GITHUB_BASE,
+    )
+)
```
ElliottKasoar: Are you happy for these to stay here, or would you prefer them to be in our S3 bucket?

ml_peg/calcs/bulk_crystal/ti64_phonons/calc_ti64_phonons.py
```diff
@@ -0,0 +1,558 @@
...
+
+    Parameters
+    ----------
+    model_name
+        Name of the ML-PEG model to evaluate.
+    """
+    patch_ase_castep_reader_dummy_energy(dummy_energy=0.0)
+
+    model = load_models(model_name)[model_name]
+    calc = model.get_calculator()
```
ElliottKasoar:
```suggestion
    calc = model.get_calculator(precision="high")
```
Or "low", as above

ml_peg/calcs/bulk_crystal/diamond_phonons/calc_diamond_phonons.py
```diff
@@ -0,0 +1,136 @@
...
+    phonons_ref = load_phonopy(str(DIAMOND_YAML))
```
ElliottKasoar: Do we want want to relax this for each model?

ml_peg/analysis/bulk_crystal/phonons_diamond/metrics.yml
ElliottKasoar, line 1: Same as the other metrics: are you aware of/have you thought much about selection of these good and bad thresholds?

ml_peg/calcs/bulk_crystal/ti64_phonons/calc_ti64_phonons.py
```diff
@@ -0,0 +1,558 @@
...
+
+    for spec in CASES:
```
ElliottKasoar: It might be nice to show process over this loop?

ml_peg/calcs/bulk_crystal/phonons_ti64/calc_phonons_ti64.py
ElliottKasoar, line 1: Can you set a (default) spin (multiplicity) and charge for the structures, as this is required by some models.

ml_peg/app/bulk_crystal/ti64_phonons/ti64_interactive_helpers.py
ElliottKasoar, line 1: At the point when we run the app, we should generally just be loading figures etc. I think a lot of this should probably be moved to analysis?

Have you also checked if similar things are already done for the general phonon benchmark?

ml_peg/app/bulk_crystal/diamond_phonons/diamond_interactive_helpers.py
ElliottKasoar, line 1: As with the other app, I think a lot of this ought to be done in analysis, and I'd also check for overlap with existing functions.

ElliottKasoar (changes requested):
Hi @7radians, thanks for this! Apologies for such a delayed review. I'm still taking a look at some parts, but I thought there's probably enough to start the discussion.

In addition to the above comments, please can you take a look at our new filtering guidelines: https://ddmms.github.io/ml-peg/developer_guide/filter.html.

The principles are relatively simple, but you do have to be a little careful, so again, if anything is unclear, please do ask! The biggest thing to check is that the info.json file is created during analysis, and this is read by the app.

(You'll need to rebase/merge the latest main for this)

One other question was whether it might make more sense to combine these into a single test? @joehart2001 may also have thoughts, but I think I recall some uncertainty about whether it's appropriate to have the diamond phonons as a standalone test compared to the full set of phonons in the current benchmark we have, for example.

docs/source/user_guide/benchmarks/bulk_crystal.rst
```diff
@@ -514,3 +514,278 @@ Reference data:
...
+
+The Slack formula is a semi-empirical approximation that captures the dominant anharmonic
+scattering channel. It does not include phonon–phonon scattering from third-order force
+constants and should be treated as an order-of-magnitude estimate.
+
+
+Computational cost
+------------------
+
+Medium: band-structure and thermal-property calculations together typically take
```
ElliottKasoar:
```suggestion
Low: band-structure and thermal-property calculations together typically take
```
Compared to other benchmarks (and given we're not marking the test as @slow / very_slow

docs/source/user_guide/benchmarks/bulk_crystal.rst
```diff
@@ -514,3 +514,278 @@ Reference data:
...
+   Mean absolute error in Helmholtz free energy at 2000 K, reported as eV/atom, over the
+   subset of cases where thermodynamic outputs are available.
+
+   Computed as in (4), but using the final temperature point (2000 K).
+
+
+Computational cost
+------------------
+
+Medium: dispersion, DOS/PDOS and thermodynamic calculations typically take minutes per model on CPU.
```
ElliottKasoar:
```suggestion
Low: dispersion, DOS/PDOS and thermodynamic calculations typically take minutes per model on CPU.
```
Compared to other benchmarks (and given we're not marking the test as @slow / very_slow

ml_peg/models/models.py
7radians: Sorry, accidental revert from a stale local state. Restored now, the file is identical to main

ml_peg/calcs/bulk_crystal/diamond_phonons/calc_diamond_phonons.py
7radians: I'm happy either way, what would you prefer?

ml_peg/calcs/bulk_crystal/ti64_phonons/calc_ti64_phonons.py
7radians: Reduced to 1000 + switched to FIRE, matching the general phonon benchmark (kept fmax=0.001 as the cells are small, happy to align that too if you prefer). I used the settings from the reference paper initially but this seems to work well

ml_peg/calcs/bulk_crystal/ti64_phonons/calc_ti64_phonons.py
7radians: This block no longer exists, the calc now uses the same shared helpers as the general phonon benchmark

ml_peg/calcs/bulk_crystal/diamond_phonons/calc_diamond_phonons.py
7radians: Done, explicit precision="high" in both

ml_peg/calcs/bulk_crystal/ti64_phonons/calc_ti64_phonons.py
7radians: Aligned now, the initial npz/json format is gone. Both benchmarks write the general benchmark's per-system files, with references under outputs/DFT/. That's also what let the apps reuse the general dispersion renderer

ml_peg/calcs/bulk_crystal/phonons/phonons_utils.py
7radians: It's now an explicit is_plusminus parameter defaulting to phonopy's "auto". This was important as otherwise the transformers weren't making sensible predictions

ml_peg/calcs/bulk_crystal/phonons/thermal_utils.py
7radians: done

ml_peg/calcs/bulk_crystal/ti64_phonons/calc_ti64_phonons.py
7radians: Added FIRE + FixSymmetry per model, as in the general benchmark. For ideal diamond the position relax doesn't do anything because of symmetry, but if we want to be consistent and potentially combine the diamond with the general benchmark, I can keep it?

ml_peg/calcs/bulk_crystal/ti64_phonons/calc_ti64_phonons.py
7radians: Again I'm happy either way, what would you prefer?

ml_peg/calcs/bulk_crystal/ti64_phonons/calc_ti64_phonons.py
7radians: done

ml_peg/calcs/bulk_crystal/diamond_phonons/calc_diamond_phonons.py
7radians: As mentioned above, for ideal diamond the position relax doesn't do anything because of symmetry, but if we want to be consistent and potentially combine the diamond with the general benchmark, I can keep it?

ml_peg/calcs/bulk_crystal/ti64_phonons/calc_ti64_phonons.py
7radians: Added tqdm over the cases

ml_peg/calcs/bulk_crystal/phonons_ti64/calc_phonons_ti64.py
7radians: Done

ml_peg/app/bulk_crystal/ti64_phonons/ti64_interactive_helpers.py
7radians: Both modules are deleted. Ti64 reuses the general benchmark's render_dispersion_component/lookup_system_entry directly (the shared formats made that possible). For diamond, the dispersion comparison is pre-rendered as a PNG during analysis and the app just loads it, so no shared-module changes needed for either.

ml_peg/app/bulk_crystal/diamond_phonons/diamond_interactive_helpers.py
7radians: Removed this module as mentioned above

docs/source/user_guide/benchmarks/bulk_crystal.rst
7radians: Done

docs/source/user_guide/benchmarks/bulk_crystal.rst
7radians: Done

docs/source/user_guide/benchmarks/bulk_crystal.rst
7radians: diamond.yaml is gone, the structure comes from the reference .xyz

7radians:
Hi @ElliottKasoar, thank you very much for the review! All points should now be addressed, I've left replies on the individual threads, but the short version:

I've pre-converted the reference data to the general phonon benchmark's formats (band/DOS pickles, thermal JSON, structures), and this simplified things a lot and allowed sharing functions throughout: both calcs now use the general benchmark's init_phonopy_from_ref/get_fc2_and_freqs, write the same per-system output files with references under outputs/DFT/, and the apps reuse the general dispersion renderer, so the custom CASTEP parser, the ASE reader patch, diamond.yaml, and both app helper modules are gone entirely, and the reference steps are now just a copy of the downloaded data. Outside the two benchmark folders, the only changes left are the is_plusminus parameter in phonons_utils (as you suggested, defaulting to the current behaviour) and the new thermal_utils module.

I've also merged the latest main and followed the filtering guidelines (info.json from analysis, info_path in the apps, calcs catch per-case errors and run with the mock calculators), and while going through thermal_utils I found and fixed a factor-of-2 error in the Grüneisen parameter, the corrected diamond reference gives γ = 1.01, in line with literature. All model results have been regenerated with the corrected pipeline.

Happy to move the data zips to the S3 bucket, and to discuss the metric thresholds, some I've recalibrated with physical scales in mind (please see the replies in the thread), and some others I guess are just about how strict we want to be.

ml_peg/app/bulk_crystal/diamond_phonons/app_diamond_phonons.py
7radians: Done

ml_peg/analysis/bulk_crystal/phonons_ti64/metrics.yml
7radians: Yes, I've calibrated these deliberately, partly against physical scales and partly by how strict I want the benchmark to be. Please let me know what you think about these, I'm happy to discuss!
Dispersion RMSE mean (bad = 2.0 THz, weight 2): 20% of the ~10 THz Ti64 spectrum, the scale at which soft-mode and phase-stability conclusions can become unreliable. 
Dispersion RMSE max (bad = 2.0 THz, weight 0.5): intentionally the same absolute bar applied to the worst case, a strictness statement that no single phase may be qualitatively wrong, with a low weight so it acts as a tiebreaker rather than dominating.
ω_avg MAE (bad = 1.0 THz): 10% of ω_max; this isolates systematic softening/stiffening of the lattice (sign-cancelling shape errors are already captured by the RMSE, hence the separate metric).
ΔF at 0 K (bad = 0.007 eV/atom): physically anchored to two scales. It's 20% of the ~35 meV/atom zero-point energy, and it's the scale of free-energy differences that decide phase competition in Ti alloys (tens of meV/atom), so errors beyond this scramble phase ordering.
ΔF at 2000 K (bad = 0.03 eV/atom): the same phase-stability logic at high temperature. |F| itself grows to ~1 eV/atom, but the differences that matter for the α/β competition stay at tens of meV/atom, so the absolute bar stays tight. This is knowingly the strictest metric in the set, current foundation models sit at or beyond it, which I think is fair. High-temperature free energies are where all models are weakest.

ml_peg/analysis/bulk_crystal/phonons_diamond/metrics.yml
7radians: Same here, a mix of physical anchors and deliberate strictness. Again, please let me know if you have any thoughts about this, I'm happy to discuss!

Band MAE (bad = 1.0 THz, weight 2): 1 THz is 2.5% of diamond's 40 THz spectrum. This is a strictness choice. diamond is a simple, rigid, well-sampled crystal, so I want only quantitatively accurate force constants to score here.
Band RMSE (bad = 2.0 THz): the outlier-sensitive counterpart, at 2× the MAE bar since RMSE runs systematically higher for the same data.
Δγ (bad = 0.2): physically anchored - γ enters thermal expansion linearly (α ∝ γ), so 0.2 on γ_ref ≈ 1.01 means thermal expansion wrong by 20%, and it's also roughly the spread among experimental determinations of diamond's γ, i.e. beyond it a model is distinguishably wrong.
Δθ_D (bad = 90 K): θ_D here is derived from ω_max, at ≈48 K per THz,so 90 K ≈ a 1.9 THz error in the stiffest phonon, deliberately consistent with the band-structure bars above rather than an independent scale.
Δκ_L (bad = 400 W/m/K): κ_Slack ∝ θ_D³/γ² amplifies the other two errors, a model right at the Δγ bar alone already produces a ~45% κ error, so 400 (≈53% of κ_ref = 751) is approximately the propagation of the γ and θ_D thresholds through the Slack formula, keeping the three thermal metrics mutually consistent.

ml_peg/analysis/bulk_crystal/phonons_ti64/metrics.yml
joehart2001: it could be good to commnet on this very briefly in the docs. also to define what terms such as "unreliable" mean, as these will need to be explained fully for your section of the paper, so probably easiest to do in one go

ml_peg/calcs/bulk_crystal/diamond_phonons/calc_diamond_phonons.py
joehart2001: for consistency we can do s3?

7radians: Sure @joehart2001 , would you like to pull the data from https://github.com/7radians/ml-peg-data, or I can just send it to you?

ml_peg/analysis/bulk_crystal/phonons_ti64/metrics.yml
7radians: Done

joehart2001:
hey @7radians ive made a few changes
- rebase
- move to s3
- rename to phonons_diamond so we keep all the phonon benchmarks grouped together
- added bar chart plots for some metrics
- just kept mae not rmse as we will have options to change to this in the future

were you also planning on adding some of the additional metrics in diamond to the ti64 benchmark?

overall i think theres not much left to do, @ElliottKasoar could you take a look when you get a chance, should all be working wel and is quick to test

7radians:
@joehart2001 thank you! Yes, I think I’ve added everything I had in mind in terms of metrics

ElliottKasoar (approved):

ElliottKasoar:
Thanks for all the work on this, @7radians! We'll let you know if anything else comes up as part of adding this to the live site.

PR #410, opened 2026-03-05
Add superacids category: HF/SbF5 mixture densities and HF structure factor (Ref #379)

### LINKED ISSUE #379: HF/SbF5 mixture
### Summary

Benchmark against experimental data of liquid superacid mixtures composed by HF + SbF5.
The first test compares the reproduced density as a function of % of SbF5. Ideally, it consists in NPT simulation of pure HF, pure Sbf5, and 1 or 2 mixtures.
The second test compares the reproduced structure factor S(q). It consists in two NVT simulation of pure HF and pure SbF5. 

### Interactive features

plots of density and S(q), showing simulated vs experiment

### Category

liquid/density and liquid/structure_factors
or
superacids/HF-SbF5

### Data availability

Experimental data reported in publications

### Computational cost

High
100 ps for density (3/4 NPT per model)
50 for S(q) (2 simulations per model)

### Additional context

https://en.wikipedia.org/wiki/Fluoroantimonic_acid

mattiaperr:

## Summary

  test 1) HF/SbF5 mixture densities: NPT molecular dynamics at 288.65 K and 1 atm for three
  compositions (pure HF, 10 mol % SbF5, pure SbF5), ~200 atoms each, 100 ps per system.
  The density is taken from the average volume of the second half of the run, and the
  metric is the MAPE against experimental densities (Shair and Schurig, 1951).

  test 2) HF structure factor: a single NPT run of 100 HF molecules at 296 K and 1.2 bar for
  50 ps. S(q) iis computed with the python library "MDANSE". Two metrics: the R-factor against experiment
  (McLain et al., 2004) up to 4 A^-1, and the error on the position of the first peak.

## Linked issue

Resolves #379 

## Progress

  - [x] Calculations
  - [x] Analysis
  - [x] Application
  - [x] Documentation

todo:
- `HF_structure.zip` has to be uploaded to S3

## Testing

mace-mp-0a, mace-mp-0b3 and mace-mpa-0

## New decorators/callbacks

Two changes to `plot_scatter` in `ml_peg/analysis/utils/decorators.py`, in their own commit:
1. `highlight_range` was being redrawn once per model, so the shaded band stacked up and the annotation was printed on top of itself. It is now drawn once per figure.
2. A new `hlines` option to draw dashed horizontal reference lines, used to mark the experimental target density on the density-vs-time plots. It defaults to `None`, so no existing plot is affected.

  ## Other notes

  This benchmark adds `mdanse` as a dependency, used for the structure factor.

  Unit tests were added in `tests/test_superacids.py` covering the restart logic, the density calculation and the R-factor.

  AI disclosure: I used an AI assistant (Claude) while working on this PR, mainly for refactoring and docstrings. All the physics of the simulations and the design of the workflow in the tests are mine, and I have reviewed every line.


ElliottKasoar (commented):
Thanks for this, @mattiaperr, this looks great so far! I've left a few minor comments.

If you can share the input data, it would be great to start testing this!

ml_peg/calcs/superacids/HF_SbF5_density/calc_HF_SbF5_density.py
```diff
@@ -0,0 +1,134 @@
...
+    calc = model.get_calculator()
+
+    # Add D3 calculator for this test
+    calc = model.add_d3_calculator(calc)
+
+    # download dataset
+    hf_sbf5_density_dir = (
+        download_s3_data(
+            key="inputs/superacids/HF_SbF5_density/HF_SbF5_density.zip",
+            filename="HF_SbF5_density.zip",
```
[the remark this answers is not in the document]
joehart2001: @ElliottKasoar @mattiaperr any update on this? would like to do some testing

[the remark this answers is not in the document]
mattiaperr: [HF_SbF5_density.zip](https://github.com/user-attachments/files/27359712/HF_SbF5_density.zip)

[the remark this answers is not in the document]
joehart2001: thanks, uplaoded!

ml_peg/calcs/superacids/HF_SbF5_density/calc_HF_SbF5_density.py
```diff
@@ -0,0 +1,134 @@
...
+# Unit conversion
+EV_TO_KJ_PER_MOL = units.mol / units.kJ
+
+
+# Simulation parameters
+TEMPERATURE_K = 288.6
+PRESSURE_ATM = 1
+DT_FS = 0.5  # DT in femtoseconds
+
+N_MIN_STEPS = 100  # maximum minimization steps
```
joehart2001: It seems like many models do not converge within 100 steps. as this is very cheap compared to the MD, do you think its worth bumping this up to even 1000?

joehart2001:
As this test involves running mutliple MDs in series, it could be a good idea to change parameterise this like in the [liquid densities test](https://github.com/ddmms/ml-peg/blob/main/ml_peg/calcs/molecular_dynamics/liquid_densities/calc_liquid_densities.py)

when you add the docs could you give an estiamte of how long the MD takes on GPU? if its not too expensive then it may not be woth doing this.

Otherwise, the calc, analysis and app seem to work well. For the analysis good and bad thresholds, is a 0% good value realistic due to error associated with the experimental measurements? For the app, we could consider some structure visualisation. as this is a trajectory we could visualise this, but this depends on what you think would be best for your test. Thanks!

mattiaperr:
thanks!

I estimate 6 GPU hours for the pure HF system and 2 GPU hours for the pure SbF5 system, for a model that includes dispersion.
What do you think? 

I wanted to follow up on this to discuss the structure of the test
I have developed, and currently use, a routine to estimate the HF structure factor and compare it with experimental data, in the style of the test: https://github.com/ddmms/ml-peg/pull/445 .
Can I integrate such test into these same scripts, or should I create a separate calc/analysis/application within my same benchmark? 
This would consist in a single NVT simulation at experimental density, which is much faster (25 ps to obtain the structure factor versus 100 ps for the density).

Given the concern about long simulation times, the whole benchmark could be reformulated as follows:

- if the HF density is too computationally expensive, use the structure factor as a metric (whose first peak position correlates with the density, and this fact can be taken into account as done in https://github.com/ddmms/ml-peg/pull/445); and then perform only a single NPT simulation of the HF/Sbf5 density at a single point on the diagram.

What do you think?

ml_peg/calcs/superacids/HF_SbF5_density/calc_HF_SbF5_density.py
```diff
@@ -0,0 +1,140 @@
...
+
+MODELS = load_models(current_models)
+
+OUT_PATH = Path(__file__).parent / "outputs"
+
+# Simulation parameters
+TEMPERATURE_K = 288.6
+PRESSURE_ATM = 1
+DT_FS = 0.5  # DT in femtoseconds
+N_MIN_STEPS = 1000  # maximum minimization steps
```
ElliottKasoar: Do you find models often take more than a few hundred steps? In my experience, unless the convergence criteria are very strict, the only models that tend to are those that are exploding and have no real hope of convergence, in which case it's just slowing things down.

Then again, this is probably still much cheaper than the full MD, so it doesn't matter too much.

ml_peg/calcs/superacids/HF_SbF5_density/calc_HF_SbF5_density.py
```diff
@@ -0,0 +1,140 @@
...
+
+    Parameters
+    ----------
+    mlip
+        Name of model and model to get calculator.
+    system
+        System identifier (X_0, X_10, X_100).
+    """
+    model_name, model = mlip
+    calc = model.get_calculator()
```
ElliottKasoar:
```suggestion
    calc = model.get_calculator(precision="low")
```
This used to be the default (via parameters we'd set models.yml), but now we've introduced this option, the default is "high" (float64), so I suspect we want "low"? Even if not, I'd still suggest explicitly setting it.

ml_peg/calcs/superacids/HF_SbF5_density/calc_HF_SbF5_density.py
ElliottKasoar, line 1: It would be useful to provide a way to restart the MD. Again, you can refer to the densities tests, or you can also look into [janus-core's MD](https://stfc.github.io/janus-core/tutorials/python/md.html), which has restart options too, and would also deal with a lot of things e.g. relaxing the geometry, saving volumes/densities, setting up the velocity distribution/removing rotations etc.

ml_peg/analysis/superacids/HF_SbF5_density/analyse_HF_SbF5_density.py
```diff
@@ -0,0 +1,203 @@
...
+from ase.io import read
+import numpy as np
+import pytest
+
+from ml_peg.analysis.utils.decorators import build_table, plot_parity
+from ml_peg.analysis.utils.utils import build_dispersion_name_map, load_metrics_config
+from ml_peg.app import APP_ROOT
+from ml_peg.calcs import CALCS_ROOT
+from ml_peg.models.get_models import get_model_names
+from ml_peg.models.models import current_models
```
ElliottKasoar:
```suggestion
from ml_peg.models import current_models
from ml_peg.models.get_models import get_model_names
```
This has moved, sorry!

ml_peg/app/superacids/HF_SbF5_density/app_HF_SbF5_density.py
```diff
@@ -0,0 +1,74 @@
...
+from dash.html import Div
+
+from ml_peg.app import APP_ROOT
+from ml_peg.app.base_app import BaseApp
+from ml_peg.app.utils.build_callbacks import (
+    plot_from_table_column,
+)
+from ml_peg.app.utils.load import read_plot
+from ml_peg.models.get_models import get_model_names
+from ml_peg.models.models import current_models
```
ElliottKasoar:
```suggestion
from ml_peg.models import current_models
from ml_peg.models.get_models import get_model_names
```
This has moved, sorry!

ml_peg/analysis/superacids/HF_SbF5_density/metrics.yml
```diff
@@ -0,0 +1,7 @@
+metrics:
+    MAPE:
+        good: 0.0
+        bad: 10.0
```
ElliottKasoar: Apologies if this has been discussed already, but are you familiar with what these represent? If so, great, but we're also happy to discuss further

ElliottKasoar (commented):
Hi @mattiaperr, thanks again for all the work on this! 

In addition to the above comments, please can you take a look at our new filtering guidelines: https://ddmms.github.io/ml-peg/developer_guide/filter.html.

The main things to check are that:

1) The calculation is robust to errors (or this is suitably handled in analysis)
2) The analysis saves an info.json file along with tables etc
3) The analysis handles missing/incomplete data from models (and usually gives them a score of NaN/None)
4) The app reads the info.json file

I also wondered if you think there's anything else that would be interesting to visualise, beyond the scatter plot?

ElliottKasoar:
> thanks!
> 
> I estimate 6 GPU hours for the pure HF system and 2 GPU hours for the pure SbF5 system, for a model that includes dispersion. What do you think?
> 
> I wanted to follow up on this to discuss the structure of the test I have developed, and currently use, a routine to estimate the HF structure factor and compare it with experimental data, in the style of the test: #445 . Can I integrate such test into these same scripts, or should I create a separate calc/analysis/application within my same benchmark? This would consist in a single NVT simulation at experimental density, which is much faster (25 ps to obtain the structure factor versus 100 ps for the density).
> 
> Given the concern about long simulation times, the whole benchmark could be reformulated as follows:
> 
> * if the HF density is too computationally expensive, use the structure factor as a metric (whose first peak position correlates with the density, and this fact can be taken into account as done in [Add WiSE electrolyte benchmark (density, X-ray S(q), Li-O RDF) #445](https://github.com/ddmms/ml-peg/pull/445)); and then perform only a single NPT simulation of the HF/Sbf5 density at a single point on the diagram.
> 
> What do you think?

Sorry for such a delayed response. I've been working to introduce the element filtering I mention above, and wanted that in place before revisiting things to avoid things getting out of sync, and then I was away for quite a lot of June.

Personally I don't think it hurts to estimate the runtime, however quick it is, but I'd say that's definitely long enough that it's worth being clear.

Regarding the structure factor, I think it's somewhat up to you. In most cases, I'd say if it requires a different calculation (e.g. an entirely new MD simulation), rather than further analysis on the same data, it would usually be a different benchmark (so separate calc/analysis/app). If it's looking at exactly the same structure, we do have some examples of doing multiple types of calculation (e.g. iron_properties), so as long as it's set up in a way where we can run them individually if necessary (which might just be because one is labelled very slow and one isn't), we could do them all together.

I don't think we're overly concerned by the computational cost, we have tests which take days on GPU, I think @joehart2001 was making sure it's set up to run smoothly given the cost and that we document it, so as long as the cost tells us something interesting, I'd be inclined to keep it (but we can structure factor too).

joehart2001 (changes requested):

ml_peg/app/superacids/superacids.yml
```diff
@@ -0,0 +1,2 @@
+title: Superacids
+description: Properties of HF/SbF5 superacid systems
```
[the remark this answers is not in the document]
mattiaperr: you are right, done

ml_peg/calcs/superacids/HF_SbF5_density/calc_HF_SbF5_density.py
mattiaperr: correct, 300 steps is enough

ml_peg/calcs/superacids/HF_SbF5_density/calc_HF_SbF5_density.py
mattiaperr: yes, precision="low" is what we want for these kinds of tests. thanks

ml_peg/calcs/superacids/HF_SbF5_density/calc_HF_SbF5_density.py
mattiaperr: the two calc (density and structure factor) now handle restarting, as in the other densities tests.

ml_peg/analysis/superacids/HF_SbF5_density/metrics.yml
mattiaperr: I tuned the thresholds of the two tests to new values

mattiaperr:
Thank you for your review!

The superacids category now has the density benchmark plus a separate one for the structure factor.
  
Main things implemented since the last review:
 - restart handling for the MD runs
- errors are caught rather than raised
- element filtering: the analyses write info.json and both apps read it

As we discussed a while back, I used the Python MDANSE library to compute S(q). It is declared in pyproject.toml.
I am attaching HF_structure.zip that contains the input for runnining the structure factor benchmark.
[HF_structure.zip](https://github.com/user-attachments/files/30508708/HF_structure.zip)


One thing to flag on shared code (analysis/utils/decorators.py): I added an `hlines` option to `plot_scatter` to draw the experimental target on the density-vs-time plots. It defaults to None, so no existing plot changes. The edits to this shared code are on a separate commit


Tested on mace-mp-0a, mace-mp-0b3 and mace-mpa-0. If this all looks fine, let me know whether you want to run the full set of models yourselves or whether I should do it.

I am also attaching the app JSONs, so you can look at the results: extract it into ml_peg/app/data/ and it creates the superacids/ folder.
[superacids.zip](https://github.com/user-attachments/files/30508645/superacids.zip)

docs/source/user_guide/benchmarks/index.rst
```diff
@@ -19,3 +19,7 @@ Benchmarks
     supramolecular
     surfaces
     tm_complexes
+    conformers
+    molecular_dynamics
+    defect
+    superacids
```
ElliottKasoar, line 25:
```suggestion
    superacids
    supramolecular
    surfaces
    tm_complexes
```
I think a rebase/merge went a little wrong here, since I quite recently reordered these (alphabetically).

ml_peg/analysis/superacids/HF_structure/metrics.yml
```diff
@@ -0,0 +1,19 @@
...
+        bad: 0.5
+        unit: null
+        tooltip: "sum|S_exp - S_calc| / sum|S_exp| for the total neutron structure factor of liquid DF, from the first experimental point to 4 1/A. Experimental reference: McLain et al., Angew. Chem. Int. Ed. 43, 1952 (2004)"
+        level_of_theory: Experimental
+        weight: 1.0
+    First Peak Position Error:
+        good: 0.05
+        bad: 0.8
+        unit: "1/A"
+        tooltip: "Absolute error in the position of the maximum of S(q), located over the same range as the R-factor. Experimental reference: McLain et al., Angew. Chem. Int. Ed. 43, 1952 (2004)"
```
ElliottKasoar, line 17:
```suggestion
        tooltip: "sum|S_exp - S_calc| / sum|S_exp| for the total neutron structure factor of liquid DF, from the first experimental point to 4 1/A."
        level_of_theory: Experimental
        weight: 1.0
    First Peak Position Error:
        good: 0.05
        bad: 0.8
        unit: "1/A"
        tooltip: "Absolute error in the position of the maximum of S(q), located over the same range as the R-factor."
```
We probably don't need the references in the tooltip, since these should be in the docs, which will be linked as part of the benchmark section of the app?

ml_peg/app/superacids/HF_structure/app_HF_structure.py
```diff
@@ -0,0 +1,81 @@
...
+    # Create Dash app
+    full_app = Dash(__name__, assets_folder=DATA_PATH.parent.parent)
+
+    # Construct layout and register callbacks
+    app = get_app()
+    full_app.layout = app.layout
+    app.register_callbacks()
+
+    # Run app
+    full_app.run(port=8057, debug=True)
```
ElliottKasoar, line 81:
```suggestion
```
We don't actually need/use this anymore

ml_peg/app/superacids/HF_SbF5_density/app_HF_SbF5_density.py
```diff
@@ -0,0 +1,85 @@
...
+from ml_peg.app.utils.build_callbacks import (
+    plot_from_table_column,
+)
+from ml_peg.app.utils.load import read_plot
+from ml_peg.models import current_models
+from ml_peg.models.get_models import get_model_names
+
+# Get all models
+MODELS = get_model_names(current_models)
+BENCHMARK_NAME = "HF/SbF5 Mixture Densities"
```
ElliottKasoar, line 19:
```suggestion
from ml_peg.app.utils.load import read_plot

BENCHMARK_NAME = "HF/SbF5 Mixture Densities"
```
I don't think these are actually used?

ml_peg/app/superacids/HF_SbF5_density/app_HF_SbF5_density.py
```diff
@@ -0,0 +1,85 @@
...
+    # Create Dash app
+    full_app = Dash(__name__, assets_folder=DATA_PATH.parent.parent)
+
+    # Construct layout and register callbacks
+    app = get_app()
+    full_app.layout = app.layout
+    app.register_callbacks()
+
+    # Run app
+    full_app.run(port=8056, debug=True)
```
ElliottKasoar, line 85:
```suggestion
```
We don't actually need/use this anymore

ml_peg/app/superacids/HF_structure/app_HF_structure.py
```diff
@@ -0,0 +1,81 @@
...
+from ml_peg.app.utils.build_callbacks import (
+    plot_from_table_column,
+)
+from ml_peg.app.utils.load import read_plot
+from ml_peg.models import current_models
+from ml_peg.models.get_models import get_model_names
+
+# Get all models
+MODELS = get_model_names(current_models)
+BENCHMARK_NAME = "HF Structure Factor"
```
ElliottKasoar, line 19:
```suggestion
from ml_peg.app.utils.load import read_plot

BENCHMARK_NAME = "HF Structure Factor"
```
I don't think these are used?

ElliottKasoar (commented):
Hi @mattiaperr, thanks for all the changes, and for sharing some example data, it's looking amazing!

The input data should all be available now.

I've left a few pretty trivial suggestions, but otherwise I think this is pretty much good to go.

Generally we offer to all the benchmarks for you, but it might be useful if you have spare compute to at least make a start, especially since otherwise you can't write up a section for the paper, and we have quite a few quite complex benchmarks already on the go (@joehart2001 may have some capacity?).

ml_peg/analysis/superacids/HF_structure/analyse_HF_structure.py
```diff
@@ -0,0 +1,391 @@
...
+        download_s3_data(
+            key="inputs/superacids/HF_structure/HF_structure.zip",
+            filename="HF_structure.zip",
+        )
+        / "HF_structure"
+    )
+
+    data = np.loadtxt(hf_structure_dir / "SQ_EXP.dat")
+
+    return data[:, 0], data[:, 1]
```
ElliottKasoar, line 90: Assuming these are pretty small, what we normally do is use something like `shutil.copy` to copy this to the calc outputs directory (e.g. outputs/references, rather than outputs/[model], so the analysis can be completely offline

joehart2001:
hey @mattiaperr just following up

PR #416, opened 2026-03-08
Density scatter -> structures

### LINKED ISSUE #322: Density plots show structures at one point using a trajectory
### Summary

each point may correspond to e.g. 500 structures -> use a trajectory for each collection of points

### Example

_No response_

### Alternatives

_No response_

### Additional context

_No response_

joehart2001:

## Summary
Added point -> structures visulation for the density scatter
This works by writing an extxyz file for each point in the density scatter, which is visualised as a traj when clicked on

Also wrote a more general sampling function `sample_density_grid` used for both density scatter plot and the new write_density_trajectories. previously was just used in the density scatter decorator so have replaced this.

## Linked issue

Resolves #322 

## Testing
Currently RDB7, and will do more

### DIFF
--- a/ml_peg/analysis/utils/utils.py
+++ b/ml_peg/analysis/utils/utils.py
@@ -167,13 +167,18 @@ def rmse(ref: list, prediction: list) -> float:
     return mean_squared_error(ref, prediction)
 
 
+DENSITY_GRID_SIZE = 80
+DENSITY_MAX_POINTS_PER_CELL = 5
+DENSITY_SAMPLE_SEED = 0
+
+
 def sample_density_grid(
-    ref_vals: list[float],
-    pred_vals: list[float],
+    ref_vals: list[float] | np.ndarray,
+    pred_vals: list[float] | np.ndarray,
     *,
-    grid_size: int = 80,
-    max_points_per_cell: int = 5,
-    seed: int = 0,
+    grid_size: int = DENSITY_GRID_SIZE,
+    max_points_per_cell: int = DENSITY_MAX_POINTS_PER_CELL,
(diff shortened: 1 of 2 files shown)


ml_peg/app/non_covalent_interactions/NCIA_HB300SPXx10/app_NCIA_HB300SPXx10.py
```diff
@@ -7,8 +7,11 @@
 
 from ml_peg.app import APP_ROOT
 from ml_peg.app.base_app import BaseApp
-from ml_peg.app.utils.build_callbacks import plot_from_table_cell
-from ml_peg.app.utils.load import read_density_plot_for_model
+from ml_peg.app.utils.build_callbacks import (
+    plot_from_table_cell,
+    struct_from_scatter,
+)
```
ElliottKasoar:
```suggestion
from ml_peg.app.utils.build_callbacks import plot_from_table_cell, struct_from_scatter
```

ml_peg/analysis/utils/utils.py
```diff
@@ -165,6 +167,78 @@ def rmse(ref: list, prediction: list) -> float:
     return mean_squared_error(ref, prediction)
 
 
+def sample_density_grid(
+    ref_vals: list[float],
+    pred_vals: list[float],
+    *,
+    grid_size: int = 80,
+    max_points_per_cell: int = 5,
+    seed: int = 0,
```
ElliottKasoar: Given that these have to match `plot_desity_scatter` (and I don't think this is called anywhere else) perhaps we get rid of the defaults - less risk of mismatches?

PR #420, opened 2026-03-10
Add table and plot download button for csv, png, svg, html

### LINKED ISSUE #218: download table data
### Summary

download data from tables in csv svg etc

### Example

_No response_

### Alternatives

_No response_

### Additional context

_No response_

joehart2001:

## Summary
Created a better way of doing table downloads which uses less code  + added plot downlaods

TODO: (maybe in another PR) asset downlaods e.g. phonon dispersions, diatomics etc and maybe structure downlaods?

## Linked issue

Resolves #218


.gitignore
```diff
@@ -40,4 +40,6 @@ pip-wheel-metadata/
 __pycache__/
 ml_peg/app/data/*
 !ml_peg/app/data/onboarding/
+!ml_peg/app/data/table_download/
+!ml_peg/app/data/table_download/*
```
ElliottKasoar: I don't think both of these should be needed? Git doesn't track directories, only files

ml_peg/app/data/table_download/controls.css
ElliottKasoar, line 1: Is this loaded automatically? I don't see any references to it

I don't fully understand what this does that isn't done by `container_style` etc below?

ElliottKasoar (commented):
This works really well!

So does this only make it possible to download tables, and not the plots etc? Is there a way we could extend it to the full set of test/category outputs?

ml_peg/app/data/table_download/controls.css
joehart2001: changing the container style does not edit the size of the buttons so that we can make them the same size as it only controls the outer wrapper layout around the dropdown/butto to the  CSS targets the dropdown’s generated inner elements, plus button hover/active/focus states, which can’t be handled from the wrapper style

joehart2001: Dash loads this automatically because ml_peg/app/data is passed as the app’s assets_folder, so CSS files under that directory are picked up without an explicit import

joehart2001:
@ElliottKasoar opinions on structure downloads? we can do images but maybe even xyz files too. could be cool but can come under the asset download PR

ml_peg/app/data/table_download/controls.css
ElliottKasoar: Ah ok cool thanks!

ElliottKasoar:
> @ElliottKasoar opinions on structure downloads? we can do images but maybe even xyz files too. could be cool but can come under the asset download PR

WEAS should have some sort of download option: https://github.com/superstar54/weas-widget/pull/13 / https://weas-widget.readthedocs.io/en/latest/gui.html#buttons

We may just need to enable the GUI?

Then I agree being able to download the file itself would be nice.

ml_peg/app/utils/build_callbacks.py
```diff
@@ -23,9 +23,31 @@
     PERIODIC_TABLE_POSITIONS,
     PERIODIC_TABLE_ROWS,
 )
+from ml_peg.app.utils.build_components import build_plot_download_controls
+from ml_peg.app.utils.register_callbacks import register_plot_download_callbacks
 from ml_peg.app.utils.weas import generate_weas_html
 
 
+def _plot_with_download_controls(graph: Graph) -> Div:
```
ElliottKasoar:
```suggestion
def plot_with_download_controls(graph: Graph) -> Div:
```
This is something we're seemingly expecting "users" to directly call, which `_` implies would not be the case.

ml_peg/app/utils/build_callbacks.py
```diff
@@ -66,7 +89,7 @@ def show_plot(active_cell) -> Div:
         column_id = active_cell.get("column_id", None)
         if column_id:
             if column_id in column_to_plot:
-                return Div(column_to_plot[column_id]), None
+                return _plot_with_download_controls(column_to_plot[column_id]), None
```
ElliottKasoar:
```suggestion
                return plot_with_download_controls(column_to_plot[column_id]), None
```
If you agree with the above

ml_peg/app/utils/build_callbacks.py
```diff
@@ -131,7 +155,7 @@ def show_plot(active_cell, current_table_data) -> Div:
                 pass  # Fall through to normal handling
 
         if row_id in cell_to_plot and column_id in cell_to_plot[row_id]:
-            return Div(cell_to_plot[row_id][column_id]), None
+            return _plot_with_download_controls(cell_to_plot[row_id][column_id]), None
```
ElliottKasoar:
```suggestion
            return plot_with_download_controls(cell_to_plot[row_id][column_id]), None
```
If you agree with the above

ml_peg/app/utils/build_callbacks.py
```diff
@@ -177,7 +202,7 @@ def show_plot(click_data) -> Div:
         idx = click_data["points"][0]["pointNumber"]
 
         if idx >= 0 and idx < len(plots_list):
-            return Div(plots_list[idx])
+            return _plot_with_download_controls(plots_list[idx])
```
ElliottKasoar:
```suggestion
            return plot_with_download_controls(plots_list[idx])
```
If you agree with the above

ml_peg/app/utils/register_callbacks.py
```diff
@@ -24,6 +36,8 @@
     get_scores,
 )
 
+_PLOT_DOWNLOAD_CALLBACK_REGISTERED = False
```
ElliottKasoar: Why is this needed?

My impression is globals in Dash are not particularly reliable for various reasons, so I think we probably want to avoid this.

ElliottKasoar (approved):

PR #421, opened 2026-03-10
Volume scans going to very high densities

### LINKED ISSUE #308: Volume scans going to very high densities
### Summary

Probe behaviour of the higher body-order terms at close-approach, similar to the diatomic test for the two-body. Uniformly compress a few (~10 in total) small (<=20 atoms) prototype crystal structures + some randomly generated crystals for each element. Possible extension to ~100 randomly generated structures with 2, 3 and 4 elements to probe (some) multi-element terms.

Can also go to low densities and check there is a smooth transition to isolated atoms.

### Interactive features

Energy vs density / shortest-interatomic distance on log scale. 

### Category

Physicality

### Data availability

No reference data available at present. Probably not required as can use similar metrics to the diatomic test e.g. number of force flips

### Computational cost

~10-30x the diatomic test. 

### Additional context

_No response_

JPDarby:

## Summary

This is similar to the diatomic test for the two-body but uses 3D crystals to probe higher body terms.  For single element structures, 5 prototypes are formed (fcc, sc, bcc, hcp and diamond). A  further 5 structures are randomly generated, subject to sensible constraints, using the pyxtal package. Multi-element terms are probed by picking 10 random binary and 10 random ternary compositions, then forming 10 random structures for each composition. The structures are then isotropically scaled across a wide range, covering extremely high and low densities. The energy and pressure are computed for each structure and these are used to form similar metrics to the diatomic test e.g. Pressure sign flips etc. 

## Linked issue

Resolves #308 

## Progress

- [x] Calculations
- [x] Analysis
- [x] Application
- [x] Documentation

## Testing

Tested all models below using a subset of the elements. The 
- `mace-mp-0a`
- `mace-mp-0b3`
- `mace-mpa-0 `
- `mace-omat-0`
- `mace-matpes-r2scan`
- `orb-v3-consv-inf-omat`
- `pet-mad`

## Discussion Points

- I've included the code to generate all structures in calc_compression.py. They are then generated by running this file directly. This could be removed but including the code here allows the benchmark to be tweaked more easily in the future.
- Currently showing one plot at a time (due to multiple curves on each plot) rather than the periodic table style used to show all dimer curves. This could be adjusted.
- Some materials should have more than one minimum in the (isotropic scaling) energy vs volume curve. See this [preprint](https://arxiv.org/html/2506.03578v1#:~:text=Cerium%20shows%20an%20isostructural%20%CE%B1,electrons%20in%20each%20phase%2C%20respectively.) on Ce. Given this, some of the metrics, e.g. number of pressure flips, are "usually" useful but break down in some rare cases. The "Holes" metric does not suffer from this issue and identifies many-body holes at close approach in multiple models.


docs/source/user_guide/benchmarks/physicality.rst
```diff
@@ -135,3 +135,64 @@ Data availability
 -----------------
 
 None required; diatomics are generated in ASE.
+
+
+Compression
+=========
```
ElliottKasoar:
```suggestion
Compression
===========
```
sphinx warns that this is too short

ml_peg/analysis/physicality/compression/metrics.yml
```diff
@@ -0,0 +1,41 @@
...
+  Energy minima:
+    good: 1.0
+    bad: 5.0
+    unit: null
+    tooltip: Average number of energy-per-atom minima per structure. Tolerance of 1 meV/atom is used.
+  # Energy inflections:
+  #   good: 1.0
+  #   bad: 5.0
+  #   unit: null
+  #   tooltip: Average number of energy curvature sign changes per structure
```
ElliottKasoar:
```suggestion
```
If no longer needed?

ml_peg/calcs/physicality/compression/calc_compression.py
```diff
@@ -0,0 +1,688 @@
...
+DATA_PATH = Path(__file__).parent / "data"
+
+# Benchmark configuration
+# Compression/expansion factors applied uniformly to the cell volume.
+# A factor of 1.0 corresponds to the equilibrium structure.
+MIN_SCALE = 0.25
+MAX_SCALE = 3.0
+N_POINTS = 100
+
+# ELEMENTS: list[str] = [symbol for symbol in chemical_symbols if symbol]
```
ElliottKasoar: Thanks for adding these. So we don't forget, can you swap everything over to the "production" settings?

ml_peg/calcs/physicality/compression/calc_compression.py
ElliottKasoar, line 1: If this is very slow even on a GPU, you could consider some sort of parameterisation, similar to #388, to make it easier to parallelise.

ml_peg/calcs/physicality/compression/calc_compression.py
```diff
@@ -0,0 +1,688 @@
...
+        ASE ``Atoms`` with periodic boundary conditions and a cell
+        scaled so that nearest-neighbour distances match the sum of
+        covalent radii.
+
+    Raises
+    ------
+    RuntimeError
+        If no valid structure could be generated within *max_attempts*.
+    """
+    from pyxtal import pyxtal as pyxtal_cls
```
ElliottKasoar, line 233: Can you add this as an (optional) dependency in the pyproject.toml?

ml_peg/calcs/physicality/compression/calc_compression.py
```diff
@@ -0,0 +1,688 @@
...
+        "config": {
+            "min_scale": MIN_SCALE,
+            "max_scale": MAX_SCALE,
+            "n_points": N_POINTS,
+        },
+    }
+    (write_dir / "metadata.json").write_text(json.dumps(metadata, indent=2))
+
+
+@pytest.mark.slow
```
ElliottKasoar:
```suggestion
@pytest.mark.very_slow
```
If it's up to a day on GPU, probably

ml_peg/calcs/physicality/compression/calc_compression.py
```diff
@@ -0,0 +1,688 @@
...
+    run_compression(model_name, MODELS[model_name])
+
+
+if __name__ == "__main__":
+    # generate the random structures and save to data directory
+    # this only needs to be done once
+    # code included here so that the structures can be easily
+    # regenerated in the same way if needed, or more can be
+    # created etc.
+    _generate_all_random(
```
ElliottKasoar, line 696: Assuming this is relatively cheap compared to the actual test, I'd probably call this as part of the actual test, to make it more consistent in how we run it

ElliottKasoar (commented):
Thanks for this, @JPDarby!

I need to go through everything much more carefully, but generally it's looking great!

I've left some (largely trivial) suggestions for now

ml_peg/calcs/physicality/compression/calc_compression.py
JPDarby: Done

ml_peg/calcs/physicality/compression/calc_compression.py
JPDarby: Done

ElliottKasoar:
Also note there are a couple of conflicting files to resolve, both hopefully pretty straightforward since it doesn't involve the code.

ml_peg/calcs/physicality/compression/calc_compression.py
JPDarby: Yep it's pretty cheap, have moved it into run_compression + added a check so it won't regenerate the random structures if they already exist locally.

ml_peg/calcs/physicality/compression/calc_compression.py
JPDarby: So this allows different models to be run in parallel? The structure generation (now moved to run_compression) uses pyxtal and I can't guarantee that will be deterministic and produce the same random structures each time. Given this, I think it's best to avoid this for now. If the test really is too slow we could move back to generating the structures once and for all and then parametrize.

ElliottKasoar:
Hi @JPDarby, thanks again for the work on this!

I'll do another review of everything else this week.

Also if you can resolve the pre-commit errors, and the merge conflict, that would be great!

pyproject.toml
```diff
@@ -34,6 +34,7 @@ dependencies = [
     "mlipx<0.2,>=0.1.5",
     "molify<0.3,>=0.2.2",
     "openpyxl<4,>=3.1.5",
+    "pyxtal>=1.1.3",
```
ElliottKasoar:
```suggestion
    "pyxta<2,l>=1.1.3",
```

ml_peg/calcs/physicality/compression/calc_compression.py
```diff
@@ -0,0 +1,693 @@
...
+from ase.io import write
+import numpy as np
+import pandas as pd
+import pytest
+from pyxtal.tolerance import Tol_matrix
+from pyxtal.database.element import Element
+import tqdm
+
+from ml_peg.models.get_models import load_models
+from ml_peg.models.models import current_models
```
ElliottKasoar:
```suggestion
from ml_peg.models import current_models
from ml_peg.models.get_models import load_models
```
This has moved, sorry!

ml_peg/app/physicality/compression/app_compression.py
```diff
@@ -0,0 +1,175 @@
...
+from dash import Dash, Input, Output, callback, dcc
+from dash.dcc import Loading
+from dash.exceptions import PreventUpdate
+from dash.html import Div, Label
+from plotly.io import read_json
+
+from ml_peg.app import APP_ROOT
+from ml_peg.app.base_app import BaseApp
+from ml_peg.models.get_models import get_model_names
+from ml_peg.models.models import current_models
```
ElliottKasoar:
```suggestion
from ml_peg.models import current_models
from ml_peg.models.get_models import get_model_names
```
This has moved, sorry!

ml_peg/analysis/physicality/compression/analyse_compression.py
```diff
@@ -0,0 +1,686 @@
...
+import plotly.graph_objects as go
+import pytest
+from scipy.signal import find_peaks
+
+from ml_peg.analysis.utils.decorators import build_table
+from ml_peg.analysis.utils.utils import load_metrics_config
+from ml_peg.app import APP_ROOT
+from ml_peg.calcs import CALCS_ROOT
+from ml_peg.models.get_models import get_model_names
+from ml_peg.models.models import current_models
```
ElliottKasoar:
```suggestion
from ml_peg.models import current_models
from ml_peg.models.get_models import get_model_names
```
This has moved, sorry!

ml_peg/calcs/physicality/compression/calc_compression.py
```diff
@@ -0,0 +1,693 @@
...
+    volume (per atom), and stress tensor are recorded.
+
+    Parameters
+    ----------
+    model_name
+        Name of the model being evaluated.
+    model
+        Model wrapper providing ``get_calculator``.
+    """
+    calc = model.get_calculator()
```
ElliottKasoar:
```suggestion
    calc = model.get_calculator(precision="high")
```
This is an option we added somewhat recently, which sets either ~float64 ("high") or ~float32 ("low").

I think when you made this PR, the default would have been equivalent to  "low". The default is now actually "high", but it's good to be explicit either way

ml_peg/calcs/physicality/compression/calc_compression.py
ElliottKasoar: By default using `@pytest.mark.parametrize` still runs sequentially, it just isolates them slightly more. It enables parallelism since we can use the `-k` (see the [docs](https://docs.pytest.org/en/stable/reference/reference.html#cmdoption-k)) to run based on the marker, and/or we can add custom CLI inputs (as we did for the liquids I linked, with the custom `conftest.py` file).

Since you already check that the structures are generated (and can do this quickly with the new `mock` model), we probably could run it fairly safely, but yeah I'm happy to leave it for now too.

ElliottKasoar (commented):
A few more minor suggestions/fixes.

The biggest thing is the changes needed for element filtering I mentioned earlier. I think this is already pretty robust to missing data/errors, so I think the main thing is to make sure we save the info.json file and pass this to the app.

Do you think it would be useful to also visualise the structures we've generated as part of this?

ml_peg/calcs/physicality/compression/calc_compression.py
JPDarby: ok let's leave it for now and return to it later (once this is merged) if it's problematic

JPDarby:
Hi @ElliottKasoar  I've add the changes you recommended, resolved the merge conflicts and precommit errors and have also added in the info.json needed for element filtering. 

Let me know if there's anything else that needs changing and hopefully we can get this merged in soon!

JPDarby:
> Do you think it would be useful to also visualise the structures we've generated as part of this?

Yeah could be nice to see what the local motifs are when things go wrong! Is there a standard way of doing this within ml-peg?

ml_peg/calcs/physicality/compression/calc_compression.py
```diff
@@ -0,0 +1,702 @@
...
+            else:
+                chosen_elements = sorted(
+                    rng.choice(elements, size=n_elements, replace=False).tolist()
+                )
+            # Random atom counts that sum to at most max_atoms (each >= 1)
+            counts = rng.integers(
+                1, max(2, max_atoms - n_elements + 2), size=n_elements
+            )
+            if len(counts) == 1:
+                counts = np.random.choice(
```
joehart2001: this uses the numpy random instead of the seed generator, would this be consistent?

ml_peg/calcs/physicality/compression/calc_compression.py
```diff
@@ -0,0 +1,702 @@
...
+    # The required scale factor is the ratio for the closest pair
+    closest_idx = np.unravel_index(np.argmin(distances), distances.shape)
+    min_dist = distances[closest_idx]
+    target_bond = target_matrix[closest_idx]
+
+    if min_dist <= 0.0:
+        raise ValueError("Degenerate geometry: shortest interatomic distance <= 0")
+
+    scale_factor = target_bond / min_dist
+    atoms.set_cell(atoms.cell * scale_factor, scale_atoms=True)
```
joehart2001: is this the same definition as in the docs file? i think this selects the pair with the smallest distance regardless of atomic radii

joehart2001:
hey @JPDarby, sorry for the delay:
- could you update the `uv.lock` as youve changed the pyproject
- could you check over how the analysis handles NaNs? for some metrics i think a NaN value would work out favourably e.g. for number of minima?

joehart2001:
I also upgraded the composition dropdown from an assortment of Si, CuCs7 etc to have its own categories. i think we can leave structure vis for now and maybe add it in for a later pr:

Screenshot 2026-09-11 at 16 11 11

ml_peg/calcs/physicality/compression/calc_compression.py
JPDarby: good spot! I've pushed a bugfix

ml_peg/calcs/physicality/compression/calc_compression.py
JPDarby: also a good spot! updated to used the seeded rng and checked for this elsewhere

joehart2001:
added structure vis:
Screenshot 2026-09-12 at 14 40 13

joehart2001:
@ElliottKasoar ive tested and seems to work well + element filtering working

PR #429, opened 2026-03-16
Feat/add folmsbee conformer benchmark

### LINKED ISSUE #427: feat: new benchmark with folmsbee conformer dataset
### Summary

The Folmsbee dataset of low-energy conformers of drug-like molecules. The differences in energy are smaller compared to the Wiggle500 dataset and it features a greater number of molecules. The highest available level of theory for energy evaluations to be used as ground-truth is DLPNO-CCSD(T).

### Interactive features

For now just a table with MAE.

### Category

Conformers

### Data availability

Assessing conformer energies using electronic structure and
machine learning methods

Dakota Folmsbee, Geoffrey Hutchinson
International Journal of Quantum Chemistry 2020 121 (1) e26381
DOI: 10.1002/qua.26381

### Computational cost

Low: Only inference of 693 molecules (varying number of conformers) is needed.

### Additional context

This is a test for moving the benchmarks of [mlip-audit](https://github.com/instadeepai/mlipaudit) into this repository. I have included an analysis script for this benchmark, however would like to kindly ask for assistance with building and harmonizing the Dash layout.

lwehrhan:

## Summary

The Folmsbee dataset of low-energy conformers of drug-like molecules. The differences in energy are smaller compared to the Wiggle500 dataset and it features a greater number of molecules. The highest available level of theory for energy evaluations to be used as ground-truth is DLPNO-CCSD(T). This is a test for moving the benchmarks of [mlip-audit](https://github.com/instadeepai/mlipaudit) into this repository. I have included an analysis script for this benchmark, however would like to kindly ask for assistance with building and harmonizing the Dash layout.

## Linked issue

Resolves #427 

## Progress

- [x] Calculations
- [x] Analysis
- [x] Application
- [x] Documentation


joehart2001:
Hey @lwehrhan, we've just merged a PR so that you can tag your benchmark with the mlip-auditl, add your logo and have your own dedicated tab (PR #434). Please see the [framework credit tags docs](https://ddmms.github.io/ml-peg/developer_guide/add_benchmarks.html#framework-credit-tags)

ml_peg/calcs/conformers/Folmsbee/calc_Folmsbee.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    Benchmark the Folmsbee dataset.
+
+    Parameters
+    ----------
+    mlip
+        Name of model and model object to get calculator.
+    """
+    model_name, model = mlip
+    model.default_dtype = "float64"
```
lwalew: **question**: What's the reason for this?

ml_peg/calcs/conformers/Folmsbee/calc_Folmsbee.py
```diff
@@ -0,0 +1,61 @@
...
+    benchmark = MlPegConformerSelectionBenchmark(
+        force_field=calc,
+        data_input_dir=data_input_dir,
+        run_mode="standard",
+    )
+    benchmark.run_model()
+
+    (out_path / "model_output.json").write_text(
+        benchmark.model_output.model_dump_json()
+    )
```
lwalew, line 65: **question**: Why not `json.dump`?

ml_peg/analysis/conformers/Folmsbee/analyse_Folmsbee.py
```diff
@@ -0,0 +1,180 @@
...
+    -------
+    dict[str, float]
+        Dictionary of predicted conformer energies errors for all models.
+    """
+    results = {}
+    for model_name in MODELS:
+        results[model_name] = mae(
+            conformer_energies["ref"], conformer_energies[model_name]
+        )
+    return results
```
lwalew, line 208: IIUC, here we are putting all conformers into a single flat list then taking the mae, whereas in the upstream benchmark we compute per-molecule MAEs which we then average to give `avg_mae`.

ml_peg/analysis/conformers/Folmsbee/analyse_Folmsbee.py
```diff
@@ -0,0 +1,180 @@
...
+
+        result_by_name = {m.molecule_name: m for m in result.molecules}
+        data_by_name = {m.molecule_name: m for m in benchmark._folmsbee_data}
+
+        for label in labels():
+            mol_name, conf_str = label.rsplit("_conf", 1)
+            i = int(conf_str)
+            molecule = result_by_name[mol_name]
+
+            results[model_name].append(float(molecule.predicted_energy_profile[i]))
```
lwalew: This might be None if a molecule had an unsupported element.

lwalew (approved):

pyproject.toml
```diff
@@ -216,3 +223,4 @@ module-root = ""
 
 [tool.uv.sources]
 asemolec = { git = "https://github.com/imagdau/aseMolec.git" }
+mlipaudit = { git = "https://github.com/instadeepai/MLIPAudit.git", branch = "mlpeg-migration" }
```
lwalew, line 231: **remark**: Note that this will have to be udpated to main at some point, either once everything is migrated or per-benchmark. Probably the former.

ml_peg/calcs/conformers/Folmsbee/calc_Folmsbee.py
lwehrhan: it was meant to ensure the same precision for all models, but really is not necessary. I removed it.

joehart2001:
Hey @lwehrhan what’s the status on the unresolved comments? I’ll tag @ElliottKasoar so he can take a look over the benchmark

ml_peg/analysis/conformers/Folmsbee/analyse_Folmsbee.py
lwehrhan: This would be an actual problem in this case. I think we will have to drop the molecules with unsupported elements or skip the benchmark entirely in that case (as in mlip audit). We cannot test for supported elements because the calculators do not expose the supported elements.

lwehrhan:
@joehart2001 I have addressed the comments now. The only open issue (this will also apply to some of the other benchmarks) is that we cannot test the supported elements of the models prior to running the benchmark like we do in mlip audit, which sometimes may lead to bugs. I saw this is handled "on-the-fly" e.g. in `physicality/diatomics`. Here, not supporting molecules is not penalized. Is that the expected behavior for these benchmarks here?

pyproject.toml
lwehrhan: I also agree this should be updated once we have everything migrated.

ElliottKasoar:
> @joehart2001 I have addressed the comments now. The only open issue (this will also apply to some of the other benchmarks) is that we cannot test the supported elements of the models prior to running the benchmark like we do in mlip audit, which sometimes may lead to bugs. I saw this is handled "on-the-fly" e.g. in `physicality/diatomics`. Here, not supporting molecules is not penalized. Is that the expected behavior for these benchmarks here?

Hi @leonwehrhan, can you please take a look at the (very new) guidance for element filtering: https://ddmms.github.io/ml-peg/developer_guide/filter.html.

Essentially, in an ideal world, we'd catch any errors that occur during calculations, so you shouldn't need to know what calculators support, but have an awareness that they failed for certain molecules, which could be due to element support or otherwise.

Then in the analysis, if it fails for any of the molecules, the model can't do the benchmark, so it would get a score of `NaN`/`None`. The current implementation of filtering allows us to exclude this benchmark if any matches are found to elements required in the entire benchmark, but the aim is to allow specific molecules to be excluded - see https://github.com/ddmms/ml-peg/pull/625 for an example.

If it wouldn't be possible to set things up to do this partial filtering (e.g. if you can't catch errors and continue for other molecules), it's also fine to only implement the current 'binary' filter, but we would still rather say a model can't do a benchmark than have the scores represent different things for different models.

The diatomics example is probably not the best guide, since it came before we added filtering, and so the approach is not quite as consistent.

ml_peg/calcs/conformers/Folmsbee/calc_Folmsbee.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    Benchmark the Folmsbee dataset.
+
+    Parameters
+    ----------
+    mlip
+        Name of model and model object to get calculator.
+    """
+    model_name, model = mlip
+    calc = model.get_calculator()
```
ElliottKasoar, line 43:
```suggestion
    calc = model.get_calculator(precision="high")
```
This may not have existed when you initially wrote this, but we encourage explicitly setting this, and generally for static tests, use "high" precision (float64, where possible), which is the current default anyway

ml_peg/analysis/conformers/Folmsbee/analyse_Folmsbee.py
```diff
@@ -0,0 +1,190 @@
...
+    Returns
+    -------
+    dict[str, list]
+        Dictionary of all reference and predicted barrier heights.
+    """
+    results = {"ref": []} | {mlip: [] for mlip in MODELS}
+    ref_stored = False
+
+    data_input_dir = download_s3_data(
+        key="inputs/conformers/Folmsbee/conformer_selection.zip",
```
ElliottKasoar:
```suggestion
        key="inputs/conformers/Folmsbee/Folmsbee.zip",
```
We normally use this naming convention for consistency (and this is how it's currently been uploaded)

ml_peg/analysis/conformers/Folmsbee/metrics.yml
```diff
@@ -0,0 +1,7 @@
+metrics:
+  MAE:
+    good: 0.0
+    bad: 20.0
```
ElliottKasoar: How much consideration has gone into these/are you familiar with the principles we have in mind for setting them?

If you're happy with them, great! We're happy to clarify/discuss more though

ElliottKasoar:
Given how many points are in the scatter plot, I also wonder if we may want to use the density scatter instead? I think beyond a handful of models the plots will get very crowed otherwise?

ElliottKasoar:
Also note there is now a small conflict due to changes we've made to frameworks.yml. Please could this be resolved?

ml_peg/analysis/conformers/Folmsbee/metrics.yml
lwehrhan: Hi @ElliottKasoar. These were indeed placeholders copied from the `DipCONFS` benchmark, I think a "bad" threshold of 2.0 kcal/mol is more sensible. For the final scoring, we use different scoring functions in MLIP Audit that differ per benchmark, e.g. here working with a per-molecule threshold. I have pushed an update to the `metrics.yaml` and `analyse_folmsbee.py` files that would read the MLIP Audit scores directly from the benchmark analysis and use this for the final score computation. This would allow easier maintainability from our side and also align the benchmark scores of MLIP audit and ML-PEG. MAE and other metrics are still presented in the metrics table. Is this acceptable for the ML-PEG implementation of the MLIP audit benchmarks?

lwehrhan:
> Given how many points are in the scatter plot, I also wonder if we may want to use the density scatter instead? I think beyond a handful of models the plots will get very crowed otherwise?

my understanding is that this would mean dropping the per-conformer structure viewer?

joehart2001:
> > Given how many points are in the scatter plot, I also wonder if we may want to use the density scatter instead? I think beyond a handful of models the plots will get very crowed otherwise?
> 
> my understanding is that this would mean dropping the per-conformer structure viewer?

we can still visualise structures with the denity plots (check e.g. elasticity benchmark), but we could also use the `plot_from_table_cell` decorator to do a single scatter plot per model instead of all the models' scatters being combined into one. opinions @ElliottKasoar ?

pyproject.toml
```diff
@@ -64,6 +64,9 @@ mace = [
 mattersim = [
     "mattersim==1.2.2",
 ]
+mlipaudit = [
+    "mlipaudit; python_version >= '3.11'",
```
joehart2001, line 68: dm-tree (via the mlipaudit extra: mlipaudit -> mlip -> dm-tree) ships no Python 3.13 wheel so we get an error for this with 3.13
```suggestion
    "mlipaudit; python_version >= '3.11' and python_version < '3.13'",
```

ml_peg/calcs/conformers/Folmsbee/calc_Folmsbee.py
```diff
@@ -0,0 +1,60 @@
...
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any
+
+import pytest
+
+from ml_peg.calcs.utils.mlipaudit import MlPegConformerSelectionBenchmark
```
joehart2001, line 22: if mliap audit is not installed and `ml_peg calc` is run then we will fail on this benchmark so we can skip this more gracefully

```suggestion
# Optional extra (ml-peg[mlipaudit]); skip if not installed.
pytest.importorskip("mlipaudit")

from ml_peg.calcs.utils.mlipaudit import (  # noqa: E402
    MlPegConformerSelectionBenchmark,
)
```

ml_peg/analysis/conformers/Folmsbee/analyse_Folmsbee.py
```diff
@@ -0,0 +1,235 @@
...
+from __future__ import annotations
+
+from pathlib import Path
+
+from ase import Atoms
+from ase.calculators.calculator import Calculator
+from ase.io import write
+from mlipaudit.benchmarks.conformer_selection.conformer_selection import (
+    ConformerSelectionModelOutput,
+)
```
joehart2001, line 19:
```suggestion

```

ml_peg/analysis/conformers/Folmsbee/analyse_Folmsbee.py
```diff
@@ -0,0 +1,235 @@
...
+
+from ml_peg.analysis.utils.decorators import build_table, plot_parity
+from ml_peg.analysis.utils.utils import (
+    build_dispersion_name_map,
+    load_metrics_config,
+    mae,
+)
+from ml_peg.app import APP_ROOT
+from ml_peg.calcs import CALCS_ROOT
+from ml_peg.calcs.utils.mlipaudit import MlPegConformerSelectionBenchmark
```
joehart2001, line 29:
```suggestion
```

ml_peg/analysis/conformers/Folmsbee/analyse_Folmsbee.py
```diff
@@ -0,0 +1,235 @@
...
+from ml_peg.calcs.utils.utils import download_s3_data  # noqa: F401
+from ml_peg.models import current_models
+from ml_peg.models.get_models import load_models
+
```
joehart2001, line 32:
```suggestion

# Optional extra (ml-peg[mlipaudit]); skip if not installed.
pytest.importorskip("mlipaudit")

from mlipaudit.benchmarks.conformer_selection.conformer_selection import (  # noqa: E402
    ConformerSelectionModelOutput,
)

from ml_peg.calcs.utils.mlipaudit import (  # noqa: E402
    MlPegConformerSelectionBenchmark,
)

```

ml_peg/calcs/conformers/Folmsbee/calc_Folmsbee.py
joehart2001, line 1: is it possible to add some form of progress bar?

ml_peg/analysis/conformers/Folmsbee/analyse_Folmsbee.py
```diff
@@ -0,0 +1,235 @@
...
+def analyze_results() -> dict:
+    """
+    Run the mlipaudit analysis for each model.
+
+    Returns
+    -------
+    dict
+        Mapping of model name to ``(benchmark, ConformerSelectionResult)``.
+    """
+    data_input_dir = download_s3_data(
```
joehart2001: do we need to download this again?

ml_peg/analysis/conformers/Folmsbee/analyse_Folmsbee.py
joehart2001: @ElliottKasoar ?

docs/source/user_guide/benchmarks/conformers.rst
```diff
@@ -40,3 +40,49 @@ Reference data:
...
+to the lowest-energy reference conformer (set to zero). The predicted energies
+are aligned to the same reference conformer and converted to kcal/mol. The
+mean
+absolute error (MAE) between predicted and reference relative energies is
+reported, averaged across all systems.
+
+Computational cost
+------------------
+
+Medium: tests may take hours on CPU.
```
joehart2001: what about gpu? also we need to improve our guidance on this but usually we report the time for a single model to be benchmarked, so this would be on the order of minutes/tens of minues for cpu

docs/source/user_guide/benchmarks/conformers.rst
```diff
@@ -40,3 +40,49 @@ Reference data:
...
+
+Performance in predicting relative conformer energies for a set of drug-like
+molecules. Each molecule has a small set of conformers (3-10 per molecule)
+whose energies are ranked relative to the lowest-energy conformer. Reference
+data from DLPNO-CCSD(T) calculations.
+
+Metrics
+-------
+
+1. Conformer energy error
```
joehart2001: update this to include the MAE and conformer score as two separate metrics as in the `metrics.yml`

joehart2001:
I think its looking nearly ready from my side of things. thanks for contributing this!

ml_peg/analysis/conformers/Folmsbee/metrics.yml
ElliottKasoar: That sounds great, thanks!

ml_peg/analysis/conformers/Folmsbee/analyse_Folmsbee.py
ElliottKasoar: I tried to address this a little in the my comment below about filtering, but I think this is fine - if a model is unable to do the test because it's missing any elements required, ML-PEG would set the score to None/NaN as well.

I think the only thing we'd need in terms of filtering is similarly to how you build the `labels` currently, if you're also able to get the full set of elements, save this in an info.json file, and pass it to the app, so we can enable filtering for this benchmark?

It doesn't need to be a list per structure, since we don't explicitly loop through them in ML-PEG, so unless the per-structure data is accessible somehow to reevaluate the score, there's no benefit to the future plan for "partial" filtering, and it'll be a binary "is this element present in any of the structures or not" - if is is contained, then the benchmark can have "failed" without impacting the overall score for all the benchmarks.

joehart2001:
@lwehrhan just a ping on this, would be great to get this in soon to trial out the framework featuring in the live site

ElliottKasoar (commented):
Hi @lwehrhan, I've made some changes, a lot of which were based on @joehart2001's suggestions, including:

- Save the elemental info during analysis (there may be a neater way to do this, and maybe to only save the mock structures for visualisation)
- Allow analysis if some models have no data/datafile
- Skip benchmarks if mlipaudit is not installed
- Copy the reference data, to avoid needing to recalculate (despite the cache, it's useful for outputs to be self-contained)

@joehart2001 if you can do a final review since I've touched this in quite a few places (e.g. run end-to-end with at least one model), but from my perspective I think it's looking great!

(Addressing the remaining comments for the docs, and checking you're happy with the changes I made, would be appreciated, @lwehrhan!)

joehart2001:
> Hi @lwehrhan, I've made some changes, a lot of which were based on @joehart2001's suggestions, including:
> 
> * Save the elemental info during analysis (there may be a neater way to do this, and maybe to only save the mock structures for visualisation)
> * Allow analysis if some models have no data/datafile
> * Skip benchmarks if mlipaudit is not installed
> * Copy the reference data, to avoid needing to recalculate (despite the cache, it's useful for outputs to be self-contained)
> 
> @joehart2001 if you can do a final review since I've touched this in quite a few places (e.g. run end-to-end with at least one model), but from my perspective I think it's looking great!
> 
> (Addressing the remaining comments for the docs, and checking you're happy with the changes I made, would be appreciated, @lwehrhan!)

tested and all good from my side. happy to merge once small docs change is done

lwehrhan:
The changes look good to me, thanks @ElliottKasoar ! And thank you @joehart2001 for the review

joehart2001:
> The changes look good to me, thanks @ElliottKasoar ! And thank you @joehart2001 for the review

thanks for your work @lwalew @leonwehrhan!

ElliottKasoar (approved):

PR #437, opened 2026-03-23
Split vacancies benchmark

### LINKED ISSUE #335: Add Split Vacancy Defect Tests
### Summary

Formation energies of split-vacancy defects from regular vacancy defects.

Based on doi.org/10.1088/2515-7655/ade916

### Interactive features

View relaxed defect structures, with DFT and/or model.

### Category

_No response_

### Data availability

Either bulk crystals, or a new category: (bulk?) defects

### Computational cost

~10,000 short relaxations, so could be quite long.

### Additional context

_No response_

ThomasWarford:

## Summary

Based on this paper: [Identifying split vacancy defects with machine-learned foundation models and electrostatics
](doi.org/10.1088/2515-7655/ade916)

The metrics:
- Formation energy of split-vacancy defects from fully ionised defects
- Spearman's coefficient for ranking the energies of initial defect structures.
- RMSD of MLIP relaxed structures vs DFT relaxed structures. (Note the initial structure for the MLIP relaxation is currently the DFT relaxed structure.)

## Linked issue

Closes #335 

## Progress

- [x] Calculations
- [x] Analysis
- [x] Application
- [x] Documentation


kavanase:
I had a look through this code @ThomasWarford. It's very nice!

One comment:
For the `max_dist` description:
> Maximum atomic displacement between the MLIP-relaxed and DFT-relaxed matched
structures, normalised by :math:`(N/V)^{1/3}` (where :math:`N` is the number of
atoms and :math:`V` the cell volume) to give a unitless quantity comparable across
different supercell sizes. Only computed for structure pairs that pass the
StructureMatcher test. The match criterion itself is a normalised max dist below 0.3. 

I would suggest changing "across different supercell sizes" to "across different crystal structures". 'Different supercell sizes' sounds like this number will differ for a defect in e.g. a 2x2x2 MgO supercell and a 4x4x4 supercell, which is not the case. Also just to check, is it not (V/N)^{1/3} ? i.e. average free length per atom (https://github.com/materialsproject/pymatgen/blob/v2025.6.14/src/pymatgen/analysis/structure_matcher.py#L534)

Fyi, if the StructureMatcher scans are very slow to run, the `StructureMatcher_scan_stol` function in `doped` gives the same functionalities and orders of magnitude faster.

ThomasWarford:
@kavanase thanks for taking a look, I've made the changes you suggested to the docs.

I'll profile this later and see how long the structure matching is taking (I suspect it's quite a small amount of time compared to the relaxations, but maybe not on GPU!). @joehart2001 @ElliottKasoar how would you suggest adding [StructureMatcher_scan_stol](https://doped.readthedocs.io/en/latest/_modules/doped/utils/efficiency.html#StructureMatcher_scan_stol) from doped, and what's your attitude to dependencies?

It looks like doped.utils.efficiency doesn't depend much on the rest of doped so the code could be copied with proper attribution (MIT license)

kavanase:
Ok cool! The structure matching stuff was just if it's annoyingly slow

ThomasWarford:
It seems about ~1/4 of the time is spent structure matching, when running on GPU

joehart2001:
> @kavanase thanks for taking a look, I've made the changes you suggested to the docs.
> 
> I'll profile this later and see how long the structure matching is taking (I suspect it's quite a small amount of time compared to the relaxations, but maybe not on GPU!). @joehart2001 @ElliottKasoar how would you suggest adding [StructureMatcher_scan_stol](https://doped.readthedocs.io/en/latest/_modules/doped/utils/efficiency.html#StructureMatcher_scan_stol) from doped, and what's your attitude to dependencies?
> 
> It looks like doped.utils.efficiency doesn't depend much on the rest of doped so the code could be copied with proper attribution (MIT license)

In general, the fewer dependencies the better, but if this is more than a small helper and the rest of doped is likely to be useful elsewhere, adding it could be justified. @ElliottKasoar thoughts?

ThomasWarford:
Here's the app:
image

Clicking on a violin plot point shows the MLIP relaxed and DFT relaxed structures.

ThomasWarford:
@joehart2001 perhaps it's not worth it for a 25% speedup. If a future PR uses `StructureMatcher` heavily it might be worth looking at again.

kavanase:
Looks very nice! One thing that was initially surprising to me looking at that table (before remembering the origins) -- the errors are larger for PBE than PBEsol, despite the models mostly all being PBE models (outsite the R2SCAN one). But this is more due to their compositions (PBE -> nitrides, PBEsol -> oxides here) right? 

I see this is shown in the lower violin plot ("Oxides, PBEsol"), but I think it would be best if that could be included in the table headers, just to minimise confusion about this. I know this is more text to add so could make it messy, but if needed I think using "Oxides/Nitrides" in the headers would be better than "PBEsol/PBE" as it's the former that is the bigger difference here, whereas the slightly different functional choices are not so big factors?

ElliottKasoar:
> > @kavanase thanks for taking a look, I've made the changes you suggested to the docs.
> > I'll profile this later and see how long the structure matching is taking (I suspect it's quite a small amount of time compared to the relaxations, but maybe not on GPU!). @joehart2001 @ElliottKasoar how would you suggest adding [StructureMatcher_scan_stol](https://doped.readthedocs.io/en/latest/_modules/doped/utils/efficiency.html#StructureMatcher_scan_stol) from doped, and what's your attitude to dependencies?
> > It looks like doped.utils.efficiency doesn't depend much on the rest of doped so the code could be copied with proper attribution (MIT license)
> 
> In general, the fewer dependencies the better, but if this is more than a small helper and the rest of doped is likely to be useful elsewhere, adding it could be justified. @ElliottKasoar thoughts?

Generally I prefer adding a dependency over copying code, otherwise we miss out on any bug fixes, performance improvements, etc., but if it were say a single function we needed and licensing was fine, copying things over could be considered (I'd still reference it, even if it were not required).

Adding dependencies of course comes with potential conflicts etc. to worry about, but we can also make things optional and declare conflicting dependencies if we need to as well, so I also wouldn't overthink it.

(More generally, we need to formalise and potentially revisit some of our decisions with respect to test-specific dependencies, but that's not a job for this!)

ThomasWarford:
@kavanase thanks for pointing this out, I changed it in the latest commit. Materials project is pretty oxide-heavy, which probably explains it.

ElliottKasoar:
Thanks for this, @ThomasWarford, it's looking great! I think something may have gone slightly wrong with your merges, it looks like you've picked up some stray file changes.

ThomasWarford:
Sorry about that, fixed in the latest commit @ElliottKasoar

ml_peg/calcs/defects/split_vacancy/calc_split_vacancies.py
```diff
@@ -0,0 +1,151 @@
...
+from pymatgen.core import Structure
+import pytest
+from tqdm.auto import tqdm
+
+from ml_peg.calcs.utils.utils import download_github_data
+from ml_peg.models.get_models import load_models
+from ml_peg.models.models import current_models
+
+MODELS = load_models(current_models)
+github_uri = "https://github.com/ThomasWarford/defect_data/raw/refs/heads/main/"
```
joehart2001: uri or url?

joehart2001: also would you prefer for us to upload the data to the s3 bucket (cant rememebr if we've already discussed this) or are you downloading from your link?

ElliottKasoar: We use `github_uri`, so I imagine it's based on that. URLs are URIs, URI is just a broader term

ThomasWarford: @joehart2001 Yes, upload the data to S3 if you'd like. I didn't do it yet in case the data would have to change but I think it won't change now.

docs/source/user_guide/benchmarks/defects.rst
ElliottKasoar, line 1: There is (now) a merged `defect.rst` file, so can you combine these two?

docs/source/user_guide/benchmarks/index.rst
```diff
@@ -12,6 +12,7 @@ Benchmarks
     molecular_crystal
     molecular
     bulk_crystal
+    defects
```
ElliottKasoar, line 10: See above

ml_peg/app/defects/defects.yml
ElliottKasoar, line 1: As above, this exists now, and should be merged

ml_peg/calcs/defects/split_vacancy/calc_split_vacancy.py
ElliottKasoar, line 1: Can this be renamed to `calc_split_vacancy.py` to work with the pattern matching (or change the directory name and name elsewhere)?

Also as above, merging of the defect/defects category

ml_peg/app/defects/split_vacancy/app_split_vacancy.py
```diff
@@ -0,0 +1,206 @@
...
+from dash import Dash, Input, Output, callback
+from dash.html import B, Div, Iframe, P
+
+from ml_peg.app import APP_ROOT
+from ml_peg.app.base_app import BaseApp
+from ml_peg.app.utils.build_callbacks import plot_from_table_column
+from ml_peg.app.utils.load import read_plot
+from ml_peg.app.utils.weas import generate_weas_html
+from ml_peg.models.get_models import get_model_names
+from ml_peg.models.models import current_models
```
ElliottKasoar:
```suggestion
from ml_peg.models import current_models
from ml_peg.models.get_models import get_model_names
```
This has moved, sorry!

ml_peg/calcs/defects/split_vacancy/calc_split_vacancy.py
```diff
@@ -0,0 +1,151 @@
...
+from ase.optimize import LBFGS
+import numpy as np
+from pymatgen.analysis.structure_matcher import StructureMatcher
+from pymatgen.core import Structure
+import pytest
+from tqdm.auto import tqdm
+
+from ml_peg.calcs.utils.utils import download_github_data
+from ml_peg.models.get_models import load_models
+from ml_peg.models.models import current_models
```
ElliottKasoar:
```suggestion
from ml_peg.models import current_models
from ml_peg.models.get_models import load_models
```
This has moved, sorry!

ml_peg/analysis/defects/split_vacancy/analyse_split_vacancy.py
```diff
@@ -0,0 +1,717 @@
...
+import pytest
+from scipy.stats import spearmanr
+from tqdm.auto import tqdm
+
+from ml_peg.analysis.utils.decorators import build_table, plot_parity, plot_violin
+from ml_peg.analysis.utils.utils import load_metrics_config, mae
+from ml_peg.app import APP_ROOT
+from ml_peg.calcs import CALCS_ROOT
+from ml_peg.models.get_models import get_model_names
+from ml_peg.models.models import current_models
```
ElliottKasoar:
```suggestion
from ml_peg.models import current_models
from ml_peg.models.get_models import get_model_names
```
This has moved, sorry!

ml_peg/analysis/defects/split_vacancy/analyse_split_vacancy.py
```diff
@@ -0,0 +1,717 @@
...
+print(f"Copying xyz files from {CALC_PATH} to {OUT_PATH} for flask app.")
+shutil.copytree(CALC_PATH, OUT_PATH, dirs_exist_ok=True)
+
+METRICS_CONFIG_PATH = Path(__file__).with_name("metrics.yml")
+DEFAULT_THRESHOLDS, DEFAULT_TOOLTIPS, DEFAULT_WEIGHTS = load_metrics_config(
+    METRICS_CONFIG_PATH
+)
+
+
+def get_hoverdata(functional_path: Path) -> tuple[list, list, list]:
```
ElliottKasoar, line 76: Can you take a look at our new filtering guidelines: https://ddmms.github.io/ml-peg/developer_guide/filter.html, and try to apply that here?

Generally, it should be possible to do this sort of thing through the function we provide, and it makes sure we save the elements in each structure, which are needed to do filtering.

ml_peg/calcs/defects/split_vacancy/calc_split_vacancies.py
ElliottKasoar: This is now uploaded as `inputs/defects/split_vacancy/split_vacancy.zip` (and unpacks to `split_vacancy`), so you can update this @ThomasWarford

ml_peg/calcs/defects/split_vacancy/calc_split_vacancies.py
```diff
@@ -0,0 +1,151 @@
...
+
+                    for initial_atoms in atoms_list:
+                        atoms = deepcopy(initial_atoms)
+                        atoms.calc = deepcopy(calc)
+                        atoms.info["initial_energy"] = atoms.get_potential_energy()
+
+                        converged = opt = LBFGS(atoms, logfile=None)
+                        opt.run(fmax=fmax, steps=steps)
+
+                        atoms.info["relaxed_energy"] = atoms.get_potential_energy()
```
ElliottKasoar: Can you try to catch errors that might be raised by these. It doesn't have to be per calculation, we could do a try...except around the whole thing, but for example, if a model doesn't support any of the elements in one of the structures, this will error and prevent other structures being looked at

ml_peg/calcs/defects/split_vacancy/calc_split_vacancies.py
```diff
@@ -0,0 +1,151 @@
...
+    Run calculations required for split vacancy formation energies.
+
+    Parameters
+    ----------
+    mlip
+        Name of model use and model to get calculator.
+    """
+    model_name, model = mlip
+    model.default_dtype = "float64"
+    calc = model.get_calculator()
```
ElliottKasoar:
```suggestion
    calc = model.get_calculator(precision="high")
```
This is practically the same, but more future-proof

ml_peg/calcs/defects/split_vacancy/calc_split_vacancy.py
```diff
@@ -0,0 +1,151 @@
...
+        for material_dir in tqdm(list((DATA_PATH / functional).iterdir())):
+            cation_dirs = [
+                p for p in material_dir.iterdir() if p.is_dir()
+            ]  # skip pristine supercell.xyz files (not used)
+            for cation_dir in tqdm(cation_dirs, leave=False):
+                nv_xyz_path = cation_dir / "normal_vacancy.xyz"
+                sv_xyz_path = cation_dir / "split_vacancy.xyz"
+
+                if not (nv_xyz_path.exists() and sv_xyz_path.exists()):
+                    continue
```
ElliottKasoar, line 91: Does/should this ever happen?

ml_peg/calcs/defects/split_vacancy/calc_split_vacancies.py
```diff
@@ -0,0 +1,151 @@
...
+                sv_xyz_path = cation_dir / "split_vacancy.xyz"
+
+                if not (nv_xyz_path.exists() and sv_xyz_path.exists()):
+                    continue
+
+                atoms_paths = [nv_xyz_path, sv_xyz_path]
+
+                for atoms_path in atoms_paths:
+                    relaxed_atoms = []
+                    atoms_list = read(atoms_path, ":")
```
ElliottKasoar: Can you add (default) spin (multiplicity) and charge to these structures? Some models (e.g. orb-v3-consv-omol) require these to be set

ml_peg/calcs/defects/split_vacancy/calc_split_vacancy.py
```diff
@@ -0,0 +1,151 @@
...
+        Structure.from_ase_atoms(atoms_1), Structure.from_ase_atoms(atoms_2)
+    )
+
+    if result is None:
+        return (np.nan, np.nan)
+    return result
+
+
+@pytest.mark.parametrize("mlip", MODELS.items())
+def test_relax_and_calculate_energy(mlip: tuple[str, Any]):
```
ElliottKasoar, line 61: I think we probably want to mark this as `slow`?

ml_peg/analysis/defects/analyse_split_vacancy/analyse_split_vacancy.py
```diff
@@ -0,0 +1,717 @@
...
+
+# @pytest.fixture
+# @plot_violin(
+#     title="RMSD Distribution (Oxides, PBEsol)",
+#     y_label="RMSD / Å",
+#     filename=OUT_PATH / "figure_rmsd_pbesol.json",
+# )
+# def rmsd_pbesol_dist(build_results_pbesol) -> dict[str, list]:
+#     _, _, result_rmsd, _, _ = build_results_pbesol
+#     return result_rmsd
```
ElliottKasoar: Can we delete this?

ml_peg/analysis/defects/analyse_split_vacancy/analyse_split_vacancy.py
```diff
@@ -0,0 +1,717 @@
...
+
+# @pytest.fixture
+# @plot_violin(
+#     title="RMSD Distribution (Nitrides, PBE(+U))",
+#     y_label="RMSD / Å",
+#     filename=OUT_PATH / "figure_rmsd_pbe.json",
+# )
+# def rmsd_pbe_dist(build_results_pbe) -> dict[str, list]:
+#     _, _, result_rmsd, _, _ = build_results_pbe
+#     return result_rmsd
```
ElliottKasoar: Can we delete this?

ml_peg/analysis/defects/split_vacancy/analyse_split_vacancy.py
ElliottKasoar, line 1: I'm in the process of running this to test, but can you check that this works with models that have no/missing data (e.g. NaNs, or models that weren't run)?

Also note with the new mock calculator (once you rebase) that you can potentially use that as the source of reference data etc. which can be safer than relying on the first model you happen to analyse.

ElliottKasoar (commented):
Hi @ThomasWarford, I'm running this for a couple of models to give more in depth feedback for the analysis and app, but it's looking great!

Sorry about the delayed review.

With regard to merging with the existing defect category, I'm also very happy to move those into your defects directory. `defects` is arguably a nicer name than `defect` anyway. You can use `git mv` in the same way as you'd use `mv` to move files and track them, if that helps!

ml_peg/analysis/defects/split_vacancy/analyse_split_vacancy.py
ElliottKasoar, line 1: The directory name for this should be the benchmark name, so `ml_peg/analysis/defects/split_vacancy/analyse_split_vacancy.py`, otherwise it won't be automatically discovered

ml_peg/app/defects/split_vacancy/app_split_vacancy.py
```diff
@@ -0,0 +1,206 @@
...
+# Get all models
+MODELS = get_model_names(current_models)
+BENCHMARK_NAME = "Split vacancy"
+DOCS_URL = (
+    "https://ddmms.github.io/ml-peg/user_guide/benchmarks/defects.html#split-vacancy"
+)
+DATA_PATH = APP_ROOT / "data" / "defects" / "split_vacancy"
+
+# for dash, assets/ is equivalent to APP_ROOT/data/
+STRUCTS_URL = "assets/defects/split_vacancy"
```
ElliottKasoar:
```suggestion
STRUCTS_URL = "/assets/defects/split_vacancy"
```
This is needed for the path to work with more recent changes

ml_peg/app/defects/split_vacancy/app_split_vacancy.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+def struct_pair_from_violin(
```
ElliottKasoar, line 266: It might be nice to be able to visualise from the scatters too, if the data is already there, and we have the callbacks already?

joehart2001 (changes requested):

ml_peg/calcs/defects/split_vacancy/calc_split_vacancy.py
ThomasWarford: It happens quite a lot. In theory, we could still calculate some metrics (max_dist between relaxed structures), if only one exists.

If desired, I could remove compound-cation pairs without files and save a few MB on AWS.

ml_peg/analysis/defects/split_vacancy/analyse_split_vacancy.py
ThomasWarford: I merged instead of rebasing, hope that isn't a problem

ThomasWarford: Otherwise resolved now.

ThomasWarford:
I've renamed defect to defects and everything appears to be working

image

ml_peg/app/defects/split_vacancy/app_split_vacancy.py
ThomasWarford: I'm not too sure what you mean.

Do you mean visualise the structures when you click on a scatter point?

For the violin plot, we start with a DFT relaxed structure, relax it with the MLIP and get the RMSD between the two structures. Hence we can link each point to a DFT structure and an MLIP structure neatly.

A formation energy scatter point has four corresponding structures. 

For DFT or the MLIP it is the difference between the lowest energy split vacancy structure and the lowest energy regular vacancy structure. I could show the four but it would require making a new layout and some though to make things clear.

docs/source/user_guide/benchmarks/defects.rst
```diff
@@ -222,3 +221,88 @@ Input structures:
...
+
+3. Match Rate
+
+Fraction of MLIP-relaxed structures that converge to the same geometry as the
+DFT-relaxed reference, determined using the pymatgen
+`StructureMatcher <https://pymatgen.org/pymatgen.analysis.html#module-pymatgen.analysis.structure_matcher>`_.
+The match criterion is a normalised maximum atomic displacement below 0.3 (see
+metric 4).
+
+4. Max Dist
```
ElliottKasoar, line 276: It would be useful to clarify that these are now split into the two groups with different levels of theory

ElliottKasoar (commented):
Thanks again for this, @ThomasWarford!

I just pushed a rebase since we introduced a bunch of conflicts, and a couple of tiny changes for the download (including the other two defect -> defects) 

This may just be my unfamiliarity, but I'm a little confused by the violin plot. Is it supposed to look like this?

image

Other than the minor docs comment I left, the only other thing I wanted to suggest was potentially visualising from the scatter plots for the other three columns? Assuming you think that makes sense, we can do that later.

ml_peg/app/defects/split_vacancy/app_split_vacancy.py
ElliottKasoar: Sorry, I missed this.

image

So from each of these points, it might be nice to visualise the relevant structure(s). You can have a single point open to multiple structures if they're in the same file (e.g. look at the live OC157 in surfaces).

We can also definitely come back to this.

Or maybe it just doesn't make sense in this case?

joehart2001:
hey @ThomasWarford, ive ran the full benchmark and overall looking great.

- could you add a slightly more specific timing to the docs e.g. "medium, roughly 10 mins - 1 hour on gpu"
- for me the structure visualisation only works for the violin plot, are you planning on adding this for the scatters too?

other than this i dont have any further comments and happy to merge after

joehart2001 (approved):
Awesome, this seems to be working nicely! thanks for your work on this @ThomasWarford @kavanase!

PR #394, opened 2026-02-26
Setup solid-state electrolytes benchmark

ThiagoReschutzegger:
## Pre-review checklist for PR author

- [x] I've confirmed the [contribution guidelines](https://github.com/ddmms/ml-peg/blob/main/contributing.md).

## Summary

This pull request introduces initial support for benchmarking solid-state electrolytes (SSEs) within the project. It adds documentation, configuration, metrics, calculation scripts, and an app interface for the new benchmark, called SSE-MD.

Implemented the main calculation script for the SSE-MD RDF benchmark, including test generation, MD simulation setup, and trajectory processing for multiple ML models. Added a metrics definition file for SSE-MD, specifying the score metric and its properties.

## Linked issue

Resolves #393 

## Progress

- [x] Calculations
- [x] Analysis
- [x] Application
- [x] Documentation

## Testing

I tested the benchmark on MACE-MP-0 and MACE-MP-0B3.

## New decorators/callbacks

No new callbacks are required.

PR #360, opened 2026-02-11
Aqueous Iron Chloride Oxidation States

PKourtis:
## Pre-review checklist for PR author

PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines](https://github.com/ddmms/ml-peg/blob/main/contributing.md).

## Summary

Benchmarks how well models capture the different oxidation states of Iron. 

Main tests compares the O-Fe RDFs of Fe+2 and Fe+3 by running 20ps NVT MD for Fe 2Cl and Fe 3Cl ions in water. 

## Linked issue

Resolves #264 

## Progress

- [x] Calculations
- [x] Analysis
- [x] Application
- [x] Documentation

## Testing

Tested on mp0-b3, currently running omol-0 MD.

## New decorators/callbacks

Added new option to the plot_scatter decorator to remove markers from the plots.

PR #445, opened 2026-04-03
Add WiSE electrolyte benchmark (density, X-ray S(q), Li-O RDF)

### LINKED ISSUE #304: Water-in-salt electrolytes
### Summary

We would like to contribute a new ML-PEG benchmark based on water-in-salt LiTFSI–H₂O, focused on the archetypal 21 molal system. This benchmark will be built around DFT/AIMD reference data generated with CP2K (paper in preparation).
It will test MLIPs on bulk properties such as the experimental structure factor S(q), the AIMD RDF and coordination number.

### Interactive features

Visualisation and comparison of RDFs, structure factor S(q) (vs experiment), and Li coordination numbers across MLIPs

### Category

Molecular liquids/Electrolytes

### Data availability

LiTFSI–H₂O at 21 molal concentration.
r²SCAN CP2K NVT AIMDs at 300 K
Labels available: Energies, Forces, Stress
Chemical elements present: Li, H, O, C, N, F, S.

### Computational cost

On a single GPU: ~6–24 hours per model on one NVIDIA H100, depending on box size, simulation length, and the foundation model used.

### Additional context

_No response_

LucaBrugnoli:
## Summary

  New benchmark for 21 molal LiTFSI/H₂O **water-in-salt electrolyte** (WiSE), evaluating MLIP foundation models on three
  experimental observables:

  1. **Density** — NPT equilibrium density vs [Gilbert et al., *JCED* 62, 2056
  (2017)](https://doi.org/10.1021/acs.jced.7b00135)
  2. **X-ray S(q)** — Structure factor R-factor vs SAXS data, [Zhang et al., *JPCB* 125, 4501
  (2021)](https://doi.org/10.1021/acs.jpcb.1c02189). Computed via [dynasor](https://dynasor.materialsmodeling.org/).
  3. **Li-O RDF** — Coordination numbers (Li-O_water, Li-O_TFSI) vs [Watanabe et al., *JPCB* 125, 7477
  (2021)](https://doi.org/10.1021/acs.jpcb.1c04693)

  Further details on the simulation protocol and MLIP assessment for this system: [L. Brugnoli, arXiv:2603.22099
  (2026)](https://arxiv.org/abs/2603.22099).

  ## Linked issue

  Resolves #304

  ## Progress

  - [x] Calculations
  - [x] Analysis
  - [x] Application
  - [x] Documentation

  **Note:** Trajectory data (~500 MB, extxyz) needs to be uploaded to the ml-peg S3 bucket. Data files are available on
  request.

  ## Testing

  Tested on 6 models: `matpes-r2scan`, `mace-mpa-0-medium`, `mace-omat-0-medium`, `mace-mp-0b3`, `mace-mh-1-omat`,
  `mace-mh-1-omol`.

  **Requirement:** ASE < 3.28 (3.28.0 has a bug in `ase.io.extxyz.ixyzchunks`). Tests run with `--noconftest`.

  ## New decorators/callbacks

  No new callbacks required. The RDF app uses existing `plot_from_table_column` from ml-peg utils.


ml_peg/analysis/wise_electrolytes/density/analyse_density.py
```diff
@@ -0,0 +1,298 @@
...
+    results = {}
+    for model in MODELS:
+        json_path = CALC_PATH / model / "density.json"
+        if json_path.exists():
+            with open(json_path) as f:
+                results[model] = json.load(f)
+    return results
+
+
+def normalize_metric(value: float, good: float, bad: float) -> float:
```
joehart2001: this will be taken care of when you build the table using the decorator. you could define your own function if you dont want to use our defualt, see the [docs](https://ddmms.github.io/ml-peg/developer_guide/scoring_and_normalisation.html)

ml_peg/analysis/wise_electrolytes/density/analyse_density.py
```diff
@@ -0,0 +1,298 @@
...
+from ml_peg.analysis.utils.utils import load_metrics_config
+
+# --- Paths -------------------------------------------------------------------
+
+CALCS_ROOT = Path(__file__).resolve().parents[3] / "calcs"
+CALC_PATH = CALCS_ROOT / "wise_electrolytes" / "density" / "outputs"
+APP_ROOT = Path(__file__).resolve().parents[3] / "app"
+OUT_PATH = APP_ROOT / "data" / "wise_electrolytes" / "density"
+
+MODELS = [
```
joehart2001: usually we import this to get all the models. this is better for in the future when we have more models than now.
```
from ml_peg.models.models import current_models
```

ml_peg/analysis/wise_electrolytes/density/analyse_density.py
```diff
@@ -0,0 +1,298 @@
...
+        xaxis_title="Time / ps",
+        yaxis_title="Density / g cm⁻³",
+    )
+    return fig
+
+
+# --- Metrics table -----------------------------------------------------------
+
+
+def build_metrics_table(data: dict[str, dict]) -> dict:
```
joehart2001: we have decorators to build this automatically, see the [tutorial](https://github.com/ddmms/ml-peg/blob/main/docs/source/tutorials/python/adding_benchmark.ipynb)

joehart2001:
Hi @LucaBrugnoli thanks for the PR! From what i understand, you've provided us with NVT trajectories for each model and then your calc script takes these trajectories to calculate e.g. the rdf etc. Thank you for providing these trajectories! However, to make this test not rely on you computing these for each new model that is added in the future (as im sure you would prefer us running them for you!), I think we need to do some reshuffling. 

The ideal workflow would be:
**calc**: 
- this is where your reproducable NVT script would go, and you can look at examples such as PR #360  (direct link to [line](https://github.com/ddmms/ml-peg/pull/360/changes#diff-78cb42931c4778794846bc6f1a785d7fbd541e17f0a44ecf01a1c32b3d4ffd7dR60)) for an easy way to do this with e.g. Janus (used for many of our MD benchmarks). 
- To keep things modular and separated, we like to keep anything that could be considered analysis in the anlysis scripts only. The PR linked computes the rdf as part of the calc, so you could do this too, or separate it out into the analysis script. This format will mean you will only have one directory for your benchmark instead of three.

**analysis**
- ive left some comments about using our helpers and decorators to help be consistent and also keep up with future changes to table builidng etc.

**app**
- overall looks good, just need to add the docs link later. do you think structure visualisation would be useful in this benchmark? if so we could help add that

ml_peg/calcs/wise_electrolytes/md_reference/calc_md_reference.py
ElliottKasoar, line 1: Thanks for adding this reference!

Apologies if you've stated this somewhere and I missed it, but how confident are you that this reproduces the LAMMPS dynamics?

I would suggest we don't use `pytest.skip`, and instead mark them as very slow (`@pytest.mark.very_slow`). This already requires users to explicitly request it, with the understanding that they will likely take a minimum of several hours on GPU, so we would run a single model at a time.

Since we can't guarantee all models can be run through LAMMPS/with additional acceleration, we would expect to run this test for several, if not all, of the models using the ASE calculator.

While of course it will be considerably slower, it means we can guarantee we can run every model, and that the settings are identical for each model.

We already have several tests that take over of day of GPU time, including #388, so this isn't a problem - it's a high, but one-off cost.

joehart2001:
Thanks for the code updates! I think this test could also be well suited in the molecular dynamics category instead of its own one as its quite specific, opinions @ElliottKasoar?

LucaBrugnoli:
Thanks for both messages, I'll let you and @ElliottKasoar decide on the location; happy with whichever directory you prefer.
On the LAMMPS/ASE reproducibility question: I'm currently running the protocol on a cluster with GPUs MI250X, with mace-mp-0b3 and mace-matpes-r2scan to validate it empirically.  
I've tried to match the integrator formulations exactly,  NVT_NH for LAMMPS fix nvt (Nosé–Hoover chain, length 3) and NPT_MTK for fix npt iso (Martyna–Tobias–Klein chain, length 3), with the same TDAMP=50 fs / PDAMP=500 fs and dt=0.5 fs. 
Even so, I don't expect bit-for-bit trajectory agreement: the LAMMPS production used the SymmetriX/Kokkos MACE kernels, while ASE/janus uses the reference PyTorch path; combined with different RNG-seeded initial velocities, chaotic divergence is unavoidable. What should agree, within sampling uncertainty, are the statistical observables, i.e the equilibrium density, RDFs, coordination numbers, and S(q). 
I'll update the PR as soon as the runs are done with the LAMMPS/ASE comparison and the wallclock cost per ns of MD on a single MI250X GCD, so we have a realistic estimate for running the marker across the registry.

docs/source/user_guide/benchmarks/index.rst
ElliottKasoar, line 1: These are subject to reshuffling later anyway, but we have another PR (#394) proposing to add a general electrolytes category. Do you think we could use that, which is perhaps a bit broader?

ml_peg/models/models.yml
ElliottKasoar, line 1: Can you revert these changes? We will test these models, but we make the changes locally (or would do this change as a separate PR, now that MACE has released these properly)

ml_peg/calcs/wise_electrolytes/md_reference/calc_md_reference.py
```diff
@@ -0,0 +1,246 @@
...
+        If True, append a further 50 ps of NVT after the NPT block
+        (analogue of ``in.nvt_continue.lmp``). Default False.
+    """
+    # Lazy imports: janus_core pulls in torch, which may be absent in
+    # environments that only run the analysis stack.
+    from janus_core.calculations.geom_opt import GeomOpt
+    from janus_core.calculations.md import NPT_MTK, NVT_NH
+
+    model.default_dtype = "float64"
+    calc = model.get_calculator()
```
ElliottKasoar:
```suggestion
    calc = model.get_calculator(precision="high")
```
This is effectively the same, but more future-proof.

ml_peg/calcs/wise_electrolytes/md_reference/calc_md_reference.py
ElliottKasoar, line 1: Can this be combined with `calc_litfsi_h2o_21m.py`? Generally the idea is to have `ml_peg/calcs/[category]/[benchmark]/calc_[benchmark].py,`, but this isn't really its own benchmark?

We have similar examples e..g [oxidation_states](https://github.com/ddmms/ml-peg/blob/main/ml_peg/calcs/physicality/oxidation_states/calc_oxidation_states.py) where there's a slow part that we may well run separately, and a faster part.

As long as the slow part is defined first, I think it's fine since they can be run either together or independently thanks to the very slow marker.

ml_peg/calcs/wise_electrolytes/md_reference/calc_md_reference.py
```diff
@@ -0,0 +1,246 @@
...
+        "--continuation",
+        action="store_true",
+        help="append the optional 50 ps NVT continuation after NPT.",
+    )
+    args = parser.parse_args()
+
+    if args.model not in MODELS:
+        parser.error(f"unknown model '{args.model}'. Registered: {sorted(MODELS)}")
+
+    run_reference_md(args.model, MODELS[args.model], run_continuation=args.continuation)
```
ElliottKasoar: A couple of questions/suggestions:

- What's the context of `continuation` - when would we/wouldn't we want to use it? Generally, I think it's good to define a single protocol for all the models
- It's preferable to integrate with the `ml_peg calc` CLI, which also allows us to add custom inputs if you wanted to add an new conftest.py (see the water and liquid densities in the MD category for examples). That already has built in things for choosing the model(s) etc.

LucaBrugnoli: Thanks you both for the detailed review, and apologies for the delayed response. Here's a summary of what I've changed, plus two open questions I can't settle on my own.
  
1. models.yml
Reverted to the upstream state, as requested. This undoes the model additions that went in with e1984696; ok for the MH-1 heads to be enabled on your side or in a separate PR now that MACE has released them properly. 
My results for those two models don't depend on the registry flag, so nothing is lost by leaving them commented.
  
2. Reference MD protocol
Your question about continuation prompted me to re-check the script against the protocol that actually produced the published trajectories, and I found two discrepancies. The script was running minimization → 50 ps NVT → 200 ps NPT on p64_w170, with an optional NVT continuation, and it never used p16_w42. This differs from the production workflow in two important ways: it included an NPT stage that was not used in production, and, because the density data were obtained from the smaller p16_w42 cell, it could not have regenerated density.json. 
  
This is what production actually did, and what the script now does:
- p64_w170 (S(q) and RDF): Min → NVT 50 ps equilibration → NVT 50 ps production, held at the experimental volume (L =27.49 Å) throughout. No NPT: the box is fixed at the experimental density, so that S(q) and the coordination numbers are compared at the reference density rather than at each model's own.
- p16_w42 (density): Min → NVT 50 ps → NPT 200 ps, with the density averaged over the last 150 ps. NPT is used only here, on the smaller cell, because it's cheaper to converge.
  
One consequence worth flagging: the mace-mp-0b3 density I sent in April came from a 50 ps NPT run, so it no longer matched  the protocol above. I've rerun it at 200 ps like the other three. The density was practically unchanged  (from 1.5632 to 1.5623 g/cm³), but the set is now homogeneous. The updated zip is attached below.
  
3. Continuation
This option was a consequence of the protocol mismatch described above. The second 50 ps NVT block is not an optional continuation after NPT, but the production stage that follows equilibration directly, with no intervening NPT stage. Once the protocol was corrected, the flag no longer served a purpose, so I removed it. All models now follow the same single protocol.
When a test is rerun, it instead detects and skips any stages that have already reached their final step, allowing interrupted runs to resume without repeating the entire workflow.
  
4. Scripts combined
I merged the two scripts into a single calc_litfsi_h2o_21m.py, following the structure used in [oxidation_states](https://github.com/ddmms/ml-peg/blob/main/ml_peg/calcs/physicality/oxidation_states/calc_oxidation_states.py). 
The very_slow MD test is defined first and generates the trajectories. The extraction tests then read those trajectories and, when the required data are unavailable, skip with a reference to the MD test that produces them. The MD and analysis stages can therefore be run either together or independently. The separate md_reference/ directory has been removed, leaving the benchmark contained in a single directory as intended.
  
5. Precision
Applied as suggested. I also removed the model.default_dtype = "float64" assignment immediately above it.
   
About the earlier review comments
@joehart2001, the three comments from April are now addressed in the consolidated analyse_litfsi_h2o_21m.py. It uses the build_table decorator for table generation and scoring, and loads the registry through load_models(current_models) rather than using a hard-coded model list.
analyse_density.py has been removed, so I believe those three threads can now be resolved.

Category change
I agree that the electrolytes category introduced in PR (https://github.com/ddmms/ml-peg/pull/394) is the most appropriate location for this benchmark. Since that PR has not yet been completed, would you prefer me to restructure the benchmark against it now, or leave the current location unchanged and move it once PR (https://github.com/ddmms/ml-peg/pull/394) is merged?

Additional data upload

The April upload contained only the generated outputs, not the initial structures. I have attached the missing inputs here, together with the updated mace-mp-0b3 results:

litfsi_h2o_21m_initial_structures.zip: contains p64_w170_initial.xyz (1,534 atoms) and p16_w42_initial.xyz (382 atoms), both at the experimental density.
mace-mp-0b3.zip: replaces the archive uploaded in April. It contains the same NVT trajectory, together with the updated density.json from the 200 ps rerun.

[mace-mp-0b3.zip](https://github.com/user-attachments/files/30106261/mace-mp-0b3.zip)
[litfsi_h2o_21m_initial_structures.zip](https://github.com/user-attachments/files/30106353/litfsi_h2o_21m_initial_structures.zip)

LucaBrugnoli:
Reposting here because my earlier reply ended up in a review thread on `md_reference/calc_md_reference.py`.                Since that file has now been removed, GitHub marks the thread as outdated and hides the comment by default.

Thank you both for the detailed review, and apologies for the delayed response. Here's a summary of what I've changed, plus two open questions I can't settle on my own.

1. models.yml
   Reverted to the upstream state, as requested. This undoes the model additions that went in with e1984696; I am happy for the MH-1 heads to be enabled on your side or in a separate PR now that MACE has released them properly.
   My results for those two models don't depend on the registry flag, so nothing is lost by leaving them commented.

2. Reference MD protocol
   Your question about continuation prompted me to re-check the script against the protocol that actually produced the published trajectories, and I found two discrepancies. The script was running minimization → 50 ps NVT → 200 ps NPT on p64_w170, with an optional NVT continuation, and it never used p16_w42. This differed from the production workflow in two important ways: it included an NPT stage that was not used in production, and, because the density data were obtained from the smaller p16_w42 cell, it could not have regenerated density.json.

This is what production actually did, and what the script now does:

* p64_w170 (S(q) and RDF): minimization → NVT 50 ps equilibration → NVT 50 ps production, held at the experimental volume (L = 27.49 Å) throughout. No NPT: the box is fixed at the experimental density, so that S(q) and the coordination numbers are compared at the reference density rather than at each model's own.
* p16_w42 (density): minimization → NVT 50 ps → NPT 200 ps, with the density averaged over the last 150 ps. NPT is used only here, on the smaller cell, because it is less computationally expensive to run for the required duration.

One consequence worth flagging: the mace-mp-0b3 density I sent in April came from a 50 ps NPT run, so it did not match the protocol above. I've rerun it for 200 ps, as for the other three models. The density was practically unchanged (from 1.5632 to 1.5623 g/cm³), but the dataset is now consistent. The updated zip is attached below.

3. Continuation
   This option was a consequence of the protocol mismatch described above. The second 50 ps NVT block is not an optional continuation after NPT, but the production stage that follows equilibration directly, with no intervening NPT stage. Once the protocol was corrected, the flag no longer served a purpose, so I removed it. All models now follow the same protocol for each observable.
   When a test is rerun, it instead detects and skips any stages that have already reached their final step, allowing interrupted runs to resume without repeating the entire workflow.

4. Scripts combined
   I merged the two scripts into a single calc_litfsi_h2o_21m.py, following the structure used in [oxidation_states](https://github.com/ddmms/ml-peg/blob/main/ml_peg/calcs/physicality/oxidation_states/calc_oxidation_states.py). 
   The very_slow MD test is defined first and generates the trajectories. The extraction tests then read those trajectories and, when the required data are unavailable, skip with a reference to the MD test that produces them. The MD and analysis stages can therefore be run either together or independently. The separate md_reference/ directory has been removed, leaving the benchmark contained in a single directory as intended.

5. Precision
   Applied as suggested. I also removed the model.default_dtype = "float64" assignment immediately above it. That assignment took precedence over the precision argument passed to get_calculator, so changing the argument alone would not have changed the dtype.

About the earlier review comments
@joehart2001, the three comments from April are now addressed in the consolidated analyse_litfsi_h2o_21m.py. It uses the build_table decorator for table generation and scoring, and loads the registry through load_models(current_models) rather than using a hard-coded model list.
analyse_density.py has been removed, so I believe those three threads can now be resolved.

Category change
I agree that the electrolytes category introduced in (https://github.com/ddmms/ml-peg/pull/394) is the most appropriate location for this benchmark, and I have now moved it there. I also renamed the documentation page to electrolytes.rst. I kept its entry in index.rst in the same position used in (https://github.com/ddmms/ml-peg/pull/394).

Additional data upload

The April upload contained only the generated outputs, not the initial structures, so the MD test cannot currently be run without the local input files. I have attached the missing inputs here, together with the updated mace-mp-0b3 results:

litfsi_h2o_21m_initial_structures.zip: contains p64_w170_initial.xyz (1,534 atoms) and p16_w42_initial.xyz (382 atoms), both at the experimental density.
mace-mp-0b3.zip: replaces the archive uploaded in April. It contains the same NVT trajectory, together with the updated density.json from the 200 ps rerun.


[litfsi_h2o_21m_initial_structures.zip](https://github.com/user-attachments/files/30122084/litfsi_h2o_21m_initial_structures.zip)
[mace-mp-0b3.zip](https://github.com/user-attachments/files/30155036/mace-mp-0b3.zip)



The calculation still reads its inputs from a local directory rather than through download_s3_data, which is the last remaining change I am aware of. What would be the best way to handle this from here?

The pre-commit failure shown for 6a972ba was fixed in 4cd1ff2, including the ruff-format changes and two NumPy-style docstring fixes. The CI run for the latest commit is currently waiting for approval.

ml_peg/calcs/electrolytes/litfsi_h2o_21m/calc_litfsi_h2o_21m.py
```diff
@@ -0,0 +1,959 @@
...
+
+warnings.filterwarnings("ignore")
+
+# --- Configuration -----------------------------------------------------------
+
+MODELS = load_models(current_models)
+
+DATA_ROOT = Path(
+    os.environ.get(
+        "ML_PEG_WISE_LITFSI_H2O_21M_DATA_ROOT",
```
joehart2001, line 87: i dont think we want to use an evironment variable here. we should always refer to the data dir

ml_peg/calcs/electrolytes/litfsi_h2o_21m/calc_litfsi_h2o_21m.py
```diff
@@ -0,0 +1,959 @@
...
+        Registry name of the MLIP model.
+    model : Any
+        Model object from :func:`load_models`.
+    """
+    # Imported lazily: janus-core pulls in torch, which is not needed by the
+    # extraction tests.
+    from janus_core.calculations.geom_opt import GeomOpt
+    from janus_core.calculations.md import NPT_MTK, NVT_NH
+
+    calc = model.get_calculator(precision="high")
```
joehart2001, line 358: do we want float64 here?

ml_peg/calcs/electrolytes/litfsi_h2o_21m/calc_litfsi_h2o_21m.py
```diff
@@ -0,0 +1,959 @@
...
+
+    data_dir = DATA_ROOT / model_name
+    work_dir = OUT_PATH / model_name / "reference_md"
```
joehart2001, line 361:
```suggestion
    input_dir = (
        download_s3_data(
            filename="litfsi_h2o_21m.zip",
            key="inputs/electrolytes/litfsi_h2o_21m/litfsi_h2o_21m.zip",
        )
        / "litfsi_h2o_21m"
    )
    data_dir = OUT_PATH / model_name
    work_dir = data_dir / "md"
```

ml_peg/calcs/electrolytes/litfsi_h2o_21m/calc_litfsi_h2o_21m.py
joehart2001, line 1: it could be nice to have a utils file in this directory as the calc file is very busy. so the calc file could contain simple functions which are easy to follow and a lot of the current functiosn could be in a utils

joehart2001:
hey @LucaBrugnoli ive uplaoded `p16_w42_initial.xyz`, `p64_w170_initial.xyz` and `saxs_maginn.txt` to the s3 bucket. is there any additional data we need? you'll also need to update some of the paths so it finds the s3 downloads. thanks!

LucaBrugnoli:
Hi @joehart2001, thanks for uploading the data!

In the meantime I revised the methodology of this benchmark, and I have an updated version of the PR ready, but before pushing it I'd like to check a few points with you (see the end of this message). In short, compared with the version currently in the PR:

- **Starting structures.** Every model now starts from the same two cells (p64_w170 and p16_w42), pre-equilibrated once with a classical force field at the experimental density. The current version starts each model from the raw packmol cells and minimises them. That start leaves the composition of the Li+ solvation shell biased, and 50 ps of MD do not remove the bias. In some models it also broke molecules during the minimisation.
- **Dispersion.** Models not trained on dispersion-corrected data now get D3(BJ) at runtime via `add_d3_calculator`, as in the other benchmarks, and show up with the `-D3` suffix.
- **Metrics.**
  - The S(q) smoothing window now goes from 5 to 27 points (0.54 Å⁻¹, close to the 2π/L resolution of the cell).
  - The good/bad thresholds are now less strict: density 2%/15%, S(q) R-factor 0.02/0.21.
  - The total Li-O CN is still reported but no longer scored.
  - Weights: density 1, S(q) 0.5, CN water 0.25, CN TFSI 0.25.
- **Code.**
  - The MD and observable helpers move to `litfsi_utils.py`.
  - The environment variables are gone and the calculations run at `precision="low"`.
  - Outputs go to the standard `outputs/` and `app/data` locations.
  - Inputs and the SAXS reference are read with `download_s3_data`, so the updated PR also fixes the paths.

I have also extended the benchmark beyond the models we initially agreed on, to the rest of the models in the project: 23 models (38 runs, counting those with and without D3).

On the S3 data: since the MD now starts from the pre-equilibrated cells, the zip on the bucket needs replacing. The key and `saxs_maginn.txt` stay the same; only the two structure files change. The updated `litfsi_h2o_21m.zip` is attached, with `p64_w170_equilibrated.xyz`, `p16_w42_equilibrated.xyz` and `saxs_maginn.txt`. The `*_initial.xyz` files are no longer used.

A few questions before I push:

1. MACE-MH-1 heads: In my tests I also included the other MACE-MH-1 heads (matpes_r2scan, spice, mp, oc20), which are not in `models.yml`. Can I add them to the benchmark, or should I leave them out? Including them would bring it to 27 models (45 runs).
2. D4 label: The `mace-matpes-r2scan` entry in `models.yml` has `label: D4`, so its corrected results appear as `mace-matpes-r2scan-D4`. However, `add_d3_calculator` applies D3(BJ) (with the r2SCAN parameters), which is also what I used. I kept the `-D4` label for consistency with the other benchmarks. Should it be changed to D3?
3. Additional data: The analysis only reads the small per-model files written by the calc step (`density.json`, `rdf.json/.npz`, `xray_sq.json/.npz`, a few MB for all models). The full trajectories are about 11 GB. I don't have access to them right now, so I'll upload them as soon as I can. Would you like only the per-model outputs, or the trajectories as well? And where should they go?

Once I hear back from you, I'll push the updated version.
[litfsi_h2o_21m.zip](https://github.com/user-attachments/files/32199455/litfsi_h2o_21m.zip)

joehart2001:
> Hi @joehart2001, thanks for uploading the data!
> 
> In the meantime I revised the methodology of this benchmark, and I have an updated version of the PR ready, but before pushing it I'd like to check a few points with you (see the end of this message). In short, compared with the version currently in the PR:
> 
> * **Starting structures.** Every model now starts from the same two cells (p64_w170 and p16_w42), pre-equilibrated once with a classical force field at the experimental density. The current version starts each model from the raw packmol cells and minimises them. That start leaves the composition of the Li+ solvation shell biased, and 50 ps of MD do not remove the bias. In some models it also broke molecules during the minimisation.
> * **Dispersion.** Models not trained on dispersion-corrected data now get D3(BJ) at runtime via `add_d3_calculator`, as in the other benchmarks, and show up with the `-D3` suffix.
> * **Metrics.**
>   
>   * The S(q) smoothing window now goes from 5 to 27 points (0.54 Å⁻¹, close to the 2π/L resolution of the cell).
>   * The good/bad thresholds are now less strict: density 2%/15%, S(q) R-factor 0.02/0.21.
>   * The total Li-O CN is still reported but no longer scored.
>   * Weights: density 1, S(q) 0.5, CN water 0.25, CN TFSI 0.25.
> * **Code.**
>   
>   * The MD and observable helpers move to `litfsi_utils.py`.
>   * The environment variables are gone and the calculations run at `precision="low"`.
>   * Outputs go to the standard `outputs/` and `app/data` locations.
>   * Inputs and the SAXS reference are read with `download_s3_data`, so the updated PR also fixes the paths.
> 
> I have also extended the benchmark beyond the models we initially agreed on, to the rest of the models in the project: 23 models (38 runs, counting those with and without D3).
> 
> On the S3 data: since the MD now starts from the pre-equilibrated cells, the zip on the bucket needs replacing. The key and `saxs_maginn.txt` stay the same; only the two structure files change. The updated `litfsi_h2o_21m.zip` is attached, with `p64_w170_equilibrated.xyz`, `p16_w42_equilibrated.xyz` and `saxs_maginn.txt`. The `*_initial.xyz` files are no longer used.
> 
> A few questions before I push:
> 
> 1. MACE-MH-1 heads: In my tests I also included the other MACE-MH-1 heads (matpes_r2scan, spice, mp, oc20), which are not in `models.yml`. Can I add them to the benchmark, or should I leave them out? Including them would bring it to 27 models (45 runs).
> 2. D4 label: The `mace-matpes-r2scan` entry in `models.yml` has `label: D4`, so its corrected results appear as `mace-matpes-r2scan-D4`. However, `add_d3_calculator` applies D3(BJ) (with the r2SCAN parameters), which is also what I used. I kept the `-D4` label for consistency with the other benchmarks. Should it be changed to D3?
> 3. Additional data: The analysis only reads the small per-model files written by the calc step (`density.json`, `rdf.json/.npz`, `xray_sq.json/.npz`, a few MB for all models). The full trajectories are about 11 GB. I don't have access to them right now, so I'll upload them as soon as I can. Would you like only the per-model outputs, or the trajectories as well? And where should they go?
> 
> Once I hear back from you, I'll push the updated version. [litfsi_h2o_21m.zip](https://github.com/user-attachments/files/32199455/litfsi_h2o_21m.zip)

Hey @LucaBrugnoli thank you for this and the changes sound great! i've uplaoded the new data

in response to your questions:
-  i think more data is always welcome but models beyond our specified ones are not required. if they are interesting results for this benchmark then i would say include them, but its up to you. as you have some experimental reference data then it could make more sense than other benchmarks to do this. for mh-1, omat_pbe is meant to be the universal head and then we have added omol as this is an interesting head too.
- i believe D3 with the r2scan parameters is equivalent to D4 (you may have to check me on this)
- if we just need this data then it could be good to keep it light. it could also be good to save a frame for strucutre vis. maybe lets upload the trajs to a google drive just so we have them shared incase we choose to use them in the future?

ml_peg/calcs/electrolytes/litfsi_h2o_21m/calc_litfsi_h2o_21m.py
ElliottKasoar: This might be a confusing question out of context. Previously, the precision was explicitly set as `"float64"`, and I suggested setting this via `precision="high"` as our more generalised form across models.

I think the question here is if actually it ought to be `precision="low"` (usually float32), which is generally faster and often what we recommend for MD simulations

PR #468, opened 2026-04-14
⚡ Bolt: Cache framework registry loading

alinelena:
💡 **What:** Added `@lru_cache(maxsize=1)` to `load_framework_registry()` and ensured callers use `copy.deepcopy()` when accessing the loaded configuration dict.

🎯 **Why:** Loading static configurations via `yaml.safe_load` on disk is surprisingly expensive and repeatedly parsing `frameworks.yml` (which does not change during the app lifecycle) caused unnecessary overhead, particularly in a Dash environment where components may re-render or pull config data repeatedly.

📊 **Impact:** Reading the framework config is now an instantaneous in-memory lookup rather than a disk read and string-parsing operation.

🔬 **Measurement:** You can verify the improvement by analyzing execution time metrics or tracing the number of file handle opens against `frameworks.yml` during repeated sidebar component interactions in the app. Tests are passing correctly and cache isolation is preserved.

---
*PR created automatically by Jules for task [11349013377784082073](https://jules.google.com/task/11349013377784082073) started by @alinelena*


ElliottKasoar:
Duplicate of #488

PR #485, opened 2026-04-20
Add summary info

### LINKED ISSUE #137: Add CLI list option
### Summary

This is relatively low priority, but it might be nice to have a way to list all tests/categories, e.g. `ml_peg list`.

### Example

_No response_

### Alternatives

_No response_

### Additional context

_No response_

ElliottKasoar:

## Summary

Adds summary information for calculations, analysis, and applications.

Since, for example, we can have a data-only test with no calculation, each of these is collected separately.

Categories are presented as a set of case-insensitive choices, enabling autocomplete when using the CLI. We could do the same for tests, but this would make the help incredibly long.

I have also added a new `info` command, with a set of sub-commands to list the available categories and tests (including tests in a specific category), and the current models too.

We may want to extend this to the available data for download, but that means further work, as currently these are hidden by default.

## Linked issue

Resolves #137

## Testing

Tested commands locally.

### DIFF
--- a/docs/source/developer_guide/cli.rst
+++ b/docs/source/developer_guide/cli.rst
@@ -10,7 +10,7 @@ commands::
     ml_peg calc
     ml_peg analyse
     ml_peg download
-    ml_peg info
+    ml_peg list
 
 
 For example, to run the X23 test with mace-mp-0a and orb-v3-consv-inf-omat, you can run::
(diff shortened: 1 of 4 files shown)


docs/source/developer_guide/cli.rst
```diff
@@ -0,0 +1,37 @@
...
+A description of each subcommand, as well as valid options, can be listed using the
+``--help`` option. For example::
+
+
+    ml_peg calc --help
+
+The ``ml_peg info`` command provides a further set of subcommands::
+
+
+    ml_peg info calc
```
joehart2001: can we add indicaitons on which tests are slow and very slow when using this?

ElliottKasoar: Maybe, but not trivially. Currently the tests/categories are all just `glob` based.

To get the slow markers, we'd need to access the actual pytest tests, which isn't too bad in itself, but then also deal with the fact that some tests have multiple subtests (which may not all be slow/very slow).

Happy to look into it but I'd suggest at the very least turning this into a further issue.

ElliottKasoar:
I may actually change it to `ml_peg list calcs/analysis/app/models`, which maybe sounds a bit more natural (and fits the dir names)

docs/source/developer_guide/cli.rst
joehart2001: Agreed

ElliottKasoar: Done in #489

joehart2001 (approved):

PR #487, opened 2026-04-20
Remove mean position during LNCI16

ElliottKasoar:

## Summary

Removes mean positions of atoms for LNCI16, which is currently necessary to run high precision with mace-polar-1-l without memory issues on most devices, and increases the speed considerably too.

## Testing

Tested on a 32 core EPYC 7502P. Still unable to run locally.

Further testing should be done to confirm other models are unchanged, but this didn't affect `mace-polar-1-m`.


ElliottKasoar:
@joehart2001 I've tested this for a few different models (including polar models and it doesn't seem to change the values

joehart2001 (approved):
great thanks

Issue #397, opened 2026-03-01
Standardise level of theory labels


joehart2001:
### Summary

We need to prevent mismatches in the case the level of theory is e.g. PBE + D3 vs PBE+D3 etc, probably via a page in the docs

PR #493, opened 2026-04-22
Level of Theory docs

### LINKED ISSUE #309: Add level of theory flags to the docs
### Summary

- update with new flags we discussed (a few weeks ago)
- explain beyond DFT logic (e.g. including RPA in the same category as CCSD(T))

### Example

_No response_

### Alternatives

_No response_

### Additional context

_No response_

joehart2001:

## Summary
Adds docs on level of theory flags and standardising the names of each level of theory used

## Linked issue

Resolves #309 and partially #397 

## Testing
built docs

### DIFF
--- a/docs/source/developer_guide/add_benchmarks.rst
+++ b/docs/source/developer_guide/add_benchmarks.rst
@@ -36,7 +36,7 @@ e.g. CCSD(T).
 
 Each metric in ``metrics.yml`` should have a ``level_of_theory`` field identifying its reference
 method. The app compares this string against the ``level_of_theory`` set in ``models.yml`` for each
-MLIP, flagging mismatches in the app using the traffic light system. Important: an exact string match is required for the comparison to work correctly.
+MLIP, flagging mismatches using the traffic light system. An exact string match is required.
 
 .. warning::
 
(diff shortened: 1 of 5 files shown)


ElliottKasoar (approved):
Two thoughts:

I wouldn't say this resolves #397 entirely, as I do think we want this list in the code as something like a `Literal` we add to, and check against. I'm not sure if this was intentional, but when you do "resolves #x and #y", it only sets #x to close unless you also say "resolves #y", so this is actually fine, in my opinion, whether intended or not.

We probably want some "user" facing docs for this too. I'm happy for this to be merged as is, but we could also consider moving/copying some of this. I'm not sure there's an obvious solution without some duplication.

joehart2001:
> Two thoughts:
> 
> I wouldn't say this resolves #397 entirely, as I do think we want this list in the code as something like a `Literal` we add to, and check against. I'm not sure if this was intentional, but when you do "resolves #x and #y", it only sets #x to close unless you also say "resolves #y", so this is actually fine, in my opinion, whether intended or not.
> 
> We probably want some "user" facing docs for this too. I'm happy for this to be merged as is, but we could also consider moving/copying some of this. I'm not sure there's an obvious solution without some duplication.

yeah agreed, i was going to add (partially) resolves 397

PR #495, opened 2026-04-22
Fix table width

### LINKED ISSUE #481: Table column sizes broken
### Summary

Some table columns are incredibly wide for some reason

### Platform

N/A

### Python version

_No response_

joehart2001:

## Summary
Fix the issue of tables stretching to the width of the largest table on each page. 

Each table and its controls are now placed together inside a single `width: fit-content` container instead of using 100% width. The container collapses to the table's column-sum width. The weight/threshold controls sit inside the same container so their width: 100% resolves relative to the table, keeping them aligned as the table narrows. also did a small refactor to add `compact_weight_components` helper which hopefully makes the code a bit clearer

## Linked issue

Resolves #481


ml_peg/app/utils/build_components.py
```diff
@@ -328,6 +328,35 @@ def build_weight_components(
     return Div(layout)
 
 
+def compact_weight_components(weight_components: Div) -> Div:
+    """
+    Remove the leading spacer from ``build_weight_components`` output.
```
ElliottKasoar: Does "spacer" just mean line break (`Br`)?

If this is used for all of the tests, when do we actually still want it included? Does it make sense just not to add the `Br` as part of the components, and manually add it anywhere we do want it instead?

ml_peg/app/utils/build_components.py
```diff
@@ -328,6 +328,35 @@ def build_weight_components(
...
+    Div
+        Weight controls with the leading spacer removed while preserving any stores.
+    """
+    children = weight_components.children
+    if isinstance(children, tuple):
+        children = list(children)
+    elif not isinstance(children, list):
+        children = [children]
+
+    if children and children[0].__class__.__name__ == "Br":
```
ElliottKasoar: Can we not use `isinstance`?

ml_peg/app/utils/build_components.py
```diff
@@ -681,6 +711,7 @@ def build_test_layout(
     )
 
     # Inline normalization thresholds when metadata is supplied
+    threshold_controls = None
     if thresholds is not None:
```
ElliottKasoar: Why is this newly needed?

I know the second line (`if thresholds is not None:`) is not new, but in a similar vain, are either of these actually optional in practice at this point? The same for weights too.

I'm pretty sure the tests fail to build somewhere if we haven't defined thresholds? Possibly the same for weights too?

More generally, which parts are actually doing the heavy-lifting of fixing the widths?

joehart2001: It was to make thresholds optional but i think youre right and we dont want that.

ml_peg/app/utils/build_components.py
```diff
@@ -715,42 +717,29 @@ def build_test_layout(
...
+            "borderRadius": "6px",
+            "padding": "0px 0px 0px 0px",  # top right bottom left
+            "marginTop": "-5px",
+            "boxSizing": "border-box",
+            "width": "100%",
+        },
+    )
+
+    table_section: list[Component] = [Div(table)]
+    table_section.extend([Br(), controls_visual])
```
ElliottKasoar: Can we not just do this in a single line? Also not sure the `list[Component]` type hint adds much

ElliottKasoar (commented):
I think this was already the case to some extent, but just to note some model columns are quite squashed.

Image

Also the overall category summary table doesn't seem to have the same fix applied? It's not the worst offender but still takes up quite a lot of space by default

joehart2001:
> I think this was already the case to some extent, but just to note some model columns are quite squashed.
> 
> Image
> Also the overall category summary table doesn't seem to have the same fix applied? It's not the worst offender but still takes up quite a lot of space by default

ah yep i didnt apply it to that, ok will take a look

joehart2001:
made a few more commits to:
- fix the mlip name issue
- tighten up the tables with less padding and more titles over multiple lines

ml_peg/app/utils/build_components.py
```diff
@@ -715,42 +712,29 @@ def build_test_layout(
...
```
ElliottKasoar:
```suggestion
    table_section = [Div(table), Br(), controls_visual]
```
Any reason not to do this in one go?

ElliottKasoar (approved):

PR #497, opened 2026-04-24
⚡ Bolt: Fast EXYZ Metadata Parsing

alinelena:
💡 What: Created a new utility function `read_extxyz_info_fast` in `ml_peg/analysis/utils/utils.py` to quickly extract metadata from Extended XYZ files without parsing atom positions. Implemented this optimization in the Wiggle150 analysis benchmark script.
🎯 Why: `ase.io.read` parses the full atomic structure, which is extremely expensive when we only need the metadata (e.g. energies) located on the second line of the `.xyz` files for simple extraction and metric calculation.
📊 Impact: This speeds up `info` metadata reading by approximately ~6-7x, significantly reducing the bottleneck for iteration-heavy benchmarking and data-aggregation tasks.
🔬 Measurement: Run local benchmarking comparing `ase.io.read(path).info` vs. `read_extxyz_info_fast(path)` on an `.xyz` file. Passed full test suite (`uv run pytest tests/`) and linter (`uv run ruff check`).

---
*PR created automatically by Jules for task [16069877508202070340](https://jules.google.com/task/16069877508202070340) started by @alinelena*


ElliottKasoar:
We could consider this, but it doesn't seem terribly robust, and doesn't solve reading files with multiple structures.

On a related note, we could try to support https://github.com/libAtoms/extxyz

PR #500, opened 2026-04-26
precision keywords

### LINKED ISSUE #412: Document/recommend setting default dtype
### Summary

It would be good to clarify how/where to set the default dtype for new benchmarks.

It's probably worth recommending explicitly setting this, even if it aligns with existing defaults.

### Example

_No response_

### Alternatives

_No response_

### Additional context

_No response_

kuryla:

## Summary

Introduce high and low precision keywords for the float dtype used in tests. Introduce default precision (if the test does not specify) and option to overwrite the dtype by the user.

## Linked issue

Resolves #412


ml_peg/calcs/bulk_crystal/elasticity/calc_elasticity.py
```diff
@@ -230,55 +82,24 @@ def run_elasticity_benchmark(
...
+    Run elasticity benchmark for a single model.
 
     Parameters
     ----------
     mlip
         Model entry containing name and object capable of providing a calculator.
     """
     model_name, model = mlip
-    calc = model.get_calculator()
+    calc = model.get_calculator(precision="high")
```
ElliottKasoar:
```suggestion
    calc = model.get_calculator(precision="low")
```
I think from discussion with @joehart2001 we want this to be low

ml_peg/models/get_models.py
ElliottKasoar, line 172: This isn't a new problem as such, but can I suggest adding a warning here if `overwrite_dtype` is not `None`, which would help identity issues similar to the one we had for some mace models where the type was being ignored.

ml_peg/calcs/bulk_crystal/elasticity/calc_elasticity.py
kuryla: Ah sorry I missed this

ml_peg/models/get_models.py
kuryla: sounds good

ml_peg/models/models.py
```diff
@@ -155,19 +167,26 @@ def get_calculator(self, **kwargs) -> Calculator:
         os.environ["TORCH_DISABLE_MODULE_HIERARCHY_TRACKING"] = "1"
 
         method = getattr(pretrained, self.name)
+
+        precision_map = {"low": "float32", "high": "float64"}
```
ElliottKasoar: Do we not need to change `float32` here?

ml_peg/models/models.py
```diff
@@ -112,10 +119,13 @@ def get_calculator(self, **kwargs) -> Calculator:
...
         """
+        precision_map = {"low": "float32", "high": "float64"}
+        kwargs["default_dtype"] = precision_map[precision]
+
         if self.default_dtype is not None:
-            kwargs["dtype"] = self.default_dtype
+            kwargs["default_dtype"] = self.default_dtype
         else:
-            kwargs["dtype"] = self.default_dtype
+            kwargs["default_dtype"] = self.default_dtype
```
ElliottKasoar: These kwargs need to match the calculator specific parameters. For pet-mad (and upet when we update), I think this is supposed to be `dtype`: https://lab-cosmo.github.io/upet/latest/api.html#calculators

Otherwise it will either error out of be ignored, depending on how spare kwargs are handled

ml_peg/models/models.py
kuryla: yes

ml_peg/models/models.py
ElliottKasoar: This doesn't make sense, does it?

I know this `if...else` isn't new, but it looks like `precision_map` will never be used to set `kwargs["dtype"]`, or am I misreading?

ElliottKasoar (approved):

PR #507, opened 2026-04-28
Add colour scheme customisation

### LINKED ISSUE #518: Fix direction of thresholds based on original defined direction
e.g. original good and bad is 0 and 10, if we move good to 20, bad should update to 20 as well.

joehart2001:
…een/red theme

## Summary
update table colour from viridis_r to RdYlGn_r
- goes with good/bad thresholds
- more intuiative
- more unique and less like other frameworks

As this is on the topic with thresholds, ive added:
- matching threshold box outer to the colour scheme
- preserve direction of initially set good and bad thresholds: if we set in metric.yml good > bad then try to change bad > good in the app, the good threshold will change to be the bad threshold + our smallest increment (as them being the same doesnt make sense)
- the above also fixes an issue we had when we set good=bad

## Linked issue

Resolves #518 #517

### DIFF
--- a/ml_peg/app/build_app.py
+++ b/ml_peg/app/build_app.py
@@ -917,7 +917,7 @@ def build_nav(
                         {"label": "Viridis (colourblind safe)", "value": "viridis_r"},
                         {"label": "Blue-Red (colourblind safe)", "value": "coolwarm"},
                         {
-                            "label": "Green-Red (matches good/bad thresholds)",
+                            "label": "Green-Red",
                             "value": "RdYlGn_r",
                         },
                     ],
--- a/ml_peg/app/build_app.py
+++ b/ml_peg/app/build_app.py
@@ -898,6 +898,39 @@ def build_nav(
         style={"marginBottom": "8px", "fontSize": "13px"},
     )
 
+    _summary_label_style = {
+        "cursor": "pointer",
+        "fontWeight": "600",
+        "fontSize": "11px",
+        "textTransform": "uppercase",
+        "letterSpacing": "0.07em",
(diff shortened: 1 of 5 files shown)


ElliottKasoar (commented):
This is nice!

The text on yellow for red-green is quite hard to read:

Image

Also, is the selection supposed to persist between visits? It doesn't seem to, but it probably makes sense if it does?

joehart2001:
> This is nice!
> 
> The text on yellow for red-green is quite hard to read:
> 
> Image
> Also, is the selection supposed to persist between visits? It doesn't seem to, but it probably makes sense if it does?

ok yep agree, i hadnt thought of doing that. text wise should be the same way we do for swithcing from white to black for the viidiris table

ElliottKasoar:
As suggested by @alinelena, let's change the threshold box colours to match this too?

ml_peg/app/build_app.py
```diff
@@ -898,6 +898,42 @@ def build_nav(
...
+        [
+            Summary("Colour scheme", style=_summary_label_style),
+            Div(
+                Dropdown(
+                    id="cmap-dropdown",
+                    options=[
+                        {"label": "Viridis (colourblind safe)", "value": "viridis_r"},
+                        {"label": "Blue-Red (colourblind safe)", "value": "coolwarm"},
+                        {
+                            "label": "Green-Red (matches good/bad thresholds)",
```
ElliottKasoar:
```suggestion
                            "label": "Green-Red",
```
They all match now

ElliottKasoar (approved):

PR #508, opened 2026-04-28
Add models file input option

### LINKED ISSUE #496: Support file path for models in CLI
### Summary

Some people swap between different yaml files with model inputs.

Rather than manually swapping these files so the correct one is models.yml, we could allow a file path to be given instead of a list of models.

We'd need to check if the input is a path or not, and then decide whether to convert to a list of read the file.

### Example

_No response_

### Alternatives

_No response_

### Additional context

_No response_

ElliottKasoar:

## Summary

Allow models YAML file to be specified for all CLI commands (`ml_peg calc/analyse/app/list`).

I'd originally thought this may be an option parsed instead of the list of models, but I think a separate input makes sense, since then it's also filterable.

This also moves the 'global' inputs to the models init file, which means they can be accessed without the relatively heavy imports in models.py, which otherwise slow down the CLI noticeably.

## Linked issue

Resolves #496

## Testing

Tested locally for most commands, but could use fairly careful additional testing since it touches a lot.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -47,7 +47,9 @@ def pytest_configure(config):
 
     # Set current models from CLI input
     models.current_models = config.getoption("--models")
-    models.models_file = config.getoption("--models-file")
+    model_file = config.getoption("--models-file")
+    if model_file:
+        models.models_file = model_file
 
 
 def pytest_collection_modifyitems(config, items):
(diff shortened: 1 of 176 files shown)


joehart2001:
think you've missed: `physicality/oxidation_states/` so i get an error on app launch

ElliottKasoar:
Good catch, thanks! This predated the merge of PR. Should be fixed now.

joehart2001 (approved):

PR #522, opened 2026-05-02
add elastic tensor voigt form check and conversion if needed

joehart2001:

## Summary

    For some models, matcalc returns the tensor as a plain 
    numpy.ndarray with shape (3, 3, 3, 3), not as a pymatgen 
    tensor object. NumPy arrays do not have .voigt, so the 
    final dataframe conversion fails -> we need to be robust to this

@Fraser-Birks i can't add you as a reviewer but would be good to scan over, thanks!

## Linked issue

Resolves #

## Testing
ran elas benchmark

### DIFF
--- a/ml_peg/calcs/bulk_crystal/elasticity/calc_elasticity.py
+++ b/ml_peg/calcs/bulk_crystal/elasticity/calc_elasticity.py
@@ -160,16 +160,35 @@ class CustomElasticityBenchmark(Benchmark):
             ),
             f"crystal_system_{model_name}": key,
         }
-        
-        
+
+
 def elastic_tensor_to_voigt(x):
     """
     Convert elastic tensor-like objects to 6x6 Voigt form.
-    For some models, matcalc returns the tensor as a plain 
-    numpy.ndarray with shape (3, 3, 3, 3), not as a pymatgen 
-    tensor object. NumPy arrays do not have .voigt, so the 
-    final dataframe conversion fails -> we need to be robust to this
 
+    For some models, matcalc returns the tensor as a plain
+    numpy.ndarray with shape (3, 3, 3, 3), not as a pymatgen
+    tensor object. NumPy arrays do not have .voigt, so the
+    final dataframe conversion fails, so this conversion must handle both forms.
+
+    Parameters
+    ----------
+    x : object
+        Elastic tensor-like object to convert. This may be None, NaN, an object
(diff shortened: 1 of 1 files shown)


ml_peg/calcs/bulk_crystal/elasticity/calc_elasticity.py
```diff
@@ -162,6 +162,77 @@ def process_result(self, result: dict | None, model_name: str) -> dict:  # noqa:
...
+    TypeError
+        If ``x`` is a dictionary without a raw tensor entry.
+    ValueError
+        If the tensor shape is unsupported.
+    """
+    if x is None:
+        return None
+
+    if isinstance(x, float) and np.isnan(x):
+        return None
```
ElliottKasoar: Similar to comments in #526, we may want to keep NaNs around in general (although that can be handled elsewhere, since `None` is probably also a fail.

Do we actually get both None and NaNs?

ml_peg/calcs/bulk_crystal/elasticity/calc_elasticity.py
```diff
@@ -162,6 +162,77 @@ def process_result(self, result: dict | None, model_name: str) -> dict:  # noqa:
...
+
+    # pymatgen Tensor / ElasticTensor case
+    if hasattr(x, "voigt"):
+        return np.asarray(x.voigt)
+
+    # benchmark dict case, e.g. DFT field
+    if isinstance(x, dict):
```
ElliottKasoar: I don't entirely understand what this means? Most of the docstring suggests either an array or a pymatgen tensor, not a dict?

ml_peg/calcs/bulk_crystal/elasticity/calc_elasticity.py
```diff
@@ -162,6 +162,77 @@ def process_result(self, result: dict | None, model_name: str) -> dict:  # noqa:
         }
 
 
+def elastic_tensor_to_voigt(x):
```
ElliottKasoar: I would avoid `x` for anything more than trivial one-liners etc. Also type hints?

joehart2001:
@ElliottKasoar i think this is ok now

joehart2001:
@ElliottKasoar resolved conflicts here too

ElliottKasoar (approved):

PR #523, opened 2026-05-02
Energy response

viktorsvahn:

## Summary
Energy response of linear organic molecules of varying chain length to an external electrical field.

## Linked issue
#433 

## Progress
- [x] Calculations
- [x] Analysis
- [x] Application
- [ ] Documentation

## Testing
mace-mp-0b3
mace-polar-1-s

## New decorators/callbacks
N/A


ml_peg/calcs/electric_field/energy_response/calc_energy_response.py
```diff
@@ -0,0 +1,84 @@
...
+    model_name, model = mlip
+    model.default_dtype = "float64"
+    clean_calc = model.get_calculator()
+
+
+    # Download data
+    data_path = download_github_data(
+        filename="LINEAR_CARBON_wB97M-V.zip",
+        github_uri="https://github.com/viktorsvahn/teoroo_ML-PEG/raw/refs/heads/main/data/source",
+    )
```
joehart2001, line 44: woudl you like us to uplaod this to the s3 bucket or are you happy with this download?

ml_peg/analysis/electric_field/energy_response/metrics.yml
```diff
@@ -0,0 +1,18 @@
+metrics:
+  Total MAE:
+    good: 0.01
+    bad: 5.0
+    unit: eV
```
joehart2001: should these units be eV or kJ/mol?

ml_peg/analysis/electric_field/energy_response/metrics.yml
```diff
@@ -0,0 +1,18 @@
+metrics:
+  Total MAE:
+    good: 0.01
+    bad: 5.0
+    unit: eV
+    tooltip: "Mean Absolute Error for all systems"
+    level_of_theory: wB97V-M
```
joehart2001: should this end in V not M? like stated in the calc file

ml_peg/calcs/electric_field/energy_response/calc_energy_response.py
```diff
@@ -0,0 +1,84 @@
...
+        else:
+            mols = read(data_path/'data'/dataset,':')
+            write(DATA_PATH/dataset, mols)
+
+        for mol in mols:
+            mol.calc = copy(clean_calc)
+            _ = mol.get_potential_energy()
+
+            mol_name = str(mol.get_chemical_formula())
+            mol_dir = APP_ROOT / f"data/electric_field/energy_response/{model_name}"
```
joehart2001: we should avoid writing directory to the app data in the calc script. usually we do this in the analysis step which does often duplicate files, but keeps everything modular

joehart2001:
Hi @viktorsvahn thanks for the PR, looking really nice so far!

a few initial points:
- would it be worth adding dispersion corrections for models trained on non-dispersion corrected functionals? we do this in other benchmarks when relevent e.g. the reference data contains dispersion corrections. see [here](https://github.com/ddmms/ml-peg/blob/main/ml_peg/calcs/molecular_crystal/X23/calc_X23.py#L42)
- i know htis is a shorter benchmark but it would be good to have a tqdm progress bar
- do we need both the analysis scripts?
- i see what youve done with the `metrics.yml`, but i think we should write it out in full for consistency
- is `app_scaling_pol.py` used?
- would you be able to restore `docs/source/tutorials/python/adding_benchmark.ipynb` and `models.yml`? this will fix the conflicts + we won't need these edits in the production code
- rebase for latest changes which may effect some imports

ElliottKasoar:
Hi @viktorsvahn, just wanted to check in and see if there's anything we can do to help?

In addition to the above comments, please can you take a look at our new filtering guidelines: https://ddmms.github.io/ml-peg/developer_guide/filter.html

The principles are relatively simple, but you do have to be a little careful, so again, if anything is unclear, please do ask!

viktorsvahn:
Sorry for taking so long. I am currently on it and I will look at the new guidelines as well!

ElliottKasoar:
Hi @viktorsvahn, thanks for the updates. I hope you don't mind, I just pushed a few minor changes/fixes:

- Rebased
- Fixed reading the input data, and removed the skip, since I think it's quick enough that we don't really need to worry about it
- Updated the way the calculator precision is set
- Fixed some imports
- Reverted some stray changes
- Applied the pre-commit

The pre-commit is still failing, as I only applied the fixes that could be done automatically, so please run this (`pre-commit run --all-files` and resolve the remaining issues. 

Generally it's looking really nice! I think the filtering changes to save the info.json and pass this to the app are still missing. This can probably also get the labels at the same time, but it can also be kept as is.

There also seem to be two different analysis files. Is this an old duplicate?

Please also add this to the documentation. I think you'll need to add a new category file in `docs/source/user_guide/benchmarks/` and update the index.rst there too, since we don't currently have one for electric fields.

I'd also suggest considering a very slight rename of the benchmark, e.g. to include the type of systems, as otherwise energy response is quite generic.

ml_peg/calcs/electric_field/energy_response/calc_energy_response.py
```diff
@@ -0,0 +1,73 @@
...
+    Parameters
+    ----------
+    mlip
+        Name of model use and model to get calculator.
+    """
+    model_name, model = mlip
+    clean_calc = model.get_calculator(precision="high")
+
+    data_path = download_github_data(
+        filename="LINEAR_CARBON_wB97M-V.zip",
```
ElliottKasoar: Would you be able to reupload this as `energy_response.zip`, and make sure it unzips to `energy_response`. Currently it unzips to `data`, which can cause clashes in our cache (and it's useful if the zip file is also benchmark-specific, although that clash is less likely), so you'd need to unzip this, rename the directory, and then re-compress it.

ElliottKasoar:
Hi @viktorsvahn, just wanted to check if there's anything we can do to help with this?

joehart2001 (changes requested):

PR #526, opened 2026-05-04
make elasticity result processing robust to missing data

joehart2001:
… before the checks

## Summary
Issue was defining a key before any missing data checks, which required data to be present. now we avoid this

## Linked issue

Resolves #

### DIFF
--- a/ml_peg/calcs/bulk_crystal/elasticity/calc_elasticity.py
+++ b/ml_peg/calcs/bulk_crystal/elasticity/calc_elasticity.py
@@ -141,7 +141,6 @@ class CustomElasticityBenchmark(Benchmark):
         dict
             Dictionary of processed elastic properties.
         """
-        key = get_crystal_system(result["final_structure"])
         return {
             f"bulk_modulus_vrh_{model_name}": (
                 result["bulk_modulus_vrh"] * eVA3ToGPa
@@ -158,7 +157,11 @@ class CustomElasticityBenchmark(Benchmark):
                 if result is not None
                 else float("nan")
             ),
-            f"crystal_system_{model_name}": key,
+            f"crystal_system_{model_name}": (
+                get_crystal_system(result["final_structure"])
+                if result is not None
+                else None
+            ),
         }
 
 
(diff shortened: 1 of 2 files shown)


ml_peg/calcs/bulk_crystal/elasticity/calc_elasticity.py
```diff
@@ -158,7 +157,11 @@ def process_result(self, result: dict | None, model_name: str) -> dict:  # noqa:
                 if result is not None
                 else float("nan")
             ),
-            f"crystal_system_{model_name}": key,
+            f"crystal_system_{model_name}": (
+                get_crystal_system(result["final_structure"])
+                if result is not None
+                else None
```
ElliottKasoar: Do you think we should maybe do `NaN` rather than `None` for things that have failed?

It's a bit tricky to make sure we do this completely consistently, but in principle it's useful to distinguish say when a model hasn't been run from when a model failed to run a specific structure?

Also we use nan above for similar?

ml_peg/calcs/bulk_crystal/elasticity/calc_elasticity.py
```diff
@@ -237,11 +240,12 @@ def run_elasticity_benchmark(
     atoms_list = []
     for _, row in results.iterrows():
         struct = row.get("final_structure")
-        if struct is not None:
-            atoms = AseAtomsAdaptor.get_atoms(struct).copy()
-            atoms.calc = None
-            atoms.info = {"mp_id": row[benchmark.index_name]}
-            atoms_list.append(atoms)
+        if not isinstance(struct, Structure):
+            continue
```
ElliottKasoar, line 277: Is it a problem if the atoms_list skips some structures in terms of piecing together later which ones failed?

ElliottKasoar (dismissed):
Looks good, as long as my comment is not a problem

joehart2001:
> Looks good, as long as my comment is not a problem

i made a fix in the analysis to your comment, but then running pet-mad revealed that the tensor maes were not robust to missing data so ive added some more fixes on top. also a fix to do with saving the pet-mad tensors too, which was just hte finla part of the calc so jsut had to rerun the part that writes the csv from the checkpoints

ElliottKasoar:
@joehart2001 can you resolve the conflict/docs error?

joehart2001:
> @joehart2001 can you resolve the conflict/docs error?

@ElliottKasoar should be good now

ElliottKasoar (approved):

PR #529, opened 2026-05-05
sort tables and pages alphabetically

joehart2001:

## Summary
I think this is particularly useful as we exapnd the number of categories 

## Linked issue

Resolves #

### DIFF
--- a/ml_peg/app/build_app.py
+++ b/ml_peg/app/build_app.py
@@ -460,7 +460,7 @@ def build_category(
 
         # Build category summary table
         summary_table = build_summary_table(
-            all_tables[category],
+            dict(sorted(all_tables[category].items())),
             table_id=f"{category_title}-summary-table",
             description=category_descrip,
             weights={f"{key} Score": value for key, value in benchmark_weights.items()},
@@ -481,7 +481,7 @@ def build_category(
         )
 
         test_entries = []
-        for test_name in all_layouts[category]:
+        for test_name in sorted(all_layouts[category]):
             framework_id = all_frameworks[category][test_name]
             framework_ids.add(framework_id)
             test_entries.append(
@@ -855,7 +855,7 @@ def build_nav(
     """
     category_paths = {
         category_name: _category_to_path(category_name)
-        for category_name in category_views
+        for category_name in sorted(category_views)
     }
     framework_order = sorted(
         framework_views,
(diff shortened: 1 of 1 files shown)


ElliottKasoar (commented):
I'd been thinking we ought to do something like this, nice!

I think it works, but can we not do this earlier, e.g. if we apply a sort to `for test in tests:` in `get_all_tests`?

It might save us resorting several times/making extra copies of things, which is potentially duplicating quite a lot of data

joehart2001:
> I'd been thinking we ought to do something like this, nice!
> 
> I think it works, but can we not do this earlier, e.g. if we apply a sort to `for test in tests:` in `get_all_tests`?
> 
> It might save us resorting several times/making extra copies of things, which is potentially duplicating quite a lot of data

But i think we still need to sort at multiple levels? Im not sure it makes it better?

ElliottKasoar:
> > I'd been thinking we ought to do something like this, nice!
> > I think it works, but can we not do this earlier, e.g. if we apply a sort to `for test in tests:` in `get_all_tests`?
> > It might save us resorting several times/making extra copies of things, which is potentially duplicating quite a lot of data
> 
> But i think we still need to sort at multiple levels? Im not sure it makes it better?

So there are two things I think we want to improve:

1. How many times we sort
2. Whether we need to make copies of data, and if we do need to make copies, of what. It would be much cheaper/quicker to copy a sorted list of categories than a full dictionary storing tables, layouts etc. that are pretty large objects, for example

If you sort `{category}/{tests}/app_{test}.py`, which is what sorting `tests` does, it should automatically sort both categories and tests once you break things down. From a quick test I think the outcome is the same, and assuming I'm correct that it works, it solves both issues better than I'd have hoped for.

it should mean we build and store the layouts in order, since it will integrate through all tests in a category, then move on to the next category, so it makes sense that we wouldn't need any further sorting?

joehart2001:
I think sorting tests is good to add but it doesnt replace sorting the visible app titles e.g. sorting
```
ml_peg/app/surfaces/OC157/app_OC157.py
ml_peg/app/surfaces/S24/app_S24.py
ml_peg/app/surfaces/elemental_slab_oxygen_adsorption/...
```
vs sorting
```
Elemental Slab Oxygen Adsorption
OC157
S24
```

Also, dict(sorted(...)) is only a shallow copy of the mapping; it copies references to the existing Dash components/tables, not the full table/layout objects

opinions?

ElliottKasoar:
> I think sorting tests is good to add but it doesnt replace sorting the visible app titles e.g. sorting
> 
> ```
> ml_peg/app/surfaces/OC157/app_OC157.py
> ml_peg/app/surfaces/S24/app_S24.py
> ml_peg/app/surfaces/elemental_slab_oxygen_adsorption/...
> ```
> 
> vs sorting
> 
> ```
> Elemental Slab Oxygen Adsorption
> OC157
> S24
> ```
> 
> Also, dict(sorted(...)) is only a shallow copy of the mapping; it copies references to the existing Dash components/tables, not the full table/layout objects
> 
> opinions?

Ahh good point, I wasn't thinking about the human-readable names, and yeah ok the copy isn't as bad as I first thought.

ElliottKasoar (approved):

PR #539, opened 2026-05-08
Add polymer simulations benchmark

### LINKED ISSUE #538: Draft: Polymer Simulation Benchmark
### Summary

Add polymer-density benchmark (`molecular_dynamics/polymers`)
 
 Adds a `molecular_dynamics/polymers` benchmark that scores MLIP predictions
 of room-temperature amorphous polymer densities against the 130-polymer
 experimental dataset from [SimPoly] ([arXiv:2510.13696]).
 
 [arXiv:2510.13696]: https://arxiv.org/abs/2510.13696

### Interactive features

Scatter plot

### Category

molecular_dynamics

### Data availability

Experimental data

### Computational cost

Simulation time is around 2.5 ns per polymer.

### Additional context

_No response_

gncs:

## Summary

ASE port of [SimPoly](https://github.com/microsoft/simpoly)'s 24-stage Polymatic-style equilibration ([arXiv:2510.13696](https://arxiv.org/abs/2510.13696)). Predicts room-temperature density for 130 polymers and reports MAE (g/cm³)  against experiment. ~2.56 ns per (model, polymer); `--time-prefactor` exposed for scaling.
 
 Optional `generation/` sub-package (requires `emc-pypi>=2025.8.21`) builds  starting structures from SMILES; the benchmark itself consumes pre-built structures from S3.

ZIP file just for testing:
[polymers.zip](https://github.com/user-attachments/files/27523251/polymers.zip)

## Linked issue

Resolves #538

## Progress

- [x] Calculations
- [x] Analysis
- [x] Application
- [x] Documentation

## Testing

- Vivace [microsoft/simpoly](https://github.com/microsoft/simpoly)

## New decorators/callbacks

 None


joehart2001:
Hey @gncs 

I've done some testing for mace-mp-0a with: `--time-prefactor 0.0001` for the test polymers you sent and the calc and analysis workflows are working well! I've made the app part with structure visualiation if you want me to push this:
Screenshot 2026-05-08 at 18 12 31
Screenshot 2026-05-08 at 18 12 18


One question:
- do we want to add dispersion? How our code works is that if you add dispersion, it will only add the correction for models which are not trained on dispersion corrected data. before its been relevant for comparing models to experimental data, so just curious

gncs:
Thank you, @joehart2001.

- The GUI looks great - feel free to push this to my PR.
- Yes, adding dispersion should be an option. Could you add that, or tell me how to do it?
- How about we restrict the benchmark to the following polymers for now: `[PE, PIB, PET, PCTFE, PS, PVMS, PVC, PAN, PTPC, PODPM]`

gncs:
Full list of polymers
[polymers.zip](https://github.com/user-attachments/files/27587035/polymers.zip)

ml_peg/calcs/molecular_dynamics/polymers/calc_polymers.py
```diff
@@ -0,0 +1,114 @@
...
+        calc_utils.download_s3_data(
+            filename="polymers.zip",
+            key="inputs/molecular_dynamics/polymers/polymers.zip",
+        )
+        / "polymers"
+    )
+    input_xyz_path = data_path / f"{poly_id}.xyz"
+
+    model_name, model = mlip
+    calc = model.get_calculator(precision="low")
```
joehart2001, line 81:
```suggestion
    calc = model.get_calculator(precision="low")
    calc = model.add_d3_calculator(calc)
```

joehart2001:
> Thank you, @joehart2001.
> 
> * The GUI looks great - feel free to push this to my PR.
> * Yes, adding dispersion should be an option. Could you add that, or tell me how to do it?
> * How about we restrict the benchmark to the following polymers for now: `[PE, PIB, PET, PCTFE, PS, PVMS, PVC, PAN, PTPC, PODPM]`

Thanks, i've pushed the app and added a code suggestion for adding dispersion. Currently, this would mean all models not trained on dispersion interactions would have them applied for the tests. Is this what we want (just to double check)? This is what we do for our [liquid densities MD](https://github.com/ddmms/ml-peg/blob/main/ml_peg/calcs/molecular_dynamics/liquid_densities/calc_liquid_densities.py#L160)

To add the proposed small, medium and large suggestion here too:
- small subset: [PE, PIB, PET, PCTFE, PS, PVMS, PVC, PAN, PTPC, PODPM]
- medium subset: 50 polymers
- large subset: full set of polymers

If you think this is sensible, we can figure out the best way to implement this.

ml_peg/calcs/molecular_dynamics/polymers/calc_polymers.py
gncs: Looks good. Is there a flag to deactivate that for specific models or activate it on demand? We didn't include that with Vivace, for instance.

ElliottKasoar: Currently if a model doesn't support this, we just drop/ignore it either when we parse the `models.yml` config or within the model's `get_calculator` definition.

I'm not completely sure if this currently works for generic ASE calculators that we don't do extra handling for, but it's something that we need to make sure we handle safely, so it definitely will work in future.

gncs: Great, now I understand.

gncs:
> To add the proposed small, medium and large suggestion here too:
> 
> * small subset: [PE, PIB, PET, PCTFE, PS, PVMS, PVC, PAN, PTPC, PODPM]
> * medium subset: 50 polymers
> * large subset: full set of polymers
> 
> If you think this is sensible, we can figure out the best way to implement this.

Sounds good.

ml_peg/calcs/molecular_dynamics/polymers/conftest.py
```diff
@@ -0,0 +1,57 @@
...
+        Pytest argument parser injected by the framework.
+    """
+    parser.addoption("--poly-id", action="store", required=True, type=str)
+    parser.addoption("--time-prefactor", action="store", default=1.0, type=float)
+
+
+@pytest.fixture
+def poly_id(request: pytest.FixtureRequest) -> str:
+    """
+    Return the ``--poly-id`` value (a polymer identifier from data.csv).
```
ElliottKasoar, line 27: Apologies if I missed an alternative way to use this, but for the benefit of automation, might it be worth assigning an index to these IDs and setting that instead, so we don't need to read in the csv/hard code in a slurm script?

gncs: Good question. Are you imagining something like `--poly-idx=17`?

ml_peg/calcs/molecular_dynamics/polymers/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+Predicts room-temperature amorphous densities for **130 polymers** with an
+ML interatomic potential and compares against the experimental reference in
+[`resources/data.csv`](resources/data.csv). **Metric:** MAE (g/cm³).
+
+## Protocol
+
+24-stage Polymatic-style equilibration ported from
+[SimPoly](https://github.com/microsoft/simpoly)'s LAMMPS
+`build_21steps_protocol` to ASE. 0.5 fs timestep, **2.56 ns total** per
```
joehart2001:
```suggestion
`build_21steps_protocol` to ASE. 0.5 fs timestep, **2.06 ns total** per
```

joehart2001:
first error: `AttributeError: module 'ase.md.nose_hoover_chain' has no attribute 'MaskedMTKNPT'`
- fixed by updating ASE from 3.27 to 3.28

joehart2001:
I have added these small medium and large subsets. initially i wanted to use argumetns in the conftest to select these, but it seemed not practical for actually running them... i think this way works well. It would be good to have some explaination of why those 10 polymers have been chosen for the small set

ml_peg/calcs/molecular_dynamics/polymers/conftest.py
ElliottKasoar: Yes, exactly. For example, to run the water density test we often use something along the lines of:

```
for i in {0..3}; do
    srun --exclusive --ntasks=1 --gpus=1 \
         --output=water_${i}_${model}.out --error=water_${i}_${model}.err \
        ml_peg calc --test water_density --models $model --temperature-idx $i &
done
```

gncs: Would this not work?

```bash
 SET_FILE="ml_peg/calcs/molecular_dynamics/polymers/resources/polymer_sets/${POLY_SET}.txt"
 mapfile -t poly_ids < "$SET_FILE"
 
for poly_id in "${poly_ids[@]}"; do
     srun --exclusive --ntasks=1 --gpus=1 \
          --output=poly_${poly_id}_${model}.out --error=poly_${poly_id}_${model}.err \
         ml_peg calc --test polymers --models $model --poly-id $poly_id &
 done
 wait
```

or 

```
 #SBATCH --array=1-10
 
 POLY_SET="${POLY_SET:-small}"
 MODEL="${MODEL:-mace-mp-0a}"
 SET_FILE="ml_peg/calcs/molecular_dynamics/polymers/resources/polymer_sets/${POLY_SET}.txt"
 POLY_ID="$(sed -n "${SLURM_ARRAY_TASK_ID}p" "${SET_FILE}")"
 
 ml_peg calc \
   --category molecular_dynamics \
   --test polymers \
   --models "${MODEL}" \
   --poly-id "${POLY_ID}"
```

joehart2001: the second suggestion is the one i added to the [example slurm](https://github.com/ddmms/ml-peg/pull/539/changes#diff-c084ebbacd7254aea50ba80d3fba5d95130d446af86033784d76db429e1323a8R10) and it works well

joehart2001:
@ElliottKasoar i rebased and added in the element filtering. take a look when you can but im happy from my side

ml_peg/analysis/molecular_dynamics/polymers/metrics.yml
```diff
@@ -0,0 +1,29 @@
...
+    tooltip: Mean Absolute Error in density across the large polymer set (85 polymers)
+    level_of_theory: Experimental
+    weight: 1
+  MAE (X-large):
+    good: 0.0
+    bad: 0.1
+    unit: g/cm^3
+    tooltip: Mean Absolute Error in density across the full polymer set (130 polymers)
+    level_of_theory: Experimental
+    weight: 1
```
ElliottKasoar, line 29: Does it make sense to have all of these weighted by default, both because they all overlap in what they measure, and because we won't have results for most of the larger sets anytime soon?

Perhaps since @joehart2001 (I believe) you're mostly starting with the small set, we only weight that initially?

ml_peg/calcs/molecular_dynamics/polymers/README.md
```diff
@@ -0,0 +1,87 @@
...
+[`generation/README.md`](generation/README.md) to rebuild them.
+
+```bash
+# Full protocol (~2.06 ns; hours on a GPU)
+uv run pytest -v -s \
+    ml_peg/calcs/molecular_dynamics/polymers/calc_polymers.py \
+    --poly-id PS --models mace-mp-0a
+
+# Smoke test (~10 ps; minutes on a GPU)
+uv run pytest -v -s \
```
ElliottKasoar, line 38: I'd probably encourage using `ml_peg calc` rather than `pytest` directly to be consistent with arguments (it should pass `--poly-id` etc to pytest, still)

ml_peg/analysis/molecular_dynamics/polymers/analyse_polymers.py
```diff
@@ -0,0 +1,338 @@
...
+    pd.DataFrame
+        The polymer table indexed by ``id``, sorted alphabetically.
+    """
+    df = pd.read_csv(DATA_CSV, na_values=["NaN"], encoding="utf-8", comment="%")
+    return df.set_index("id").sort_index()
+
+
+POLYMER_TABLE = _load_polymer_table()
+
+STRUCT_INFO = analysis_utils.get_struct_info(
```
ElliottKasoar, line 89: If this isn't used anywhere I'd normally just put it in initial `test_...`  function, otherwise this is run on collection of the tests

ml_peg/calcs/molecular_dynamics/polymers/calc_polymers.py
```diff
@@ -0,0 +1,116 @@
...
+        encoding="utf-8",
+        comment="%",
+    )
+    return df.set_index("id").sort_index()
+
+
+POLYMER_TABLE = _load_polymer_table()
+
+
+@pytest.mark.parametrize("mlip", MODELS.items())
```
ElliottKasoar, line 49:
```suggestion
@pytest.mark.very_slow
@pytest.mark.parametrize("mlip", MODELS.items())
```
This will then need some `--run-very-slow` additions to the run commands

ml_peg/app/molecular_dynamics/polymers/app_polymers.py
```diff
@@ -0,0 +1,89 @@
...
+    """Polymer densities benchmark app layout and callbacks."""
+
+    def register_callbacks(self) -> None:
+        """Register callbacks to app."""
+        scatter = read_plot(
+            DATA_PATH / "figure_polymers.json",
+            id=f"{BENCHMARK_NAME}-figure",
+        )
+
+        struct_dir_name = next(
```
ElliottKasoar, line 36: If I'm understanding this correctly, this is just getting a list of polymer structure files, not the trajectories for each model?

Generally if the structures are the same for all models, I'd now try to use the `mock` calculator outputs, since otherwise if the first model happens to have failed for some polymers, it stops visualisation for all models/will find a subset that may be disordered compared to the scatter points.

This also means we don't need to write out the data for each model - that can be handled by `get_struct_info`, if `write_structs=True`

(Or we can reference somewhere independent of both e.g. copy the `polymers/polymers` data to the app directly?)

ml_peg/analysis/molecular_dynamics/polymers/analyse_polymers.py
```diff
@@ -0,0 +1,338 @@
...
+    "MAE (medium)": frozenset(
+        (POLYMER_SETS_DIR / "medium.txt").read_text().splitlines()
+    ),
+    "MAE (large)": frozenset((POLYMER_SETS_DIR / "large.txt").read_text().splitlines()),
+    "MAE (X-large)": frozenset(
+        (POLYMER_SETS_DIR / "x-large.txt").read_text().splitlines()
+    ),
+}
+OUT_PATH = app.APP_ROOT / "data" / "molecular_dynamics" / "polymers"
+LOCAL_STRUCTURES_DIR = calcs.CALCS_ROOT / "molecular_dynamics" / "polymers" / "polymers"
```
ElliottKasoar, line 53: I'm not sure I understand where this comes from? I ran the mock calculator and it wasn't populated

ml_peg/analysis/molecular_dynamics/polymers/analyse_polymers.py
```diff
@@ -0,0 +1,338 @@
...
+def labels() -> list[str]:
+    """
+    Return the list of polymer ids covered by the benchmark.
+
+    Returns
+    -------
+    list[str]
+        Polymer ids from ``data.csv``, in sorted order.
+    """
+    return [str(poly_id) for poly_id in POLYMER_TABLE.index]
```
ElliottKasoar, line 109: This is fine, but if it is/were saved during the calc, or formed part of the file names picked up by `analysis_utils.get_struct_info`, we could get it for free more or less during the `analysis_utils.get_struct_info` call

ElliottKasoar (commented):
Thanks for this, @gncs, it's looking great, and sorry again for the huge delay in review. I need to test a coupe of edge cases with a bit of actual data I generate/from @joehart2001, but generally I think there are just a few small things it would be good to tidy up.

I also wonder if perhaps we should make the benchmark name slightly more specific, e.g. `polymer_densities`? I can imagine polymers may even become its own category down the line.

ml_peg/analysis/molecular_dynamics/polymers/metrics.yml
joehart2001: yes lets weight the small 1 and others 0 by default? otherwise no score will show for most models

PR #547, opened 2026-05-12
Add benchmark for model evaluation on small clusters

### LINKED ISSUE #546: Model evaluation on small clusters
### Summary

Hello everyone and thanks for all the work you guys did on this benchmark suite!

Results from https://doi.org/10.1063/5.0303302 suggest that small clusters would be an important test for aspiring universal MLIPs. This benchmark would consist of the evaluation of model accuracy on isolated clusters of atoms (dimers, trimers, tetramers and so on).

### Interactive features

No interactive features besides modifying the weight of clusters of different sizes.

### Category

This benchmark would most likely define its own category ("clusters"), but including it into "physicality" is also a possibility.

### Data availability

This is a short description of the data we have been able to generate so far for this benchmark.

Cluster generation
We generated clusters containing between 3 and 8 atoms by randomly placing the atoms into a structure whilst ensuring that all interatomic distances were kept between 2.0 and 2.5 Å. We avoided monomers and dimers because we understand that these are tested separately in a different "dimer curve" ML-PEG benchmark, but it would be straightforward and inexpensive to add those too. We randomly chose the identities of the elements to include the first three rows of the periodic table. Indeed, we believe that light metals and main-group elements should nearly always be within the application domain of general-purpose models, including those that primarily focus on organic chemistry. This has the further advantage that all-electron DFT calculations for these clusters are inexpensive, making dataset generation efforts easily reproducible, improvable and extensible by the community. We included 10000 clusters for each size, from 3 to 8 atoms, inclusive.

DFT calculations
We used the ORCA and FHI-AIMS packages to perform DFT calculations that are fully consistent with the OMol25 (ωB97M-V/def2-TZVPD) and MAD-1.5 (r2SCAN) datasets, respectively. OMol-style calculations were performed with a total charge of zero and and spin multiplicities of 1 (in the case of an even total number of electrons in the cluster) or 2 (in the case of an odd number of electrons), as these are the most common and well-represented use cases in the chemistry literature. The convergence rate is very high for both levels of theory; we set energies and forces to NaN in the rare cases of non-convergence of the DFT calculations.

### Computational cost

The computational cost entirely depends on the availability and efficiency of batching, as the clusters are very small. In the best-case scenario (optimal batching), the total compute required by this benchmark would be equivalent to the evaluation of around 50 structures of a few thousand atoms each, hence relatively short.

Serial evaluation would be dominated by the overhead of running the model sequentially through approximately 60k clusters.

### Additional context

_No response_

frostedoyster:

## Summary

As described in #546, this benchmark evaluates force accuracies on small atomic clusters.

## Linked issue

Resolves #546

## Progress

- [x] Calculations
- [x] Analysis
- [x] Application
- [x] Documentation

Potential aspects of the benchmark to be discussed with the maintainers:
- weighting of clusters of different sizes (defaults to one for all sizes at the moment)
- energies are excluded for the moment because models can have arbitrary baselines; however, we could run a few DFT calculations to determine isolated atom energies and subtract those from the cluster energies
- the benchmark is marked as "very slow" for the moment because it consists of the evaluation of 60k structures. This is to be tested though, as all structures are very small. It would also be possible to select a smaller subset.
- the benchmark contains random elements up to (and including) the third row of the periodic table. This includes the presence of clusters including noble gases, which are often not well-represented in training sets. Also depending on the performance of the existing models on those, we might want to decide to exclude cluster containing noble gases

## Testing

We carefully checked consistency of the labels with the publicly available OMol25 and MAD-1.5 datasets. The benchmark has not been tested on any models yet.

## New decorators/callbacks

No new callbacks are needed.

### DIFF
--- a/docs/source/user_guide/benchmarks/clusters.rst
+++ b/docs/source/user_guide/benchmarks/clusters.rst
@@ -8,40 +8,65 @@ Cluster Forces
 Summary
 -------
 
-Performance in predicting forces on fixed geometries of neutral atomistic clusters
-containing three to eight atoms.
+Performance in predicting forces on neutral atomistic clusters containing three to eight
+atoms where the chemical elements are chosen randomly. This benchmark is designed to
+test the universality and generalizability of the interatomic potentials.
 
 Metrics
 -------
 
-1. Force MAE by cluster size
+1. MAD2 force MAE by cluster size
 
 Mean absolute error (MAE) of force components across all atoms and clusters,
 reported separately for clusters containing three, four, five, six, seven, and eight
-atoms.
+atoms. Each model is compared to the ``mad2_forces`` reference values.
 
-Each model is compared to the reference force array most appropriate for its training
-domain. Models trained on broad atomistic datasets, or primarily on materials, are
-compared to ``mad2_forces``. Models focused on organic chemistry are compared to
-``omol25_forces``.
+2. OMol25 force MAE by cluster size
 
-For OMOL25-routed calculations, all clusters are evaluated with ``charge=0`` and
+Mean absolute error (MAE) of force components across all atoms and clusters,
+reported separately for clusters containing three, four, five, six, seven, and eight
+atoms. Each model is compared to the ``omol25_forces`` reference values.
+
+For all calculations, clusters are evaluated with ``charge=0`` and either
 ``spin_multiplicity=1`` for clusters with an even number of electrons or
 ``spin_multiplicity=2`` for clusters with an odd number of electrons.
 
 Computational cost
 ------------------
 
-High: tests evaluate 60,000 clusters per model and are likely to take hours on CPU.
+Medium: tests evaluate 60,000 small clusters per model and are expected to take less
+than an hour locally.
 
 Data availability
 -----------------
 
-Input/reference data:
+Cluster structures and reference forces are distributed as a separate zip archive and
+downloaded on-demand from the ML-PEG data store. The calculation script uses the
+public ML-PEG S3 bucket to retrieve these inputs.
 
-* Cluster structures and reference forces are distributed as a separate zip archive and
-  downloaded on-demand from the ML-PEG data store. The calculation script uses the
-  public ML-PEG S3 bucket to retrieve these inputs.
-* The extended XYZ input contains both ``mad2_forces`` and ``omol25_forces`` reference
-  arrays. Structures with non-finite values in the routed reference array did not
-  converge at that level of theory and they are therefore excluded from the metric.
+The structures are randomly generated neutral clusters ("n-mers") containing three
+to eight atoms. There are 10,000 clusters for each size from three to eight atoms,
+inclusive. Atoms were placed randomly while enforcing all pairwise interatomic distances
+to lie between 2.0 and 2.5 Å; element identities were sampled randomly from the 18
+elements of the first three rows of the periodic table.
+  
+The extended XYZ input contains both ``mad2_forces`` and ``omol25_forces`` reference
+arrays. Non-finite reference or predicted force components are excluded from the
+evaluation of the corresponding metric. These correspond to non-converged DFT
+calculations and/or model failures in the prediction (for example, due to
+unsupported chemical elements).
+
+Reference labels:
+
+* ``mad2_forces``: r2SCAN reference forces calculated with FHI-AIMS, using settings
+  consistent with the MAD-1.5 dataset.
+* ``omol25_forces``: ``ωB97M-V/def2-TZVPD`` reference forces calculated with ORCA,
+  using settings consistent with the OMol25 dataset. OMol-style DFT calculations used
+  total charge zero and spin multiplicities of 1 for even-electron clusters or 2 for
+  odd-electron clusters.
+
+In both cases, energies and forces were set to ``NaN`` in the rare cases where the DFT
+calculation did not converge.
+
+Further details on the motivation for n-mer benchmarks of universal MLIPs are
+available at https://doi.org/10.1063/5.0303302.
--- a/docs/source/user_guide/benchmarks/clusters.rst
+++ b/docs/source/user_guide/benchmarks/clusters.rst
@@ -34,8 +34,9 @@ For all calculations, clusters are evaluated with ``charge=0`` and either
 Computational cost
 ------------------
 
-Medium: tests evaluate 60,000 small clusters per model and are expected to take less
-than an hour locally.
+Medium: these tests evaluate 60,000 small clusters per model. They are expected to take
+on the order of minutes on a GPU.
+
 
 Data availability
 -----------------
--- a/docs/source/user_guide/benchmarks/clusters.rst
+++ b/docs/source/user_guide/benchmarks/clusters.rst
@@ -1,0 +1,47 @@
+========
+Clusters
+========
+
+Cluster Forces
+==============
+
+Summary
+-------
+
+Performance in predicting forces on fixed geometries of neutral atomistic clusters
+containing three to eight atoms.
+
+Metrics
+-------
+
+1. Force MAE by cluster size
+
+Mean absolute error (MAE) of force components across all atoms and clusters,
+reported separately for clusters containing three, four, five, six, seven, and eight
+atoms.
+
+Each model is compared to the reference force array most appropriate for its training
+domain. Models trained on broad atomistic datasets, or primarily on materials, are
+compared to ``mad2_forces``. Models focused on organic chemistry are compared to
+``omol25_forces``.
+
+For OMOL25-routed calculations, all clusters are evaluated with ``charge=0`` and
+``spin_multiplicity=1`` for clusters with an even number of electrons or
+``spin_multiplicity=2`` for clusters with an odd number of electrons.
+
+Computational cost
+------------------
+
+High: tests evaluate 60,000 clusters per model and are likely to take hours on CPU.
+
+Data availability
+-----------------
+
+Input/reference data:
+
+* Cluster structures and reference forces are distributed as a separate zip archive and
+  downloaded on-demand from the ML-PEG data store. The calculation script uses the
+  public ML-PEG S3 bucket to retrieve these inputs.
+* The extended XYZ input contains both ``mad2_forces`` and ``omol25_forces`` reference
+  arrays. Structures with non-finite values in the routed reference array did not
+  converge at that level of theory and they are therefore excluded from the metric.
--- a/docs/source/user_guide/benchmarks/clusters.rst
+++ b/docs/source/user_guide/benchmarks/clusters.rst
@@ -49,7 +49,7 @@ to eight atoms. There are 10,000 clusters for each size from three to eight atom
 inclusive. Atoms were placed randomly while enforcing all pairwise interatomic distances
 to lie between 2.0 and 2.5 Å; element identities were sampled randomly from the 18
 elements of the first three rows of the periodic table.
-  
+
 The extended XYZ input contains both ``mad2_forces`` and ``omol25_forces`` reference
 arrays. Non-finite reference or predicted force components are excluded from the
 evaluation of the corresponding metric. These correspond to non-converged DFT
--- a/docs/source/user_guide/benchmarks/clusters.rst
+++ b/docs/source/user_guide/benchmarks/clusters.rst
@@ -31,6 +31,12 @@ For all calculations, clusters are evaluated with ``charge=0`` and either
 ``spin_multiplicity=1`` for clusters with an even number of electrons or
 ``spin_multiplicity=2`` for clusters with an odd number of electrons.
 
+For models whose training data do not include dispersion, results are shown for both
+the native model and the model with the configured runtime dispersion correction. The
+corrected variant is identified by a suffix such as ``-D3`` and contributes to the
+category and overall scores. Models trained with dispersion already included are shown
+once, without an additional correction.
+
 Computational cost
 ------------------
 
--- a/ml_peg/analysis/clusters/cluster_forces/analyse_cluster_forces.py
+++ b/ml_peg/analysis/clusters/cluster_forces/analyse_cluster_forces.py
@@ -3,14 +3,19 @@
 from __future__ import annotations
 
 from pathlib import Path
+from typing import Any
 
 from ase.atoms import Atoms
-from ase.io import read
+from ase.io import read, write
 import numpy as np
 import pytest
 
 from ml_peg.analysis.utils.decorators import build_table, plot_density_scatter
-from ml_peg.analysis.utils.utils import load_metrics_config, mae
+from ml_peg.analysis.utils.utils import (
+    load_metrics_config,
+    mae,
+    sample_density_grid,
+)
 from ml_peg.app import APP_ROOT
 from ml_peg.calcs import CALCS_ROOT
 from ml_peg.models import current_models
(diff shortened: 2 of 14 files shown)


ml_peg/calcs/clusters/cluster_forces/calc_cluster_forces.py
```diff
@@ -0,0 +1,227 @@
...
+        _set_charge_and_spin(out_atoms)
+        out_atoms.arrays.pop(MAD2_FORCES_KEY, None)
+        out_atoms.arrays.pop(OMOL25_FORCES_KEY, None)
+        out_atoms.info["structure_index"] = structure_index
+        out_atoms.info["reference_forces_key"] = reference_key
+        out_atoms.info["reference_target"] = reference_target
+        out_atoms.arrays["ref_forces"] = ref_forces
+        out_atoms.arrays["pred_forces"] = np.asarray(
+            atoms_pred.get_forces(), dtype=float
+        )
```
ElliottKasoar: Could we add a try/except around this, and maybe store the forces as NaNs if this fails?

Otherwise (unless I'm missing something) models which support a more limited set of elements, for example, will fail at some point.

ml_peg/calcs/clusters/cluster_forces/calc_cluster_forces.py
```diff
@@ -0,0 +1,227 @@
...
+
+OUT_PATH = Path(__file__).parent / "outputs"
+BENCHMARK_FILENAME = "cluster_forces.extxyz"
+S3_KEY = "inputs/clusters/cluster_forces/cluster_forces.zip"
+S3_FILENAME = "cluster_forces.zip"
+INPUT_FILENAME = "clusters.xyz"
+
+MAD2_FORCES_KEY = "mad2_forces"
+OMOL25_FORCES_KEY = "omol25_forces"
+ORGANIC_MODEL_MARKERS = ("omol", "off", "polar")
```
ElliottKasoar: Generally when we have multiple references, we'd still compare all models to each one. Otherwise the scores we assign mean different things to different models, which could be misleading/unfair

ml_peg/app/clusters/cluster_forces/app_cluster_forces.py
```diff
@@ -0,0 +1,80 @@
...
+    """
+    return ClusterForcesApp(
+        name=BENCHMARK_NAME,
+        description=(
+            "Component-wise force MAE for neutral small clusters, split by cluster "
+            "size. Models are routed to MAD2 or OMOL25 reference forces according "
+            "to their training domain."
+        ),
+        table_path=DATA_PATH / "cluster_forces_metrics_table.json",
+        extra_components=[Div(id=f"{BENCHMARK_NAME}-figure-placeholder")],
```
ElliottKasoar: Could you add a `docs_url` to this, corresponding to the URL that will take you to the new documentation?

ml_peg/analysis/clusters/cluster_forces/metrics.yml
```diff
@@ -0,0 +1,43 @@
...
+    unit: eV/A
+    tooltip: "Component-wise force MAE on neutral 7-atom clusters. Materials/general models are compared to MAD2 forces; organic-focused models are compared to OMOL25 forces."
+    level_of_theory: null
+    weight: 1
+  Force MAE (8 atoms):
+    good: 0.1
+    bad: 1.0
+    unit: eV/A
+    tooltip: "Component-wise force MAE on neutral 8-atom clusters. Materials/general models are compared to MAD2 forces; organic-focused models are compared to OMOL25 forces."
+    level_of_theory: null
```
ElliottKasoar: I thought the level of theory for these was either ωB97M-V/def2-TZVPD or r2SCAN? Are the `null`s  because it could be either?

docs/source/user_guide/benchmarks/clusters.rst
```diff
@@ -0,0 +1,47 @@
...
+------------------
+
+High: tests evaluate 60,000 clusters per model and are likely to take hours on CPU.
+
+Data availability
+-----------------
+
+Input/reference data:
+
+* Cluster structures and reference forces are distributed as a separate zip archive and
```
ElliottKasoar: Can you add comments on how the structures were originally obtained/built, and the level of theory of the reference data?

ml_peg/calcs/clusters/cluster_forces/calc_cluster_forces.py
```diff
@@ -0,0 +1,227 @@
...
+    reference_target = "OMOL25" if reference_key == OMOL25_FORCES_KEY else "MAD2"
+
+    calc = model.get_calculator(precision="high")
+    data_path = _cluster_data_path()
+    out_dir = OUT_PATH / model_name
+    out_dir.mkdir(parents=True, exist_ok=True)
+
+    frames = iread(data_path, index=":")
+    if sys.stderr.isatty():
+        frames = tqdm(frames, desc=f"{model_name} cluster forces", unit="cluster")
```
ElliottKasoar: If we know the full number, it might be nice to include that here, since otherwise you don't get the progress bar

ml_peg/app/clusters/cluster_forces/app_cluster_forces.py
```diff
@@ -0,0 +1,80 @@
...
+    def register_callbacks(self) -> None:
+        """Register force parity callbacks."""
+        column_to_plot = {
+            _metric_name(cluster_size): read_plot(
+                DATA_PATH / f"figure_force_parity_{cluster_size}mer.json",
+                id=f"{BENCHMARK_NAME}-{cluster_size}mer-force-parity-figure",
+            )
+            for cluster_size in CLUSTER_SIZES
+        }
+        plot_from_table_column(
```
ElliottKasoar: It might be nice to add structure visualisation to these plots. @joehart2001 can probably help, since it's slightly more complicated for density plots than normal scatters.

ElliottKasoar (commented):
Hi @frostedoyster, thanks for this, it's looking really nice already, and should be a really interesting addition!

@joehart2001 and I will try to look in more detail as soon as we can, but I've left a few comments/questions from an initial pass of everything.

docs/source/user_guide/benchmarks/clusters.rst
joehart2001: i agree, it would be useful to have a lot of info from your email

ml_peg/app/clusters/cluster_forces/app_cluster_forces.py
joehart2001: For the density scatter you can look at e.g. [analysis example](https://github.com/ddmms/ml-peg/blob/main/ml_peg/analysis/non_covalent_interactions/NCIA_D1200/analyse_NCIA_D1200.py) and [app example](https://github.com/ddmms/ml-peg/blob/main/ml_peg/app/non_covalent_interactions/NCIA_D1200/app_NCIA_D1200.py)

but let us know if you're confused about anything or any formats etc

ml_peg/calcs/clusters/cluster_forces/calc_cluster_forces.py
joehart2001: i agree, so we would test all models on all sets

ml_peg/calcs/clusters/cluster_forces/calc_cluster_forces.py
```diff
@@ -0,0 +1,227 @@
...
+    if reference_key not in atoms.arrays:
+        raise KeyError(f"Missing '{reference_key}' in cluster input.")
+
+    ref_forces = np.asarray(atoms.arrays[reference_key], dtype=float)
+    if not np.isfinite(ref_forces).all():
+        return None
+    return ref_forces
+
+
+@pytest.mark.very_slow
```
joehart2001:
```suggestion
@pytest.mark.slow
```

joehart2001: very slow is reserved for calculations which take on the order of days, so i think slow is more appropriate

joehart2001:
Overall its looking great! Would we want to apply dispersion corrections for models which have not been trained on dispersion corrected data? We have the utility to do this automatically, see [here](https://github.com/ddmms/ml-peg/blob/main/ml_peg/calcs/molecular_crystal/X23/calc_X23.py#L41). Im not suggesting that this needs dispersion, just would like to check

ml_peg/calcs/clusters/cluster_forces/calc_cluster_forces.py
ElliottKasoar: I wondered about this. If it does actually take several hours to run, say, locally, I think it could be borderline, but yeah `slow` is probably slightly more accurate still

joehart2001: from testing locally it was less than an hour

ml_peg/analysis/clusters/cluster_forces/metrics.yml
frostedoyster: Yes, indeed. The DFT labels are back now that we're evaluating every model against every level of theory

frostedoyster:
Thanks for the comments and sorry for the delay! I've tried to address all the comments, let me know what you think.

> Would we want to apply dispersion corrections for models which have not been trained on dispersion corrected data?

My understanding is that meta-GGA and hybrid functionals (which we use here) are generally able to capture short- and medium-range dispersion effects, therefore I would suspect that keeping the tests as they are is probably the safest approach. Long-range dispersion effects are not captured, but these clusters are all short-ranged. My concern is that, if we removed dispersion artificially from the predictions of models that were trained with dispersion, we would probably overcorrect and make the models look artificially worse on this specific test. That said, I'm not an expert, so I'm curious to hear your thoughts

joehart2001:
> Thanks for the comments and sorry for the delay! I've tried to address all the comments, let me know what you think.
> 
> > Would we want to apply dispersion corrections for models which have not been trained on dispersion corrected data?
> 
> My understanding is that meta-GGA and hybrid functionals (which we use here) are generally able to capture short- and medium-range dispersion effects, therefore I would suspect that keeping the tests as they are is probably the safest approach. Long-range dispersion effects are not captured, but these clusters are all short-ranged. My concern is that, if we removed dispersion artificially from the predictions of models that were trained with dispersion, we would probably overcorrect and make the models look artificially worse on this specific test. That said, I'm not an expert, so I'm curious to hear your thoughts

Yes i definitely agree for r2scan, and we never artificially remove the dispersion dont worry. But i was wondering for `ωB97M-V/def2-TZVPD` if dispersion would be good to include here for models trianed on e.g. PBE. maybe @ilyes319 is good to ask for this

ElliottKasoar:
Hi @frostedoyster, thanks for all the changes, it's looking great!

Please can you take a look at our new filtering guidelines: https://ddmms.github.io/ml-peg/developer_guide/filter.html, and make any changes accordingly?

The principles are relatively simple, but you do have to be a little careful, so if anything is unclear, please do ask!

(You may need to rebase first)

Another small note, but if you can run the pre-commit to make all of the automated checks pass, that would be great.

frostedoyster:
Hey @joehart2001 and @ElliottKasoar, thanks for the review! I've updated some of the code to handle the new conventions (rebasing, element filtering, app changes). Regarding the comparison between PBE and hybrid with or without dispersion corrections, I think this is a hard problem in general. My instinct would be to leave PBE models as is, but I leave it up to you guys to decide, especially because I suspect that this decision would impact other benchmarks in the suite too.

docs/source/user_guide/benchmarks/clusters.rst
```diff
@@ -0,0 +1,72 @@
...
+atoms. Each model is compared to the ``omol25_forces`` reference values.
+
+For all calculations, clusters are evaluated with ``charge=0`` and either
+``spin_multiplicity=1`` for clusters with an even number of electrons or
+``spin_multiplicity=2`` for clusters with an odd number of electrons.
+
+Computational cost
+------------------
+
+Medium: tests evaluate 60,000 small clusters per model and are expected to take less
```
joehart2001: gpu estimate would be great as well

joehart2001: reference from me: 16 mins on cpu for mace-mp-0b3, i suspect gpu is on the order of low number of minutes. can add: expected to take minutes on GPU

ml_peg/analysis/clusters/cluster_forces/analyse_cluster_forces.py
```diff
@@ -0,0 +1,432 @@
...
+        filename=OUT_PATH
+        / f"figure_force_parity_{reference['key']}_{cluster_size}atoms.json",
+        title=f"{reference['label']} {cluster_size}-atom cluster force components",
+        x_label="Reference force / (eV/A)",
+        y_label="Predicted force / (eV/A)",
+        annotation_metadata={
+            "level_of_theory": "Level",
+            "n_clusters": "Clusters",
+            "n_components": "Components",
+            "excluded_components": "Excluded components",
```
joehart2001:
```suggestion
            "excluded_components": "NaN force components",
```

something more specific like this would be ideal

joehart2001:
Hi @frostedoyster i think its looking and working pretty great. just some minor comments from me. @ElliottKasoar could you take a look?

ilyes319:
i would argue that if we should have the option to see with ans without dispersion correction as this would be the most informative.

frostedoyster:
Hey everyone, thanks for the suggestions, I've included dispersion-corrected variants (for the models that need them). The only problem I see with that is that, at the moment, for models which need dispersion corrections, the dispersion-corrected variant is taken as the preferred one to calculate the overall scores, while the r2SCAN-level labels for the clusters do not contain dispersion corrections. In a way, this would artificially hurt models which are dispersion-corrected.

I think a good solution here would be to add dispersion corrections to the r2SCAN labels, would you guys agree?

joehart2001:
> Hey everyone, thanks for the suggestions, I've included dispersion-corrected variants (for the models that need them). The only problem I see with that is that, at the moment, for models which need dispersion corrections, the dispersion-corrected variant is taken as the preferred one to calculate the overall scores, while the r2SCAN-level labels for the clusters do not contain dispersion corrections. In a way, this would artificially hurt models which are dispersion-corrected.
> 
> I think a good solution here would be to add dispersion corrections to the r2SCAN labels, would you guys agree?

Hey @frostedoyster, thanks for the changes. i have been thinking about this and i think the easiest way to do the `calc` part is to write a general evaluator and then two separate functions, one where you apply dispersion to the calcualtor and one where you dont e.g.:

```
@pytest.mark.slow
def test_cluster_forces_MAD(mlip: tuple[str, Any]) -> None:

.
.
.

@pytest.mark.slow
def test_cluster_forces_OMOL(mlip: tuple[str, Any]) -> None:

    calc = model.add_d3_calculator(calc) # this only adds D3 to calcultors not trained on dispersion corrected datasets
    
```

As for your question, could you clarify slightly: "the dispersion-corrected variant is taken as the preferred one to calculate the overall scores", are you saying there is basically an issue with the keys/model name in the table, so that the model says e.g. `mace-mp-0a+D3` when its actually just `mace-mp-0a` for the r2scan calculation? if this is what you mean, then this is definitely an issue and maybe the best way is just to separate this out into two benchmarks, one for MAD and one for OMOL to keep it clean (instead of the above suggestion). this would fix our other issues too. if this is not what you mean, let me know!

ElliottKasoar:
> Hey everyone, thanks for the suggestions, I've included dispersion-corrected variants (for the models that need them). The only problem I see with that is that, at the moment, for models which need dispersion corrections, the dispersion-corrected variant is taken as the preferred one to calculate the overall scores, while the r2SCAN-level labels for the clusters do not contain dispersion corrections. In a way, this would artificially hurt models which are dispersion-corrected.
> 
> I think a good solution here would be to add dispersion corrections to the r2SCAN labels, would you guys agree?

Hi, perhaps the simplest solution is just to directly compare the models without any correction to the reference data you have for now?

Also, you seem to have picked up some changes (and conflicts) with a few unrelated files. Would you be able to fix this, @frostedoyster?

joehart2001:
> > Hey everyone, thanks for the suggestions, I've included dispersion-corrected variants (for the models that need them). The only problem I see with that is that, at the moment, for models which need dispersion corrections, the dispersion-corrected variant is taken as the preferred one to calculate the overall scores, while the r2SCAN-level labels for the clusters do not contain dispersion corrections. In a way, this would artificially hurt models which are dispersion-corrected.
> > I think a good solution here would be to add dispersion corrections to the r2SCAN labels, would you guys agree?
> 
> Hi, perhaps the simplest solution is just to directly compare the models without any correction to the reference data you have for now?
> 
> Also, you seem to have picked up some changes (and conflicts) with a few unrelated files. Would you be able to fix this, @frostedoyster?

yeah i think we just go back to this appraoch tbh

PR #555, opened 2026-05-17
Phonon calc + structure visualisaiton

joehart2001:

## Summary
Added phonon calc + DFT and MLIP relaxed structure visualisation

Also, about a 1.5x speedup in the calculation time which is nice

## Linked issue

Resolves #

## Testing

running all models now, matches results for mp-0a so far


ml_peg/calcs/bulk_crystal/phonons/phonons_utils.py
```diff
@@ -0,0 +1,317 @@
...
+        atoms,
+        fc2_supercell=fc2_supercell,
+        primitive_matrix=primitive_matrix,
+        symprec=symprec,
+        **kwargs,
+    )
+    if displacement_dataset is not None:
+        phonons.dataset = displacement_dataset
+    else:
+        phonons.generate_displacements(distance=0.01)
```
ElliottKasoar: Is this a safe/sensible default? Is it something we may want to change? Might be better as a parameter of the function?

ml_peg/calcs/bulk_crystal/phonons/phonons_utils.py
```diff
@@ -0,0 +1,317 @@
...
+    return Phonopy(
+        unitcell=_aseatoms2phonoatoms(atoms),
+        supercell_matrix=fc2_supercell,
+        primitive_matrix=primitive_matrix,
+        symprec=symprec,
+        **kwargs,
+    )
+
+
+def phonopy2aseatoms(phonons: Phonopy, primitive: bool | None = None) -> Atoms:
```
ElliottKasoar: These names, especially the first and last, are a little tricky to read, maybe just add some underscores?

ml_peg/calcs/bulk_crystal/phonons/phonons_utils.py
```diff
@@ -0,0 +1,317 @@
...
+        url = f"https://alexandria.icams.rub.de/data/phonon_benchmark/pbe/{mp_id}.yaml.bz2"
+        response = requests.get(url, timeout=15)
+        response.raise_for_status()
+        save_path.write_bytes(bz2.decompress(response.content))
+        return f"{mp_id}: success"
+    except Exception as e:
+        return f"{mp_id}: failed ({e})"
+
+
+def download_alex_parallel(sample_every: int = 1, max_threads: int = 16) -> None:
```
ElliottKasoar, line 278: Can this be refactored with the download functions uses for the low dim / high pressure relaxations? Or are they too different?

ml_peg/app/utils/build_callbacks.py
```diff
@@ -799,7 +800,12 @@ def scatter_and_assets_from_table(
         Fallback callable invoked when ``column_handlers`` has no entry.
     model_key
         Key in ``table_data`` used to look up the model display name.
+    scatter_id
+        Optional graph ID; when provided, download controls are rendered above the
+        scatter and the plot-download callback is registered.
     """
+    if scatter_id is not None:
+        register_plot_download_callbacks()
```
ElliottKasoar: If we're specifically showing a scatter when we call this function, is there a reason we expect this callback to work?

ml_peg/app/utils/build_callbacks.py
```diff
@@ -867,6 +873,9 @@ def _update_scatter_and_metadata(active_cell, last_cell):
             raise PreventUpdate
         content, metadata = result
 
+        if scatter_id is not None:
+            content = Div([build_plot_download_controls(scatter_id), content])
```
ElliottKasoar: Related to above, do we just want this to be required?

ml_peg/calcs/bulk_crystal/phonons/calc_phonons.py
```diff
@@ -0,0 +1,494 @@
...
+MODEL_NAMES = get_model_names(current_models)
+
+OUT_PATH = Path(__file__).parent / "outputs"
+DFT_REF_PATH = OUT_PATH / "DFT"
+
+FMAX = 0.005
+Q_MESH = 6
+Q_MESH_THERMAL = 20
+TEMPERATURES = [0, 75, 150, 300, 600]
+# Tune to GPU memory: reduce for large models (e.g. UMA-M) or increase for >40 GB GPUs.
```
ElliottKasoar, line 47: Is it possible/useful to parameterise any of this to run things in parallel for the same model, similar to what we do for MD?

ElliottKasoar (commented):
Generally looks great! I need to test it more carefully, but I've left some minor comments for now

ElliottKasoar:
What's the main source of speedup out of interest?

ml_peg/calcs/bulk_crystal/phonons/phonons_utils.py
joehart2001: i had to do a parallel downlaod otherwise it takes like 12 hours just to download them

joehart2001:
> What's the main source of speedup out of interest?

mutliple small factors:
- changed 3/6 of the ref outputs (qpoints, labels and connectiosn files) to one pkl file
- probably mlipx node overhead
- threadpool_limits
- yaml loading: CSafeLoader is the C-extension binding to libyaml which can apparently be 10-50x faster

ml_peg/calcs/bulk_crystal/phonons/calc_phonons.py
joehart2001: the only useful parameterisation would be to split the phonons into 10 sets of ~1000, but given that this doesnt actaully take too long compared to other benchmarks its actually ok. some models ran in 6h

ml_peg/calcs/bulk_crystal/phonons/phonons_utils.py
joehart2001: yep good idea

ml_peg/calcs/bulk_crystal/phonons/phonons_utils.py
ElliottKasoar: Could the current downloads benefit from this too though?

joehart2001: theyre quite differnet as we basically scrape the alexandria database for thousnands of files and jonathans one downlaods a handful of `.bz2` files so i think probably not

ElliottKasoar (dismissed):
Works great from testing!

One small thing is that the scatter plot resizes when you open the second plot, which is a bit odd, but it's a lot better than it was.

ElliottKasoar (approved):
Looks good, thanks!

PR #512, opened 2026-04-29
Add element filter

ElliottKasoar:
## Pre-review checklist for PR author

PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines](https://github.com/ddmms/ml-peg/blob/main/contributing.md).

## Summary

Adds initial work on element filter

I'm fairly happy with the general approach for the changes needed in analysis and the app, with X23 and DMC_ICE13 as current examples.

I think the filter itself works, but the full propagation isn't quite working.

I'm working on more examples e.g. for a constant element set, and fixing the propagation of scores.

## Linked issue

Resolves #255

PR #559, opened 2026-05-19
Add basic element filter

### LINKED ISSUE #255: Filter by elements

ElliottKasoar:

## Summary

This is a (somewhat) simplified version of #512, in that each test will only filter based on the superset of elements it contains, and so we will not partially update scores.

However, this still requires most of the mechanics of the full filter:

- Introduces 'mock' calculator to ensure all test outputs are generated and saved.
- Introduces (formally unused) 'error' calculator, to check that calculations complete when non-runtime errors are raised by calculations
- Updates all calculations to pass with use of error calculator i.e. they should be robust to errors
- Add `info` and `elements` attributes to app, which can store information required for element filtering
- Add `stores` property to all apps, to make computed/raw/weights/thresholds globally accessible
- Store set of original data for all tables
- Add `filter_table` function to apps, which at base just returns the same data, but is the mechanism for applying the filter
- Separate handling of `None` and `NaN` and allow `None` as valid score option.
- Combines all updates to summary (category and overall) tables to ensure race conditions do not lead to incorrect calculations when multiple triggers occur
- Adds callback to trigger table recalculation on stored updates

How this works more generally is:

- During analysis, we must save the elements that are involved in the calculation to `info.json` or similar
  - Ideally, this would be done in a way that can be used for the full filter later, but this is not necessary
  - This is still to be done for most tests
- During the app, we define a way to read the elements from `info.json`, and a `filter_table` function that uses this information, and a list of elements to be filtered, to modify the rows of the tables
  - This is still to be done for most tests, but for the simple form, this should generally be almost identical to the current examples

This is all that benchmark contributors should need to worry about. From a backend perspective, the main change is that while previously, only one table would be updated at a time, and these tables would always be on the current page, now all tables will update simultaneously, and most of these will not be rendered.

This means that we cannot rely in table.data, so must use globally accessible stored values for tables. We then also need to make sure updates to summaries are robust to multiple triggers, rather than one at a time. The current form is relatively inefficient, as it will trigger many, many times, but I think this is an ok starting point.

There are a few things left to do:

- Update all analysis to store `info.json` data
- Update all apps to convert the dict from `info.json` to a set of elements
- Add `filter_table` to all apps to apply the filter to a table
- Test the robustness of the callbacks
- Try to improve callback efficiency e.g. a single global sync, which isn't trivial to set up robustly
- Add store for filter inputs
- Add @joehart2001's element filter input via periodic table
- Add quick-select options for models

## Linked issue

Resolves #255

### DIFF
--- a/ml_peg/app/filter.py
+++ b/ml_peg/app/filter.py
@@ -91,9 +91,9 @@ def get_element_filter() -> Details:
                     Dropdown(
                         id="element-filter",
                         options=elements,
-                        value=elements,
+                        value=None,
                         multi=True,
-                        placeholder="Select elements",
+                        placeholder="Filter elements",
                         closeOnSelect=False,
                         style={"fontSize": "13px"},
                         debounce=True,
--- a/ml_peg/app/filter.py
+++ b/ml_peg/app/filter.py
@@ -1,0 +1,108 @@
+"""Build data components."""
+
+from __future__ import annotations
+
+from ase.data import chemical_symbols
+from dash.dcc import Dropdown
+from dash.html import Details, Div, Summary
+
+
+def get_model_filter(models) -> Details:
+    """
+    Get model filter component.
+
+    Parameters
+    ----------
+    models
+        List of model names to include in filter options.
+
+    Returns
+    -------
+    Details
+        Model filter component.
+    """
+    model_options = [{"label": m, "value": m} for m in models]
+
+    return Details(
+        [
+            Summary(
+                "Visible models",
+                style={
+                    "cursor": "pointer",
+                    "fontWeight": "600",
+                    "fontSize": "11px",
+                    "textTransform": "uppercase",
+                    "letterSpacing": "0.07em",
+                    "color": "#6c757d",
+                    "padding": "5px",
+                },
+            ),
+            Div(
+                [
+                    Dropdown(
+                        id="model-filter-checklist",
+                        options=model_options,
+                        value=models,
+                        multi=True,
+                        maxHeight=600,
+                        optionHeight=10,
+                        placeholder="Select visible models",
+                        closeOnSelect=False,
+                        style={"fontSize": "12px"},
+                    ),
+                ],
+                style={"padding": "8px 12px"},
+            ),
+        ],
+        id="model-filter-details",
+        open=True,
+        style={"marginBottom": "8px", "fontSize": "13px"},
+    )
+
+
+def get_element_filter() -> Details:
+    """
+    Get element filter component.
+
+    Returns
+    -------
+    Details
+        Element filter component.
+    """
+    # Exclude placeholder symbol for index 0
+    elements = chemical_symbols[1:]
+
+    return Details(
+        [
+            Summary(
+                "Filter by elements",
+                style={
+                    "cursor": "pointer",
+                    "fontWeight": "600",
+                    "fontSize": "11px",
+                    "textTransform": "uppercase",
+                    "letterSpacing": "0.07em",
+                    "color": "#6c757d",
+                    "padding": "5px",
+                },
+            ),
+            Div(
+                [
+                    Dropdown(
+                        id="element-filter-checklist",
+                        options=elements,
+                        value=elements,
+                        multi=True,
+                        placeholder="Select elements",
+                        closeOnSelect=False,
+                        style={"fontSize": "13px"},
+                        debounce=True,
+                    ),
+                ],
+                style={"padding": "8px 12px"},
+            ),
+        ],
+        id="element-filter-details",
+        open=True,
+        style={"marginBottom": "8px", "fontSize": "13px"},
+    )
--- a/ml_peg/app/filter.py
+++ b/ml_peg/app/filter.py
@@ -89,7 +89,7 @@ def get_element_filter() -> Details:
             Div(
                 [
                     Dropdown(
-                        id="element-filter-checklist",
+                        id="element-filter",
                         options=elements,
                         value=elements,
                         multi=True,
--- a/ml_peg/app/molecular_crystal/DMC_ICE13/app_DMC_ICE13.py
+++ b/ml_peg/app/molecular_crystal/DMC_ICE13/app_DMC_ICE13.py
@@ -76,13 +76,14 @@ class DMCICE13App(BaseApp):
     def get_elements(self) -> None:
         """Get element sets for filtering."""
         try:
-            self.elements = (set(entry) for entry in self.info["elements"])
-        except KeyError:
+            self.elements = [set(entry) for entry in self.info["elements"]]
+        except (AttributeError, KeyError, TypeError):
+            self.elements = []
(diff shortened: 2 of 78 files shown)


ml_peg/app/filter.py
joehart2001, line 1: maybe best to call this file `filters.py`

ml_peg/calcs/conftest.py
```diff
@@ -0,0 +1,44 @@
...
+
+    Parameters
+    ----------
+    parser
+        Pytest parser object.
+    """
+    parser.addoption(
+        "--run-mock",
+        action="store_true",
+        default=False,
```
joehart2001, line 22: why do we default this to false?

ml_peg/app/molecular_crystal/DMC_ICE13/app_DMC_ICE13.py
```diff
@@ -2,47 +2,58 @@
...
+            self.load_data()
+
         # Assets dir will be parent directory - individual files for each polymorph
-        structs_dir = DATA_PATH / MODELS[0]
+        structs_dir = DATA_PATH / "mock"
+        if not structs_dir.exists():
+            warnings.warn(f"Structures directory {structs_dir} not found", stacklevel=2)
         structs = [
-            f"/assets/molecular_crystal/DMC_ICE13/{MODELS[0]}/{struct_file.stem}.xyz"
+            f"/assets/molecular_crystal/DMC_ICE13/mock/{struct_file.stem}.xyz"
```
joehart2001, line 49: this is a good idea we should have done a while ago

ml_peg/app/molecular_crystal/DMC_ICE13/app_DMC_ICE13.py
```diff
@@ -52,6 +63,65 @@ def register_callbacks(self) -> None:
...
+    def get_elements(self) -> None:
+        """Get element sets for filtering."""
+        try:
+            self.elements = [set(entry) for entry in self.info["elements"]]
+        except (AttributeError, KeyError, TypeError):
+            self.elements = []
+            warnings.warn("Unable to read elements lists.", stacklevel=2)
+
+    @staticmethod
+    def filter_data(
```
joehart2001, line 87: we probably need to generalise this in some way. im unsure we can fully generalise it, but at least different helper functions for different types of data

ml_peg/app/molecular_crystal/DMC_ICE13/app_DMC_ICE13.py
```diff
@@ -2,47 +2,58 @@
...
 from dash import Dash
+from dash.dcc import Graph
 from dash.html import Div
+import numpy as np
 
+from ml_peg.analysis.molecular_crystal.DMC_ICE13.analyse_DMC_ICE13 import get_metrics
 from ml_peg.app import APP_ROOT
 from ml_peg.app.base_app import BaseApp
 from ml_peg.app.utils.build_callbacks import (
+    filter_table,
```
joehart2001, line 16: i think this is in `base_app.py` not build callbacks?

ml_peg/app/utils/register_callbacks.py
```diff
@@ -1090,3 +1190,87 @@ def download_table(
...
+        app_entries.append(
+            {
+                "app": app,
+                "weight_state": State(f"{app.table_id}-weight-store", "data"),
+                "threshold_state": State(f"{app.table_id}-thresholds-store", "data"),
+            }
+        )
+
+    outputs = []
+    for entry in sorted(app_entries):
```
joehart2001, line 1215: will this sorting work?

ml_peg/app/utils/register_callbacks.py
```diff
@@ -1090,3 +1190,87 @@ def download_table(
...
+        app = entry["app"]
+        outputs.extend(
+            [
+                Output(f"{app.table_id}-computed-store", "data"),
+                Output(f"{app.table_id}-raw-data-store", "data", allow_duplicate=True),
+            ]
+        )
+
+    states = []
+    for entry in app_entries:
```
joehart2001, line 1225: here we use unsorted app entries but we use sorted above. i think this will assign the incorrect weights as we dont assingn to a specific key

ml_peg/calcs/conftest.py
ElliottKasoar: I define the actual defaults later -  it's easier to default to "no flag" here and either add it or not later based on the typer CLI inputs

ml_peg/app/molecular_crystal/DMC_ICE13/app_DMC_ICE13.py
ElliottKasoar: Ish, it does also look a bit odd, but I think it's probably sensible

ml_peg/app/molecular_crystal/DMC_ICE13/app_DMC_ICE13.py
ElliottKasoar: Ah yes the DMC_ICE/X23 examples are less up to date, sorry

ml_peg/app/molecular_crystal/DMC_ICE13/app_DMC_ICE13.py
ElliottKasoar: Yes definitely! For now, I think we'll have two main general types: one where we only have one list e.g. Li diffusion, and one where we have a list of lists as the elements, and then the filter for now should basically be the same for all apps: if there's an overlap in the superset, set all metrics to `None`, which should be entirely general, but as we expand this, they'll differentiate based on data

ml_peg/app/utils/register_callbacks.py
ElliottKasoar: I believe so?

```
>>> x = {"a": 1, "c": 2, "b": 3}
>>> for k in sorted(x):
>>>    print(k)

a
b
c
```

The order doesn't matter really, just that we can do it consistently. If it's not sortable I think you get something like

```
>>> x = {str: 1, int: 2, float: 3}
>>> for k in sorted(x):
>>>    print(k)

TypeError: '<' not supported between instances of 'type' and 'type'
```

ml_peg/app/utils/register_callbacks.py
ElliottKasoar: Ah yes I think this should be sorted too, good catch

ml_peg/app/filter.py
ElliottKasoar: Yes good point, `filter` is a keyword, so a bit dangerous

ElliottKasoar:
Broken down into #571, #577, #578

PR #560, opened 2026-05-19
Add mock calculators

ElliottKasoar:

## Summary

Adds mock calculators to return nothing or an error, as discussed in #559.

When calculations run, the default is to run the mock calculator, which generates a copy of all data to use later.

The error calculator isn't used, but can be used to test #561


joehart2001 (approved):
looks good

PR #577, opened 2026-05-29
Update apps for element filtering

ElliottKasoar:

## Summary

Continuation of #559.

This introduces the app changes needed to actually apply the element filter. This needs to be generalised and added to all apps. X23/DMC_ICE are a little out of date currently, and testing won't work until the analysis is updated and re-run.

### DIFF
--- a/ml_peg/app/base_app.py
+++ b/ml_peg/app/base_app.py
@@ -144,17 +144,15 @@ class BaseApp(ABC):
         try:
             if isinstance(self.info["elements"][0], list):
                 self.elements = {
-                    [
-                        elements
-                        for sublist in self.info["elements"]
-                        for elements in sublist
-                    ]
+                    elements
+                    for sublist in self.info["elements"]
+                    for elements in sublist
                 }
             else:
                 self.elements = set(self.info["elements"])
-        except (AttributeError, KeyError, TypeError):
+        except (AttributeError, KeyError, TypeError) as err:
             self.elements = set()
-            warnings.warn("Unable to read elements lists.", stacklevel=2)
+            warnings.warn(f"Unable to read elements lists: {err}", stacklevel=2)
 
     def filter_table(self, filter_elements: list[str] | None) -> dict[str, dict]:
         """
--- a/ml_peg/app/base_app.py
+++ b/ml_peg/app/base_app.py
@@ -3,7 +3,10 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
+from copy import deepcopy
+import json
 from pathlib import Path
+import warnings
 
 from dash.dcc import Store
 from dash.development.base_component import Component
@@ -33,6 +36,8 @@ class BaseApp(ABC):
     framework_id
         Framework identifier used for benchmark attribution tags. Default is
         `"ml_peg"`.
+    info_path
+        Path to json file containing additional info for filtering. Default is None.
     """
 
     def __init__(
@@ -43,6 +48,7 @@ class BaseApp(ABC):
         extra_components: list[Component],
         docs_url: str | None = None,
         framework_id: str = "ml_peg",
+        info_path: Path | None = None,
     ):
         """
         Initiaise class.
@@ -62,6 +68,8 @@ class BaseApp(ABC):
         framework_id
             Framework identifier used for benchmark attribution tags.
             Default is `"ml_peg"`.
+        info_path
+            Path to json file containing additional info for filtering. Default is None.
         """
         self.name = name
         self.description = description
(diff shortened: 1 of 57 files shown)


joehart2001:
whats the curernt plan for this and #578 ?

ElliottKasoar:
> whats the curernt plan for this and #578?

If you ignore the changes to DMC_ICE/X23 apps (which were starting points for the full filtering), the main change each app needs is what's in `app_si_defects.py`, so actually very minimal. There should only really be two forms of `set_elements`, one where there's a single list, and one where there's a list of lists of elements.

So if you take a look at the NEB app changes + backend changes, and see what you think? I can work through the other app changes reasonably quickly.

#578 is where the bulk of the work is left, since we need to save the `info.json` that the apps read. I've basically written two forms of helper corresponding to the two forms of `set_elements`, but could use your help to go through the remaining analyses and adding these in.

It's not conceptually that difficult, but often not entirely trivial either. Most of the ones I've changed and pushed work, but a couple e.g. CPOSS209 aren't quite working yet, but the idea should be reasonably clear.

ml_peg/app/base_app.py
```diff
@@ -102,6 +139,40 @@ def register_callbacks(self):
         """Register callbacks with app."""
         pass
 
+    def filter_table(self, filter_elements: list[str] | None) -> dict[str, dict]:
```
joehart2001, line 157: i think this funciton mutating `self.table.data` is dangerous for the hosted app with multiple users. i think we should build from `deepcopy(self.original_table.data)`?

ml_peg/app/nebs/li_diffusion/app_li_diffusion.py
```diff
@@ -66,6 +65,14 @@ def register_callbacks(self) -> None:
                     mode="traj",
                 )
 
+    def set_elements(self) -> None:
+        """Get element sets for filtering."""
+        try:
+            self.elements = set(self.info["elements"][0])
```
joehart2001: i think `"elements"[0] = "L"` here not "Li"?

ml_peg/app/base_app.py
ElliottKasoar: Hmm yes I think you're right, thanks, I'll fix this soon

joehart2001: we should also be careful if theres anything else that could have a similar effect, as its less strightforward to spot when we test locally

ml_peg/app/nebs/li_diffusion/app_li_diffusion.py
ElliottKasoar: Ah this is probably a side effect of the change I made to generalise the function that rights the info for the Si defects, was wondering about that, I'll fix it, thanks!

ml_peg/app/base_app.py
ElliottKasoar: Actually I don't think we have a problem with `self.original_table.data`. We only access specific floats from it, which are immutable, so it shouldn't be changed in this process. If it was, when we applied the filter, wouldn't the metrics all be set to `None`, and so we'd be unable to unfilter?

Still looking into `self.table.data`. I get the point, but surely we ultimately do change it through callbacks anyway? We set the user specific store first, but this is still propagated to the table data itself, and that's fine?

joehart2001: the callbacks are all user-side (client DOM) so theres no issue with those, but `self.table.data` is server side so its shared. i think its an issue as  `self.table.data` is used for the initial render, so when its edited, later users will not all load the same app. if we build from teh deepcopy then we dont touch the original

joehart2001: like `self.table.data` is a python object on the server than exists once, so editing this will edit for any other users

ElliottKasoar: Oh true, sorry yep, I forgot `self.table`s aren't the same as the `DataTable`s we modify

ml_peg/app/utils/register_callbacks.py
```diff
@@ -1300,5 +1301,88 @@ def download_table(
...
+                "threshold_state": State(f"{app.table_id}-thresholds-store", "data"),
+            }
+        )
+
+    outputs = []
+    for entry in app_entries:
+        app = entry["app"]
+        outputs.extend(
+            [
+                Output(f"{app.table_id}-computed-store", "data"),
```
joehart2001, line 1331: allow duplicate needed here?

ElliottKasoar: Do you get errors about this? I don’t think I’ve seen any?

joehart2001: no errors but was just wondering

joehart2001:
@ElliottKasoar take out of draft?

ml_peg/app/utils/register_callbacks.py
ElliottKasoar: Ah cool ok, it is a bit odd... But I think it's fine?

ElliottKasoar:
> @ElliottKasoar take out of draft?

Only thing is whether we want to change the various places we use model-specific outputs, but we could do that separately, so yeah could do?

joehart2001 (approved):

PR #578, opened 2026-05-29
Update analysis to save element info

ElliottKasoar:

## Summary

Adds start initial changes to analysis as part of #559, such that the elemental information is saved.

Currently, the NEBs should work, and some of the surfaces/molecular crystals, etc.

## Tracker

  | Category | Benchmark | Done |
  |---|---|---|
  | **bulk_crystal** | elasticity | [x] |
  | **bulk_crystal** | energy_volume_curves_metals | [x] |
  | **bulk_crystal** | high_pressure_relaxation | [x] |
  | **bulk_crystal** | iron_properties | [x] |
  | **bulk_crystal** | lattice_constants | [x] |
  | **bulk_crystal** | low_dimensional_relaxation | [x] |
  | **bulk_crystal** | phonons | [x] |
  | **conformers** | 37Conf8 | [x] |
  | **conformers** | ACONFL | [x] |
  | **conformers** | DipCONFS | [x] |
  | **conformers** | Glucose205 | [x] |
  | **conformers** | Maltose222 | [x] |
  | **conformers** | MPCONF196 | [x] |
  | **conformers** | OpenFF_Tors | [x] |
  | **conformers** | solvMPCONF196 | [x] |
  | **conformers** | UpU46 | [x] |
  | **defect** | Defectstab | [x] |
  | **defect** | Relastab | [x] |
  | **electric_field** | GSCDB138_field | [x] (via shared utility) |
  | **isomers** | GSCDB138_isomers | [x] (via shared utility) |
  | **lanthanides** | isomer_complexes | [x] |
  | **molecular** | BMIMCl_RDF | [x] |
  | **molecular** | GMTKN55 | [x] |
  | **molecular** | Wiggle150 | [x] |
  | **molecular_crystal** | CPOSS209 | [x] |
  | **molecular_crystal** | DMC_ICE13 | [x] |
  | **molecular_crystal** | X23 | [x] |
  | **molecular_dynamics** | liquid_densities | [x] |
  | **molecular_dynamics** | water_density | [x] |
  | **molecular_reactions** | BH2O_36 | [x] |
  | **molecular_reactions** | BH9 | [x] |
  | **molecular_reactions** | Criegee22 | [x] |
  | **molecular_reactions** | CYCLO70 | [x] |
  | **molecular_reactions** | GSCDB138_reactions | [x] (via shared utility) |
  | **molecular_reactions** | RDB7 | [x] |
  | **nebs** | li_diffusion | [x] |
  | **nebs** | si_defects | [x] |
  | **non_covalent_interactions** | GSCDB138_NCIs | [x] (via shared utility) |
  | **non_covalent_interactions** | IONPI19 | [x] |
  | **non_covalent_interactions** | NCIA_D1200 | [x] |
  | **non_covalent_interactions** | NCIA_D442x10 | [x] |
  | **non_covalent_interactions** | NCIA_HB300SPXx10 | [x] |
  | **non_covalent_interactions** | NCIA_HB375x10 | [x] |
  | **non_covalent_interactions** | NCIA_IHB100x10 | [x] |
  | **non_covalent_interactions** | NCIA_R739x5 | [x] |
  | **non_covalent_interactions** | NCIA_SH250x10 | [x] |
  | **non_covalent_interactions** | QUID | [x] |
  | **physicality** | diatomics | [x] |
  | **physicality** | extensivity | [x] |
  | **physicality** | locality | [x] |
  | **physicality** | oxidation_states | [x] |
  | **physicality** | water_slab_dipoles | [x] |
  | **supramolecular** | LNCI16 | [x] |
  | **supramolecular** | PLA15 | [x] |
  | **supramolecular** | PLF547 | [x] |
  | **supramolecular** | S30L | [x] |
  | **surfaces** | CMRAds200 | [x] |
  | **surfaces** | elemental_slab_oxygen_adsorption | [x] |
  | **surfaces** | graphene_wetting_under_strain | [x] |
  | **surfaces** | OC157 | [x] |
  | **surfaces** | S24 | [x] |  
  | **surfaces** | SBH17 | [x] |
  | **thermochemistry** | GSCDB138_thermochemistry | [x] (via shared utility) |
  | **tm_complexes** | 3dTMV | [x] |
  | **tm_complexes** | GSCDB138_tm_complexes | [x] (via shared utility) |


ml_peg/analysis/utils/utils.py
```diff
@@ -767,3 +769,156 @@ def normalize_metric(
...
+    -------
+    dict[str, Any]
+        Dictionary of system info for all systems.
+    """
+    if out_path is None and (write_info or write_structs):
+        raise ValueError(
+            "`out_path` must be specified if `write_info` or `write_structs` is `True`."
+        )
+    info_keys = info_keys or []
+    info = {"elements": []} | dict.fromkeys(info_keys, [])
```
joehart2001:
```suggestion
    info = {"elements": []} | {key: [] for key in info_keys}
```

joehart2001: current code reuses the same list for every key, so appending to one key appends to all keys, so would share a mixed list. think the suggestion should be good as it creates a separate empty list for each key, so system_name, sys_id, etc. stay independent.

ml_peg/analysis/utils/utils.py
```diff
@@ -767,3 +769,156 @@ def normalize_metric(
 
     # Clip to [0, 1]
     return max(min(1.0, float(t)), 0.0)
+
+
+def clean_info(info: dict[str, Any] | list[Any]) -> dict[str, Any]:
```
joehart2001: return type dict but no return? use `None`?

ml_peg/analysis/utils/utils.py
```diff
@@ -767,3 +769,156 @@ def normalize_metric(
...
+    -------
+    dict[str, Any]
+        Cleaned info.
+    """
+    if isinstance(info, list):
+        for i, item in enumerate(info):
+            if isinstance(item, dict | list):
+                clean_info(item)
+            else:
+                if isinstance(item, np.int64):
```
joehart2001, line 788: do we need to worry about np.float64, np.int32, np.bool etc?

ml_peg/analysis/utils/utils.py
```diff
@@ -767,3 +769,156 @@ def normalize_metric(
...
+    elif isinstance(info, dict):
+        for key, value in info.items():
+            if isinstance(value, dict | list):
+                clean_info(value)
+
+            if isinstance(value, np.int64):
+                info[key] = int(value)
+
+
+def get_struct_info(
```
joehart2001, line 800: maybe the names of `get_struct_info` and `write_struct_info` can be improved as `get_struct_info` does also write information. also to convery one writes the union of the elemetns

ml_peg/analysis/utils/utils.py
ElliottKasoar: I did wonder this... They haven't seemed to be a problem yet, so I'm a bit hesitant to do unnecessary conversions, but there's an argument it may be safer

ml_peg/analysis/utils/utils.py
ElliottKasoar: True yep

ml_peg/analysis/utils/utils.py
ElliottKasoar: Ah yep I think originally I was going to return and realised we can just change in place

ml_peg/analysis/utils/utils.py
```diff
@@ -767,3 +769,156 @@ def normalize_metric(
...
+
+    Parameters
+    ----------
+    info
+        Dictionary of info to clean.
+
+    Returns
+    -------
+    dict[str, Any]
+        Cleaned info.
```
ElliottKasoar:
```suggestion
def clean_info(info: dict[str, Any] | list[Any]) -> None:
    """
    Ensure all data is JSON serializable.

    Parameters
    ----------
    info
        Dictionary of info to clean.
```

ml_peg/analysis/utils/utils.py
ElliottKasoar: Actually I'm less convinced in retrospect. The difference in names is because the "primary" purpose of `get_struct_info` is to read data from the files it reads and returns this as a dict, and it optionally also writes out. 

`write_struct_info` on the other hand is purely designed to write out `info.json`.

If you have other suggestions, I'm open to them, but otherwise I'd be inclined to keep the names. I did add a little to the docstring of `get_struct_info` to make its purpose clearer.

joehart2001: maybe just clearer doc strings for both functions then? and mention their use cases?

joehart2001:
side note: is it an issue that the mock calcualtor for CPOSS209 dumps all the structures into `ml_peg/app/data/molecular_crystal/CPOSS209/mock`, whilst the models have directoires for each compound: `ml_peg/app/data/molecular_crystal/CPOSS209/mace-mp-0a/acridine`

doesnt seem to effect anything visualisation wise but odd its not the same?

joehart2001:
- finished off supramolecular
- did conformers (super easy) and molecular
- fixed calc path input for some of the current ones

ElliottKasoar:
> side note: is it an issue that the mock calcualtor for CPOSS209 dumps all the structures into `ml_peg/app/data/molecular_crystal/CPOSS209/mock`, whilst the models have directoires for each compound: `ml_peg/app/data/molecular_crystal/CPOSS209/mace-mp-0a/acridine`

Ah, that's a good point. In principle there's no need for the extra layer of directories - we ultimately convert it into a list that needs to match the scatter list, and there's no reason that needs the directories.

However, if the scatter points are initially ordered based on the sort that includes the directory names, this will be a problem (but still doesn't necessarily need creation of the directories to fix).

I'd checked the structures matched, but forgot the app actually still uses the `model_dir` version, so we don't see this problem yet, but I should fix it here and/or in #577

ml_peg/analysis/supramolecular/PLA15/analyse_PLA15.py
```diff
@@ -33,50 +34,50 @@
...
+    glob_pattern="*.xyz",
+    index=0,
+    info_keys=["identifier", "interaction_type"],
+    write_info=True,
+    write_structs=True,
+    out_path=OUT_PATH,
+)
+
+
+def _get_sub_info() -> dict[str, list]:
```
ElliottKasoar: Why can we not get this via `get_struct_info`?

(@joehart2001, not sure if you're notified of these by default)

ml_peg/analysis/supramolecular/PLF547/analyse_PLF547.py
```diff
@@ -33,35 +34,36 @@
...
+    calc_path=CALC_PATH,
+    glob_pattern="*.xyz",
+    include_filenames=True,
+    write_info=True,
+    write_structs=True,
+    out_path=OUT_PATH,
+)
+
+
+def _get_charged() -> list[bool]:
```
ElliottKasoar: Why can we not get this via `get_struct_info`?

(@joehart2001, not sure if you're notified of these by default)

ml_peg/analysis/supramolecular/PLA15/analyse_PLA15.py
joehart2001: `get_struct_info` cant provide it directly as we store complex, protein and ligand for each sytem as a mutliframe xyz. `get_struct_info` flattens selected frames and gets atoms.info keys, whereas _get_sub_info preserves the per-file grouping. we could add functionality to `get_struct_info` to do this as i assume itll come up again?

joehart2001: @ElliottKasoar ive pushed some small edits to `get_struct_info`, what do you think?

joehart2001:
@ElliottKasoar did some more and added a tracker above. Currently for diatomics, it just runs all the possible diatomics a model can handle, so i think we can skip filtering for that? this will probbaly something we look at for the more advanced filtering down the line

ml_peg/analysis/utils/utils.py
ElliottKasoar: I think we'll leave it for now but maybe come back to it in future?

joehart2001 (approved):

PR #579, opened 2026-05-29
replace phonon dispersion downlaod with direct html links

joehart2001:

## Summary
Phonon dispersion plot downloads failed with HTTP 413 (request too large) for models with complex structures. The download button worked by sending the full pre-rendered PNG, SVG, and JSON back to the server on every click -> for large structures this exceeded the server's request size limit. Fixed by  replacing the server callback with direct browser download links, and updated the generic image download callback to run in the browser rather than  making a server round-trip.

## Linked issue

Resolves #

### DIFF
--- a/ml_peg/app/bulk_crystal/phonons/interactive_helpers.py
+++ b/ml_peg/app/bulk_crystal/phonons/interactive_helpers.py
@@ -12,8 +12,6 @@ from typing import Any
 
 import matplotlib
 
-from ml_peg.app.utils.build_components import build_image_download_controls
-
 matplotlib.use("Agg")
 from dash import dcc, html
 from matplotlib import gridspec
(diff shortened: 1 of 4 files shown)


ml_peg/app/bulk_crystal/phonons/interactive_helpers.py
```diff
@@ -12,8 +12,6 @@
 
 import matplotlib
 
-from ml_peg.app.utils.build_components import build_image_download_controls
```
ElliottKasoar, line 15: We should probably delete `build_image_download_controls` since it's no longer used anywhere?

ElliottKasoar (commented):
Looks great! Just one suggestion so we don't leave any unused code lying around

joehart2001:
> Looks great! Just one suggestion so we don't leave any unused code lying around

ok great, also removed `register_image_download_callbacks`

ElliottKasoar (approved):

PR #591, opened 2026-06-05
Add upet and sevennet models

### LINKED ISSUE #382: Update pet-mad name
### Summary

There are multiple pet-mad models, and they are now under the upet umbrella, so we should try to update accordingly.

### Platform

All

### Python version

_No response_

joehart2001:

## Summary
adds:
- pet-mad-1.5-s (updated r2scan model)
- pet-oam-xl
- sevennet-omni-mpa

## Linked issue

Resolves #

### DIFF
--- a/ml_peg/models/get_models.py
+++ b/ml_peg/models/get_models.py
@@ -141,7 +141,6 @@ def load_models(
         MatterSimCalc,
         MockCalc,
         OrbCalc,
-        PetMadCalc,
         SevenNetCalc,
         UPETCalc,
     )
@@ -189,16 +188,6 @@ def load_models(
                     module=cfg["module"],
                     class_name=cfg["class_name"],
                     device=cfg.get("device", "auto"),
-                    default_dtype=cfg.get("overwrite_dtype", None),
-                    kwargs=cfg.get("kwargs", {}),
-                    trained_on_dispersion=cfg.get("trained_on_dispersion", False),
-                    dispersion_kwargs=cfg.get("dispersion_kwargs", {}),
-                )
-            case "PETMADCalculator":
-                loaded_models[name] = PetMadCalc(
-                    module=cfg["module"],
-                    class_name=cfg["class_name"],
-                    device=cfg.get("device", "cpu"),
                     default_dtype=cfg.get("overwrite_dtype", None),
                     kwargs=cfg.get("kwargs", {}),
                     trained_on_dispersion=cfg.get("trained_on_dispersion", False),
--- a/ml_peg/models/get_models.py
+++ b/ml_peg/models/get_models.py
@@ -215,11 +215,12 @@ def load_models(
                     dispersion_kwargs=cfg.get("dispersion_kwargs", {}),
                 )
             case "UPETCalculator":
-                kwargs = cfg.get("kwargs", {})
                 loaded_models[name] = UPETCalc(
-                    model=kwargs["model"],
-                    version=kwargs["version"],
+                    module=cfg["module"],
+                    class_name=cfg["class_name"],
                     device=cfg.get("device", "cpu"),
+                    default_dtype=cfg.get("overwrite_dtype", None),
+                    kwargs=cfg.get("kwargs", {}),
                     trained_on_dispersion=cfg.get("trained_on_dispersion", False),
                     dispersion_kwargs=cfg.get("dispersion_kwargs", {}),
                 )
--- a/ml_peg/models/get_models.py
+++ b/ml_peg/models/get_models.py
@@ -142,6 +142,8 @@ def load_models(
         MockCalc,
         OrbCalc,
         PetMadCalc,
+        SevenNetCalc,
+        UPETCalc,
     )
 
     if run_mock is None:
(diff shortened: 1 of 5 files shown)


ml_peg/models/get_models.py
```diff
@@ -201,6 +203,25 @@ def load_models(
                     trained_on_dispersion=cfg.get("trained_on_dispersion", False),
                     dispersion_kwargs=cfg.get("dispersion_kwargs", {}),
                 )
+            case "UPETCalculator":
```
ElliottKasoar, line 206: Why do we not want to replace `PETMADCalculator`?

ml_peg/models/get_models.py
```diff
@@ -201,6 +203,25 @@ def load_models(
...
+                    device=cfg.get("device", "cpu"),
+                    default_dtype=cfg.get("overwrite_dtype", None),
+                    kwargs=cfg.get("kwargs", {}),
+                    trained_on_dispersion=cfg.get("trained_on_dispersion", False),
+                    dispersion_kwargs=cfg.get("dispersion_kwargs", {}),
+                )
+            case "SevenNetCalculator":
+                kwargs = cfg.get("kwargs", {})
+                loaded_models[name] = SevenNetCalc(
+                    model=kwargs["model"],
```
ElliottKasoar: Why do we need to pull this out when the key word is model anyway?

ml_peg/models/models.py
```diff
@@ -283,3 +284,61 @@ def get_calculator(self, **kwargs) -> Calculator:
...
+
+@dataclasses.dataclass(kw_only=True)
+class SevenNetCalc(SumCalc):
+    """Dataclass for SevenNet calculator."""
+
+    model: str
+    device: Device | None = None
+    kwargs: dict = dataclasses.field(default_factory=dict)
+
+    def get_calculator(self, **kwargs) -> Calculator:
```
ElliottKasoar, line 384: Does this work when we set precision? I don't think they've implemented anything yet (https://github.com/MDIL-SNU/SevenNet/issues/138), and then `precision` will get passed to `SevenNetCalculator`

pyproject.toml
```diff
@@ -73,6 +73,12 @@ pet-mad = [
 uma = [
     "fairchem-core == 2.19.0; python_version >= '3.11'",
 ]
+upet = [
```
ElliottKasoar, line 80: I think we just want to replace pet-mad? It's essentially the same repo

pyproject.toml
```diff
@@ -73,6 +73,12 @@ pet-mad = [
...
+    "upet",
+]
+sevenn = [
+    "sevenn",
```
ElliottKasoar: I would pin these so we know exactly what version we are using, and can rely on their interfaces

ml_peg/models/models.yml
```diff
@@ -179,6 +179,45 @@ pet-mad:
...
+sevennet-l3i5:
+  module: sevenn.sevennet_calculator
+  class_name: SevenNetCalculator
+  device: "cpu"
+  trained_on_dispersion: false
+  level_of_theory: PBE
+  kwargs:
+    model: "7net-l3i5"
+
+sevennet-omni-i12-mpa:
```
ElliottKasoar: Why this one? From their [list](https://sevennet.readthedocs.io/en/latest/user_guide/pretrained.html), the standard omni is recommended?

joehart2001: this is the one they submitted recently to matbench discovery so i assume its their model of choice. strange that one is reccomended. hmm we can discuss

ml_peg/models/get_models.py
joehart2001: can we still use the old model?

ElliottKasoar: Should be able to, yep, it's the same repo, just renamed. Models available are here: https://github.com/lab-cosmo/upet/blob/main/src/upet/_version.py#L1. You set the version e.g.

`calculator = UPETCalculator(model="pet-mad-s", version="1.5.0", device="cuda")` but v1.0.2 should also be available based on https://lab-cosmo.github.io/upet/latest/models.html

ml_peg/models/models.yml
ElliottKasoar: Hmm odd

ElliottKasoar:
Also just to note, the pre-commit it failing

ml_peg/models/get_models.py
joehart2001: updated

pyproject.toml
```diff
@@ -156,6 +159,7 @@ select = [
 
 [tool.ruff.lint.isort]
 force-sort-within-sections = true
+known-third-party = ["mlipx"]
```
joehart2001, line 169: btw this is for if you have mlipx cloned (aka me) which is why loads of my precommits were failing before

ml_peg/models/get_models.py
joehart2001: SevenNetCalc stores model as its own field, then passes remaining options through as **self.kwargs so If "model" stayed in kwargs, this call would pass it twice

joehart2001:
@ElliottKasoar udpated and addressed comments + rebased

ml_peg/models/get_models.py
ElliottKasoar: Why does `SevenNetCalc` store it separately though? It's fine, I think, just different to how the others work, unless we need to change the key word?

ml_peg/models/models.yml
```diff
@@ -188,16 +188,49 @@ orb-v3-consv-omol:
...
+  device: "cpu"
+  trained_on_dispersion: false
+  level_of_theory: r2SCAN
+  kwargs:
+    model: "pet-mad-s"
+    version: "1.5.0"
+  dispersion_kwargs:
+    xc: r2scan
+
+pet-oam-xl:
```
ElliottKasoar, line 214: Maybe we should use pet-mad-s-v[x] etc. in the same way they do here: https://github.com/lab-cosmo/upet/blob/main/src/upet/calculator.py#L84?

So add the version to `pet-mad` and `pet-oam-xl`, since `pet-oam-xl` and change the `pet-mad-1.5-s` slightly to match? Just thinking it might be confusing otherwise, especially if we add more of their model versions in future

joehart2001: i was thinking this but all the info is available when hovering in the tooltip. what do you think

ElliottKasoar: Hmm but it does become less clear when you have figures etc without the ability to do that. We can leave it but my feeling is we'd end up coming back to it down the line.

ml_peg/models/models.py
```diff
@@ -15,6 +16,48 @@
     from ase.calculators.mixing import SumCalculator
 
 
+def _patch_metatomic_nvalchemi_max_neighbors() -> None:
+    """
+    Make metatomic's CUDA neighbor-list call compatible with nvalchemi.
```
ElliottKasoar, line 21: Can you explain the context a bit more? (Not necessarily in the docstring, but for my benefit). Is it related and/or fixed by https://github.com/metatensor/metatomic/pull/259 ?

ml_peg/models/models.yml
joehart2001: so i think they call the first pet-mad just pet-mad, and the next one pet-mad-1.5 so i think names are good for now?

joehart2001:
@ElliottKasoar shoudl be all good

ElliottKasoar (approved):

PR #596, opened 2026-06-09
Return NaN for MAE of NaNs

ElliottKasoar:

## Summary

If predictions contain NaN, currently `mae` throws an error. This checks if there are NaNs (sum is among the quickest ways to do this) and if there are, returns NaN.


joehart2001 (approved):
Looks good

PR #597, opened 2026-06-09
Filtering UI + loading improvements

joehart2001:

## Summary
- add periodic table selection to element filtering
- add presets people can choose from
- improve loading to make it more obvious

## Linked issue

Resolves #

## Testing

Screenshot 2026-06-09 at 19 07 45


ml_peg/app/utils/register_callbacks.py
```diff
@@ -1276,13 +1292,16 @@ def register_filter_tables_callback(apps: dict[str, Dash]) -> None:
...
     states = []
     for entry in app_entries:
         states.extend([entry["weight_state"], entry["threshold_state"]])
 
     @callback(
-        outputs, Input("element-filter", "value"), states, prevent_initial_call=True
+        outputs,
+        Input("element-filter", "data"),
+        states,
+        prevent_initial_call=True,
```
ElliottKasoar, line 1302: This is shorter, no, so shouldn't need to be split over several lines?

ml_peg/app/utils/register_callbacks.py
```diff
@@ -1297,8 +1316,9 @@ def recompute_tables(elements, *args):
 
         Returns
         -------
-        list[list[dict]]
-            Updated rows for each app's computed store and raw data stores.
+        list
+            Updated rows for each app's computed store and raw data stores,
+            followed by a loading marker payload.
```
ElliottKasoar: What does this mean?

ml_peg/app/utils/register_callbacks.py
```diff
@@ -1329,4 +1349,5 @@ def recompute_tables(elements, *args):
 
             results.extend([scored_rows, metrics_data])
 
+        results.append({"excluded_elements": elements or []})
```
ElliottKasoar: Is this used? I can't see any other references to `excluded_elements`

joehart2001: ive added a commnet. its basically just so the loading bar hangs around until the updates are actually done

ElliottKasoar: I don't see any comments?

joehart2001: this line is actaully removed now

ml_peg/app/utils/register_callbacks.py
joehart2001: just easier to digest?

ElliottKasoar: Hmm sure

ElliottKasoar (commented):
One UI thing is that the "Clear" button seems to essentially apply immediately, but the other buttons don't, which is a bit surprising. I'd probably only apply on "Apply" as the others seem to work?

I also didn't seem to ever get the spinner when loading/filtering?

ml_peg/app/build_app.py
```diff
@@ -832,6 +836,75 @@ def build_summary_table(
...
+    ----------
+    table
+        Summary table to wrap.
+
+    Returns
+    -------
+    Div
+        Table wrapper sized to the rendered table.
+    """
+    return Div(table, style={"position": "relative", "width": "fit-content"})
```
ElliottKasoar: What does this actually do? The summary tables already don't fill the page width, so what makes this different to Div(table)?

ml_peg/app/utils/register_callbacks.py
```diff
@@ -1327,6 +1394,14 @@ def recompute_tables(elements, *args):
...
-            results.extend([scored_rows, metrics_data])
+            if _store_data_equal(scored_rows, current_scored_rows):
+                results.append(no_update)
+            else:
+                results.append(scored_rows)
+
+            if _store_data_equal(metrics_data, current_metrics_data):
+                results.append(no_update)
+            else:
+                results.append(metrics_data)
```
ElliottKasoar: Maybe worth adding a comment here, but presumably this is basically only setting the stored outputs if there's a change?

ml_peg/app/utils/register_callbacks.py
```diff
@@ -46,6 +47,54 @@
 THRESHOLD_ROUND_DIGITS = 10
 
 
+def _store_data_equal(left: Any, right: Any) -> bool:
```
ElliottKasoar: I would move this to the utils

ml_peg/app/build_app.py
```diff
@@ -832,6 +836,75 @@ def build_summary_table(
...
+            "paddingTop": "96px",
+            "boxSizing": "border-box",
+            "backgroundColor": "rgba(255, 255, 255, 0.78)",
+            "zIndex": "1400",
+            "pointerEvents": "auto",
+        },
+    )
+
+
+def _build_loading_summary_table(table: DataTable) -> Div:
```
ElliottKasoar: I don't really know what makes these private (`_func`), but also if all these do is build components, we have `build_components`

ml_peg/app/build_app.py
joehart2001: this actually prevents a maximum depth error. im adding clarifiication

joehart2001:
> One UI thing is that the "Clear" button seems to essentially apply immediately, but the other buttons don't, which is a bit surprising. I'd probably only apply on "Apply" as the others seem to work?
> 
> I also didn't seem to ever get the spinner when loading/filtering?

done

ml_peg/app/utils/onboarding.py
```diff
@@ -38,6 +38,9 @@
     },
 ]
 
+TUTORIAL_BUTTON_Z_INDEX = "1600"
+ONBOARDING_MODAL_Z_INDEX = "2000"
```
ElliottKasoar: What makes these different compared to all of the other style options?

ml_peg/app/utils/build_components.py
```diff
@@ -393,6 +393,169 @@ def build_download_controls(table_id: str, *, row: bool = False) -> Div:
...
+            "alignItems": "center",
+            "justifyContent": "center",
+            "backgroundColor": "rgba(255, 255, 255, 0.65)",
+            "zIndex": "1200",
+            "pointerEvents": "auto",
+        },
+    )
+
+
+def _build_summary_table_wrapper(table: DataTable) -> Div:
```
ElliottKasoar: It doesn't matter too much but do you think this is actually still needed? If I just do `build_filter_overlay(table.id, Div(table))` where this is currently called, I don't seem to get the max depth error, so the other fix was probably actually the solution?

ml_peg/app/utils/register_callbacks.py
```diff
@@ -1327,6 +1353,98 @@ def recompute_tables(elements, *args):
...
+                results.append(metrics_data)
+
+        # Hold the result until a minimum time has elapsed so the loading overlay
+        # stays visible long enough to read instead of flashing on fast recomputes.
+        min_loading_s = 0.15
+        remaining = min_loading_s - (time.monotonic() - start)
+        if remaining > 0:
+            time.sleep(remaining)
 
+        results.append(time.monotonic())
```
ElliottKasoar, line 1380: This isn't described in the return docstring, but is setting `filter-recompute-done`? Would that not be a bool?

ml_peg/app/utils/register_callbacks.py
```diff
@@ -1327,6 +1353,98 @@ def recompute_tables(elements, *args):
...
         return results
+
+
+def register_filter_loading_callback() -> None:
+    """
+    Drive the filter-loading overlays and disable Apply during a recompute.
```
ElliottKasoar:
```suggestion
    Drive the filter-loading overlays and disable "Apply" button during a recompute.
```
Out of context this is quite confusing otherwise

ml_peg/app/utils/register_callbacks.py
```diff
@@ -1327,6 +1353,98 @@ def recompute_tables(elements, *args):
...
+
+    Applying an element filter is the only action that commits ``element-filter``
```
ElliottKasoar:
```suggestion
    Clicking the "Apply" button is the only action that commits ``element-filter``
```

ml_peg/app/utils/register_callbacks.py
```diff
@@ -1327,6 +1353,98 @@ def recompute_tables(elements, *args):
...
+    independent callbacks on purpose: ``show`` is triggered by ``element-filter``
+    only, so Dash can run it immediately and in parallel with ``recompute_tables``
+    rather than gating it behind the recompute's output. ``hide`` is triggered by
+    the ``filter-recompute-done`` tick that ``recompute_tables`` sets when it
+    finishes, reverting overlays to ``"auto"`` (so their ``target_components``
+    cover the brief post-recompute render) and re-enabling Apply.
+
+    Pattern-matching ``ALL`` only targets overlays present in the current layout,
+    so the spinner is automatically scoped to the visible table. ``element-filter``
+    never changes on initial load, so ``show`` never latches at rest.
```
ElliottKasoar: latches at rest?

ml_peg/app/utils/register_callbacks.py
```diff
@@ -1327,6 +1353,98 @@ def recompute_tables(elements, *args):
...
+            {"type": "filter-overlay", "index": ALL}, "display", allow_duplicate=True
+        ),
+        Output("element-filter-apply", "disabled", allow_duplicate=True),
+        Input("element-filter", "data"),
+        State({"type": "filter-overlay", "index": ALL}, "id"),
+        prevent_initial_call=True,
+    )
+    def show_filter_loading(_filter_data, overlay_ids):
+        """
+        Show overlays and disable Apply when a filter is committed.
```
ElliottKasoar:
```suggestion
        Show overlays and disable "Apply" button when a filter is committed.
```

ml_peg/app/utils/build_components.py
joehart2001: just tested and its not need now

ml_peg/app/utils/onboarding.py
joehart2001: updated

ml_peg/app/utils/register_callbacks.py
joehart2001: added comments

ml_peg/app/utils/register_callbacks.py
joehart2001: updated

joehart2001:
@ElliottKasoar should be good to go

joehart2001:
accidentally pressed close instead of comment haha

ElliottKasoar (approved):
Looks great, thanks for this!

PR #600, opened 2026-06-10
Fix filter

ElliottKasoar:

## Summary

Fixes filtering so scores are calculated based on remaining values, rather than getting set to None/NaN.

While the `None` values we write out when we store the initial tables are preserved in the `table.json` files, when the same data is written to json for the `dcc.Store`, both `NaN` and `None` are set to `null`, and so are read back in as `None`.

I therefore use `"NaN"` are a `NaN`, and reserve `None` for filtered data. If a value is `NaN`, is blocks calculation of the total score, while if a value is `None`, it does not, and I added a new style to distinguish these cases.

I also remove the distinction between `None` and `NaN` in the initial tables, as these are all unfiltered, is are `"NaN"`s in this scheme.


joehart2001 (approved):
think this is all good

PR #606, opened 2026-06-12
Add Vivace model

joehart2001:

## Summary
add Vivace model

to run it, you do need to downalod the checkpoint and add the path

## Linked issue

Resolves #

## Progress

- [x] Model configuration
- [ ] Documentation

### DIFF
--- a/ml_peg/models/get_models.py
+++ b/ml_peg/models/get_models.py
@@ -143,6 +143,7 @@ def load_models(
         OrbCalc,
         SevenNetCalc,
         UPETCalc,
+        VivaceCalc,
     )
 
     if run_mock is None:
@@ -216,6 +217,13 @@ def load_models(
             case "SevenNetCalculator":
                 loaded_models[name] = SevenNetCalc(
                     device=cfg.get("device", "cpu"),
+                    kwargs=cfg.get("kwargs", {}),
+                    trained_on_dispersion=cfg.get("trained_on_dispersion", False),
+                    dispersion_kwargs=cfg.get("dispersion_kwargs", {}),
+                )
+            case "MLFFCalculator":
+                loaded_models[name] = VivaceCalc(
+                    device=cfg.get("device", "auto"),
                     kwargs=cfg.get("kwargs", {}),
                     trained_on_dispersion=cfg.get("trained_on_dispersion", False),
                     dispersion_kwargs=cfg.get("dispersion_kwargs", {}),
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -80,6 +80,12 @@ pet-mad = [
 uma = [
     "fairchem-core == 2.19.0; python_version >= '3.11'",
 ]
+vivace = [
+    "simpoly[vivace,cuda13] @ git+https://github.com/microsoft/simpoly.git@da100ad32d971214cf9343d9d853e8ce61604dc0; sys_platform == 'linux'",
+    "simpoly[vivace] @ git+https://github.com/microsoft/simpoly.git@da100ad32d971214cf9343d9d853e8ce61604dc0; sys_platform != 'linux'",
```
ElliottKasoar, line 88: This is probably ok for now, but do you know if there are plans for this to be released, or at least tagged?

Is CUDA/cuequivariance absolutely required?

ml_peg/models/get_models.py
```diff
@@ -212,6 +213,17 @@ def load_models(
                     trained_on_dispersion=cfg.get("trained_on_dispersion", False),
                     dispersion_kwargs=cfg.get("dispersion_kwargs", {}),
                 )
+            case "MLFFCalculator":
+                kwargs = cfg.get("kwargs", {})
+                loaded_models[name] = VivaceCalc(
+                    model_path=kwargs["model_path"],
```
ElliottKasoar: I think same question as for #591, we don't necessarily need to pull out `model_path` from kwargs or set `model_path: str` as one of the TypedDict inputs?

(If we did have it as a key in the TypedDicts, I think the reason would be to normalise it in a way to expose the model name more consistently, but this doesn't do that))

ElliottKasoar (commented):
Thanks @joehart2001! Can you rebase and fix the pre-commit?

pyproject.toml
joehart2001: yep this is how they install it and was a little fiddly. unsure about the release, we can update if needed, immediately sure gregor will let us know

joehart2001:
@ElliottKasoar rebased and updated, is ok?

ElliottKasoar (approved):
I'm assuming there isn't a built-in way to download any models? I can't test this fully, but it looks ok

joehart2001:
> I'm assuming there isn't a built-in way to download any models? I can't test this fully, but it looks ok

yep doesnt seem like it

PR #617, opened 2026-06-16
In the summary table add link to model docs + add model docs

joehart2001:

## Summary
- link is added as an invisible column
- for the model docs, ive just pasted the models.yml in for now so we can expand later

## Linked issue

Resolves #

## Testing

Screenshot 2026-06-16 at 22 12 14

### DIFF
--- a/ml_peg/app/run_app.py
+++ b/ml_peg/app/run_app.py
@@ -34,35 +34,6 @@ app = Dash(
     update_title=None,  # prevent the tab changing to Updating... during callbacks
 )
 
-# Compact, underline-free styling for the summary-table model docs-link column.
-# Injected here because the assets folder (DATA_PATH) is not version controlled.
-app.index_string = """<!DOCTYPE html>
-<html>
-    <head>
-        {%metas%}
-        <title>{%title%}</title>
-        {%favicon%}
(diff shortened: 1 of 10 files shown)


ml_peg/app/run_app.py
```diff
@@ -34,6 +34,35 @@ def _build_full_app(app: Dash, category: str):
     update_title=None,  # prevent the tab changing to Updating... during callbacks
 )
 
+# Compact, underline-free styling for the summary-table model docs-link column.
+# Injected here because the assets folder (DATA_PATH) is not version controlled.
```
ElliottKasoar: I don't entirely understand these comments? What does this have to do with the assets folder? We do have some assets version controlled (intro tutorial stuff), don't we?

If it does need to be set here, I'd also probably import it. We've lost this a little as things have grown, but roughly the idea is to separate logic from the display settings etc. (see: model-view-controller design patterns).

joehart2001: sorry the comment doesnt make sense out of context as i was trying to make it remove a blue underline for hte link. and yep i agree, ive put in a `.css` file now and updated the comments

ElliottKasoar (approved):

PR #624, opened 2026-06-18
Add test option for ml_peg app

ElliottKasoar:

## Summary

Add option to run individual benchmark apps.

With the changes for filtering, the logic for updating category summary tables, as well as the new logic for filtering, are done as part of the overall `run_app`, so do not work if individual apps are launched directly.

We should probably remove the `if __name__ == "__main__":` sections from apps to discourage this, but for now, this enables the possibility of launching them individually with full functionality.


joehart2001 (approved):
nice!

Issue #623, opened 2026-06-18
Add partial filtering


ElliottKasoar:
### Summary

Filtering currently works in a basic form, where if any element in a test is selected, the whole test is discarded.

In many cases, it would be much more preferable to exclude specific data points from the metric calculated instead. This requires a few additional changes:

- Analysis most likely should be rewritten in a form where the metric dictionary can be returned as a function callable outside of `pytest` (so so reducing dependencies on filters)
- Applications will need custom filter functions, which will require processing of data to pass to the metric getter function. For example, this could be a masked list of errors from data taken from the app's scatter plot.

This could become quite expensive to run.

PR #625, opened 2026-06-18
Add example partial filter

ElliottKasoar:

## Summary

This shows an example implementation of partial filtering. In order to avoid redefining analysis logic, ideally we import and call functions from analysis during the app partial filter.

This means we need to make some changes to make this function callable (not a fixture, remove data-dependent module constants), and we then need to pass it suitable data, which we can get from the scatter plot in this case.

The changes are otherwise relatively intuitive, and is implemented similarly to initial work in #512.

The other question (which is why this is in draft) is whether it makes sense to merge these one at a time, or if we ought to have them all complete before activating this.

## Linked issue

Partially addresses #623.

### DIFF
--- a/ml_peg/app/surfaces/elemental_slab_oxygen_adsorption/app_elemental_slab_oxygen_adsorption.py
+++ b/ml_peg/app/surfaces/elemental_slab_oxygen_adsorption/app_elemental_slab_oxygen_adsorption.py
@@ -2,9 +2,16 @@
 
 from __future__ import annotations
 
+from copy import deepcopy
+import warnings
+
 from dash import Dash
 from dash.html import Div
+import numpy as np
 
+from ml_peg.analysis.surfaces.elemental_slab_oxygen_adsorption.analyse_elemental_slab_oxygen_adsorption import (  # noqa: E501
+    get_metrics,
+)
 from ml_peg.app import APP_ROOT
 from ml_peg.app.base_app import BaseApp
 from ml_peg.app.utils.build_callbacks import (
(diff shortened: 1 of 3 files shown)


ElliottKasoar:
I think we suggested in discussion that we merge this as an example if people want to start doing this sort of thing (assuming you're happy with and can test it), @joehart2001

ml_peg/app/surfaces/elemental_slab_oxygen_adsorption/app_elemental_slab_oxygen_adsorption.py
```diff
@@ -2,21 +2,24 @@
...
 
+from copy import deepcopy
+import warnings
+
 from dash import Dash
 from dash.html import Div
+import numpy as np
 
+from ml_peg.analysis.surfaces.elemental_slab_oxygen_adsorption.analyse_elemental_slab_oxygen_adsorption import (  # noqa: E501
+    get_metrics,
```
joehart2001, line 13: i think this import triggers the analysis module so if we dont have the mock outputs present then the app will fail to restart:
```
ValueError: 
/Users/joehart/Desktop/0_Cambridge/0_MPhil_Scientific_Computing/MPhil_project/mlipx_testing/new_format/mlip-testing/ml_peg/calcs/surf
aces/elemental_slab_oxygen_adsorption/outputs/mock does not exist. Please run mock calculation.
```

ml_peg/app/surfaces/elemental_slab_oxygen_adsorption/app_elemental_slab_oxygen_adsorption.py
```diff
@@ -54,6 +61,84 @@ def register_callbacks(self) -> None:
...
+            for current_row, original_row in zip(
+                table_data, self.original_table.data, strict=True
+            ):
+                for metric in self.metrics:
+                    current_row[metric] = original_row[metric]
+            return table_data
+
+        # Partial filtering
+        # Get overlap of deselected elements with each system's elements
+        filtered_indices = [
```
joehart2001, line 118: here if we choose to remove O then we will get an empty energy array and then get NaNs later `results[model_name] = np.nan`. do we want None or NaN?

PR #627, opened 2026-06-18
Add clear cache button and automatically clear on version bump

joehart2001:

## Linked issue

Resolves #

### DIFF
--- a/ml_peg/app/build_app.py
+++ b/ml_peg/app/build_app.py
@@ -17,10 +17,9 @@ from dash import (
 from dash.dash_table import DataTable
 from dash.dcc import Dropdown, Interval, Link, Loading, Location, Store
 from dash.exceptions import PreventUpdate
-from dash.html import H1, H3, A, Br, Button, Details, Div, Img, Span, Summary
+from dash.html import H1, H3, A, Br, Details, Div, Img, Span, Summary
 from yaml import safe_load
 
-from ml_peg import __version__
 from ml_peg.analysis.utils.utils import calc_table_scores, get_table_style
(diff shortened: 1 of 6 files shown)


ElliottKasoar (commented):
Nice!

What's the benefit of client-side callbacks vs some sort of pattern-matching explicit clear as a Python callback?

I also think there may be a bug in the storage, since the weights etc. don't seem to be persisted once I close the tab. I'm not sure that's introduced here, but it makes it harder to test the version clear since I think it happens anyway.

I'd probably also make the button more prominent, perhaps next to the tutorial button? It's quite out of the way currently.

joehart2001:
> Nice!
> 
> What's the benefit of client-side callbacks vs some sort of pattern-matching explicit clear as a Python callback?
> 
> I also think there may be a bug in the storage, since the weights etc. don't seem to be persisted once I close the tab. I'm not sure that's introduced here, but it makes it harder to test the version clear since I think it happens anyway.
> 
> I'd probably also make the button more prominent, perhaps next to the tutorial button? It's quite out of the way currently.

- yep the bug is on the live site too, i think we need to move the wieghts and threhsolds to local storage?
- changed the location
- the clear is a browser operation:  localStorage.clear(), sessionStorage.clear(), and window.location.reload() which a python callback cant do
- added a progress for the loading

ElliottKasoar (commented):
Similarly to #699, I think ideally this wouldn't live in ml_peg/app/build_app.py, but otherwise this is nice, thanks!

ElliottKasoar (approved):
Looks good, thanks!

PR #628, opened 2026-06-18
Improve element filtering: model union/intersection presets, kept count, apply indicator

joehart2001:

## Summary
- add model union and intersection for selected models. this is done by adding "dataset" parameter to the `models.yml`, and adding the dataset to the json
- added an element counter
- add a indicator when you have unapplied changes

- TODO: json needs to be updated for all datasets, same with modles.yml, but code is ready for review

## Linked issue

Resolves #


ml_peg/models/models.yml
```diff
@@ -60,6 +64,7 @@ mace-matpes-r2scan:
 orb-v3-consv-inf-omat:
   module: orb_models.inference.calculator
   class_name: OrbCalc
+  datasets: [OMAT]
```
joehart2001, line 69: unsure if this is totally correct as orb supports more elemtns than omat covers

ElliottKasoar: Can we add a way to explicitly add extra elements?

ml_peg/app/data/element_coverage.json
```diff
@@ -1,6 +1,192 @@
 {
   "datasets": {
-    "MPtrj/Alexandria/OMAT": {
+    "MPtrj": {
+      "supported": [
```
ElliottKasoar, line 5: Why do we need the extra nesting from `supported`?

ElliottKasoar:
> TODO: update model docs to tell people to add element converge

We can open this as a separate issue, I'm happy to merge if we address everything else, especially since it's not that common that we have new models added at the moment.

joehart2001:
@ElliottKasoar lets merge vivace sevennet etc first then will update

ml_peg/models/element_coverage/find_supported_elements.py
```diff
@@ -0,0 +1,159 @@
...
+from ase.data import atomic_numbers, covalent_radii
+import numpy as np
+
+from ml_peg.analysis.utils.periodic_table import PERIODIC_TABLE_SYMBOLS
+from ml_peg.models.get_models import get_model_names, load_models
+
+OUT_PATH = Path(__file__).parent / "model_supported_elements.json"
+ERR_PATH = Path(__file__).parent / "model_supported_elements_errors.jsonl"
+
+_BOX = 20.0
```
ElliottKasoar, line 39: Why is this `_BOX` not just `BOX`?

ml_peg/models/element_coverage/find_supported_elements.py
```diff
@@ -0,0 +1,159 @@
...
+    -------
+    list[str] | None
+        Sorted supported element symbols, or ``None`` if the model could not be
+        loaded in the current environment.
+    """
+    try:
+        model = load_models(model_name)[model_name]
+        if hasattr(model, "device"):
+            model.device = device
+        calc = model.get_calculator()
```
ElliottKasoar:
```suggestion
        calc = model.get_calculator(precision="low")
```
Speed things up a bit? (Default is "high" but that shouldn't matter here?)

ElliottKasoar (commented):
Maybe we should add the output from the element_covered scripts, and re-run it when as part of adding new models?

joehart2001:
> Maybe we should add the output from the element_covered scripts, and re-run it when as part of adding new models?

basically it finds what datasets overlap but it doesnt actaully know which datasets the model is trained on so i think safer manually

ElliottKasoar:
> @ElliottKasoar lets merge vivace sevennet etc first then will update

@joehart2001 these are all in now

ml_peg/models/element_coverage/find_supported_elements.py
ElliottKasoar: @joehart2001?

joehart2001:
@ElliottKasoar all done

ElliottKasoar (approved):
This is great, thanks!

PR #644, opened 2026-06-24
feat: add md stability benchmark

### LINKED ISSUE #640: feat: new benchmark testing molecular dynamics stability of organic molecules and peptides
### Summary

This benchmark tests if an MLIP can run stable simulations of various organic small molecules and peptides (in vacuum and in solution), as well as of a protein in vacuum. For more information see the [MLIP audit docs](https://instadeepai.github.io/mlipaudit/benchmarks/general/stability.html).

### Interactive features

For now just a table with MAE.

### Category

Molecular Dynamics

### Data availability

Data is published on MLIP Audit HuggingFace.

### Computational cost

High: 8 molecular dynamics simulations of 100 ps (100.000 inference calls) with system sizes up to 2800 atoms.

lwehrhan:

## Summary

This benchmark tests if an MLIP can run stable simulations of various organic small molecules and peptides (in vacuum and in solution), as well as of a protein in vacuum.

## Linked issue

Resolves #640 

## Progress

- [x] Calculations
- [x] Analysis
- [x] Application
- [x] Documentation


joehart2001:
Hi @lwehrhan, from a brief look this is looking nice and i think wont need many comments from us. Could you please take a look at the element filtering mentioned by @ElliottKasoar in #429, it should be pretty straightforward to add. then we will take a deeper look over. thanks!

ml_peg/calcs/molecular_dynamics/stability/calc_stability.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    Benchmark molecular dynamics stability.
+
+    Parameters
+    ----------
+    mlip
+        Name of model and model object to get calculator.
+    """
+    model_name, model = mlip
+    calc = model.get_calculator()
```
ElliottKasoar:
```suggestion
    calc = model.get_calculator(precision="low")
```
Possibly? Generally we'd use a lower precision ("float32" for most models) for MD, but either way, it would be good to be explicit here (the default is `"high"`, but it's nice to make the intent clear).

ElliottKasoar (commented):
Does the benchmark save anything in terms of the trajectories? It might be interesting to see, for example, how far a simulation gets before exploding, even if it doesn't form a part of the score.

Either through that, or otherwise, it might also be nice to be able to see which systems were successfully simulated or not. Do you think this might be possible?

joehart2001:
Thanks for the pr! same as #658 could you send the data + have a look at the element filtering?

docs: https://ddmms.github.io/ml-peg/developer_guide/filter.html

lwalew:
> Does the benchmark save anything in terms of the trajectories? It might be interesting to see, for example, how far a simulation gets before exploding, even if it doesn't form a part of the score.
> 
> Either through that, or otherwise, it might also be nice to be able to see which systems were successfully simulated or not. Do you think this might be possible?

Hi @ElliottKasoar, the benchmark should save the trajectories as part of the benchmarks's `ModelOutput`. I also believe @lwehrhan added a scatter plot under the name `Trajectory Stability` that should visualise how far a simulation went before exploding. And exploded simulations also do count towards the score.

ElliottKasoar:
Just to note, this has a few minor conflicts now to be resolved

mcarobene-id:
Hi @ElliottKasoar , have updated to `ddmms:main` and resolved conflicts now - everything else should be as reviewed before :)

ml_peg/analysis/molecular_dynamics/stability/analyse_stability.py
```diff
@@ -0,0 +1,221 @@
...
+    check_dataset()
+
+    results = {}
+    for model_name in MODELS:
+        output_dir = CALC_PATH / model_name / MlPegStabilityBenchmark.name
+        if not (output_dir / "model_output.zip").exists():
+            continue
+        benchmark = MlPegStabilityBenchmark(
+            force_field=Calculator(),
+            data_input_dir=CALC_PATH,
```
joehart2001: Analysis looks for the dataset under outputs/stability, but the calc never copies it there (its in the cache)

joehart2001:
> Hi @ElliottKasoar , have updated to `ddmms:main` and resolved conflicts now - everything else should be as reviewed before :)

Hey @mcarobene-id thanks, this is looking nice. i have tested with some of the data from the google drive and seems all good. @ElliottKasoar can you take a look when youve got time

joehart2001:
also could you mark the calc as very slow with `@pytest.mark.very_slow`

PR #652, opened 2026-06-29
Add multihead and polar frameworks

joehart2001:

## Summary
- add the frameworks for multihead and polar papers
- marked tests with mulithead
- @ElliottKasoar would you be able to mark the polar ones if you have time?

## Linked issue

Resolves #

### DIFF
--- a/docs/source/developer_guide/add_benchmarks.rst
+++ b/docs/source/developer_guide/add_benchmarks.rst
@@ -529,19 +529,28 @@ MLIP Arena), add a framework credit tag as follows:
       url: "https://huggingface.co/spaces/atomind/mlip-arena"
       logo: "https://huggingface.co/front/assets/huggingface_logo-noborder.svg"
 
-2. Set ``framework_id`` in the benchmark app constructor.
+2. Set ``framework_ids`` in the benchmark app constructor. This accepts either a
+   single framework identifier or a list of them, so a benchmark can belong to
+   multiple frameworks at once.
 
 .. code-block:: python3
 
     return SomeBenchmarkApp(
         name="SomeBenchmark",
         ...,
-        framework_id="mlip_arena",
+        framework_ids="mlip_arena",
     )
 
-That is all that is required. The benchmark header badge and the additional
-framework pages for non-default frameworks are populated automatically from
-this metadata.
+    # ...or, to credit several frameworks:
+    return SomeBenchmarkApp(
+        name="SomeBenchmark",
+        ...,
+        framework_ids=["mlip_arena", "multihead"],
+    )
+
+That is all that is required. The benchmark header shows one badge per framework,
+and the additional framework pages for non-default frameworks are populated
+automatically from this metadata.
 
 Framework sections group matching benchmarks by category, omit the category
 summary table, and reuse the same benchmark tables and controls. Updating
--- a/docs/source/developer_guide/add_benchmarks.rst
+++ b/docs/source/developer_guide/add_benchmarks.rst
@@ -512,40 +512,59 @@ but less standard plots/interactions will need setting up.
 
 For now, please contact us to help with this process.
 
-Framework credit tags
-+++++++++++++++++++++
+Framework credit and custom group tags
+++++++++++++++++++++++++++++++++++++++
 
-If a benchmark comes from an external benchmarking framework (for example,
-MLIP Arena), add a framework credit tag as follows:
+If a benchmark comes from an external benchmarking framework (for example, MLIP Audit),
+or you want to group benchmarks by a custom category (such as those involved in a
+paper), you can add a framework credit tag as follows:
 
 1. Add/update the framework entry in ``ml_peg/app/utils/frameworks.yml``.
 
 .. code-block:: yaml
 
-    mlip_arena:
-    label: MLIP Arena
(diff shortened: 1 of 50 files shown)


ElliottKasoar:
Do we not need to make it possible for multiple frameworks to be associated with a benchmark before/as part of this?

joehart2001:
> Do we not need to make it possible for multiple frameworks to be associated with a benchmark before/as part of this?

yep added now and tested on S30L

ml_peg/app/bulk_crystal/elasticity/app_elasticity.py
```diff
@@ -119,6 +119,7 @@ def get_app() -> ElasticityApp:
     """
     return ElasticityApp(
         name=BENCHMARK_NAME,
+        framework_ids="multihead",
```
ElliottKasoar: Surely we want multihead and ML-PEG (for this and all of the others)?

joehart2001: i feel like its a given? undecided

ElliottKasoar: I'd say it's better not to be ambiguous? If we develop a new model and want to use the full set of tests as part of that development (and so in a paper), it would include contributed benchmarks too?

If we're labelling some tests as ML-PEG when I think it's probably clearest to do all of them, rather than people having to assume.

joehart2001: ok yep i added it. i also added a flag where you can opt out of using it incase thats of use? true by default anyway

docs/source/developer_guide/add_benchmarks.rst
```diff
@@ -529,19 +529,28 @@ MLIP Arena), add a framework credit tag as follows:
...
+   single framework identifier or a list of them, so a benchmark can belong to
+   multiple frameworks at once.
 
 .. code-block:: python3
 
     return SomeBenchmarkApp(
         name="SomeBenchmark",
         ...,
-        framework_id="mlip_arena",
+        framework_ids="mlip_arena",
```
ElliottKasoar: Should probably not have `include_ml_peg` True/comment on this too?

ml_peg/app/base_app.py
```diff
@@ -33,9 +34,13 @@ class BaseApp(ABC):
         List of other Dash components to add to app.
     docs_url
         URL for online documentation. Default is None.
-    framework_id
-        Framework identifier used for benchmark attribution tags. Default is
-        `"ml_peg"`.
+    framework_ids
+        Extra framework identifiers used for benchmark attribution tags, as a
+        single string or a list of strings. The `"ml_peg"` tag is always added,
```
ElliottKasoar: If include_ml_peg is True ?

ml_peg/app/base_app.py
```diff
@@ -65,9 +71,13 @@ def __init__(
             List of other Dash components to add to app.
         docs_url
             URL to online documentation. Default is None.
-        framework_id
-            Framework identifier used for benchmark attribution tags.
-            Default is `"ml_peg"`.
+        framework_ids
+            Extra framework identifiers used for benchmark attribution tags, as a
+            single string or a list of strings. The `"ml_peg"` tag is always
```
ElliottKasoar: If include_ml_peg is True ?

ml_peg/app/supramolecular/S30L/app_S30L.py
```diff
@@ -71,6 +71,7 @@ def get_app() -> S30LApp:
     """
     return S30LApp(
         name=BENCHMARK_NAME,
+        framework_ids=["multihead", "mace-polar-1"],
```
ElliottKasoar: Do you think we should maybe do `mace-mh`/`mace-multihead` to be more consistent between these two? I can change these if you think so

ElliottKasoar:
I've added the remaining polar tags. I think we want to update the docs slightly, but otherwise think it's good to go? Would be nice to merge before my talk tomorrow!

ml_peg/app/base_app.py
joehart2001: what do you mean by this?

joehart2001:
> I've added the remaining polar tags. I think we want to update the docs slightly, but otherwise think it's good to go? Would be nice to merge before my talk tomorrow!

nice! yeah i agree

ml_peg/app/base_app.py
ElliottKasoar: The `"ml_peg"` tag is not **always** added, it's added if `include_ml_peg` is also True

ml_peg/app/base_app.py
ElliottKasoar: See below

ml_peg/app/base_app.py
ElliottKasoar: E.g. this comment doesn't make sense if you read it in the context of setting `framework_ids="mlip_arena"`

joehart2001:
@ElliottKasoar i agree with your suggestions, would you be able to make the changes and i review them? (some are slightly ambiguous so i feel like its better than more back and fourth)

ElliottKasoar (approved):

PR #658, opened 2026-07-02
feat: add tautobase

lwehrhan:
## Pre-review checklist for PR author

PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines](https://github.com/ddmms/ml-peg/blob/main/contributing.md).

## Summary

The tautobase is a dataset of 1,391 tautomer pairs of organic small molecules. The geometries were refined at xtb level of theory. QM single point energies (reference for this benchmark) were re-computed by InstaDeep in-house at ωB97M-D3(BJ)/def2-TZVPPD level of theory.

## Linked issue

Resolves #641 

## Progress

- [x] Calculations
- [x] Analysis
- [x] Application
- [x] Documentation

PR #666, opened 2026-07-03
feat: add bond length distribution benchmark

### LINKED ISSUE #671: New benchmark Bond Length Distribution
### Summary

Summary
This benchmark tests the ability of an MLIP to maintain physically reasonable covalent bond lengths during molecular dynamics. For each of 8 small organic molecules — covering carbon-carbon (single, double, triple), carbon-oxygen (single, double), carbon-nitrogen (single, triple) and carbon-fluorine bonds — a molecular dynamics simulation is run starting from a reference structure, and the deviation of the tracked bond from its reference equilibrium length is measured over the trajectory. A well behaved potential keeps bonds close to their reference length, so a lower average deviation is better.

Interactive features
Histogram of the bond length deviations sampled along the trajectories, per model.

Category
Molecular Dynamics

Data availability
Selection of sampled structures on MLIP Audit HuggingFace. Selected from QM9 comprised of bonds C-C, C=C, C#C, C-N, C-O, C=O and C-F. 

Computational cost
One MD simulation per molecule (8 molecules), each 1,000,000 steps.

### Interactive features

_No response_

### Category

_No response_

### Data availability

_No response_

### Computational cost

_No response_

### Additional context

_No response_

lwalew:

## Summary

Migrates the bond length distribution benchmark from the MLIP Audit suite. For each of a set of small organic molecules — covering the C-C, C=C, C#C, C-N, C-O, C=O and C-F bond types — an NVT molecular dynamics simulation is run at 300 K starting from a QM-optimised reference geometry (selected from QM9), and the deviation of a tracked covalent bond from its reference length is measured over the trajectory. A well behaved potential keeps bonds close to their reference length, so a lower average deviation is better.

The analysis reports the average bond length deviation per model in the metrics table and shows the distribution of sampled deviations as a histogram.

## Linked issue

Resolves #671

## Progress

- [x] Calculations
- [x] Analysis
- [x] Application
- [x] Documentation

## Testing

Not yet run end-to-end: the benchmark input data still needs to be uploaded to the ML-PEG S3 bucket at `inputs/molecular_dynamics/bond_length_distribution/bond_length_distribution.zip` (layout `bond_length_distribution/bond_length_distribution.json`). Kept as a draft until the data is available.

Verified locally in the meantime:

- `pre-commit` passes on all changed files (`ruff check`, `ruff format`, `numpydoc-validation`).
- Both test modules are collected by pytest with `mlipaudit` installed, and skip cleanly with `Please install mlipaudit extra` when it is not, rather than erroring during collection.
- The element info written to `info.json` was checked against the dataset: 8 molecules, and the union of the per-molecule element lists is `['C', 'F', 'H', 'N', 'O']`, matching `BondLengthDistributionBenchmark.required_elements`.
- The calculation → analysis hand-off was exercised end-to-end with a local copy of the dataset, confirming the analysis reads the dataset from the calculation outputs without downloading it again.

## New decorators/callbacks

No new callbacks required; uses the existing `plot_hist` and `build_table` decorators and the `plot_from_table_column` callback.

## Note for reviewers

Addressing the review feedback here and the equivalent points on #658:

- **Element filtering.** Follows the Folmsbee template, with elements saved as one list per molecule rather than a single union, so partial filtering can exclude individual molecules later. Ordering follows the dataset, which matches the molecule order in `analyze()`'s results. This matters here: only 1 of the 8 molecules contains F, so under a single union list, filtering out F would exclude the whole benchmark rather than one molecule.
- **Reusing the calculation output.** The calculation now copies the downloaded dataset into its outputs, and the analysis reads it from there with `data_input_dir=CALC_PATH`, so it no longer re-downloads from S3 (previously twice per run). A missing dataset raises a clear "Please run the calculation" error.
- **`pytest.importorskip`.** Added to both the calculation and analysis modules, positioned ahead of the `ml_peg.calcs.utils.mlipaudit` import, since that module imports `mlipaudit` unconditionally.
- **GPU cost estimate.** Added to the docs, per the comment thread below.
- Unlike Folmsbee and tautomers, this benchmark does not use the mock output as its source of truth. Those need it to order parity-plot hoverdata labels; the histogram here has no per-molecule labels, so the dataset in the calculation outputs is sufficient.
- A model whose simulations all fail is reported as `NaN` rather than `None`.

### DIFF
--- a/docs/source/user_guide/benchmarks/molecular_dynamics.rst
+++ b/docs/source/user_guide/benchmarks/molecular_dynamics.rst
@@ -116,3 +116,46 @@ Packmol generated
 Reference data:
 * M. Southard and D. Green, Perry’s Chemical Engineers’ Handbook, 9th Edition. McGraw-Hill Education, 2018.
 * Experimental
+
+
+Bond length distribution
+========================
+
+Summary
+-------
+
+Performance in maintaining physically reasonable covalent bond lengths during molecular
+dynamics of small organic molecules. For each of a set of molecules covering the C-C, C=C,
+C#C, C-N, C-O, C=O and C-F bond types, an NVT molecular dynamics simulation is run at 300 K
+starting from a QM-optimised reference geometry, and the deviation of a tracked bond from
+its reference length is measured along the trajectory.
+
+Metrics
+-------
+
+1. Bond length deviation
+
+The length of the tracked bond is measured at each frame of the trajectory, and its absolute
+deviation from the reference bond length is averaged over the trajectory and across all
+molecules. A well behaved potential keeps bonds close to their reference length, so a lower
+deviation is better.
+
+A histogram shows the distribution of the sampled bond length deviations for each model.
+
+Computational cost
+------------------
+
+High: one MD simulation per molecule, each 1,000,000 steps. Faster inference can be achieved
+using the jax-accelerated simulations in MLIP Audit directly.
+
+Data availability
+-----------------
+
+Input structures:
+
+* MLIP Audit benchmark suite, InstaDeep. Reference geometries selected from the QM9 dataset
+  (Ramakrishnan et al., Scientific Data 1, 140022, 2014).
+
+Reference data:
+
+* QM-optimised equilibrium bond lengths of the reference geometries.
--- a/docs/source/user_guide/benchmarks/molecular_dynamics.rst
+++ b/docs/source/user_guide/benchmarks/molecular_dynamics.rst
@@ -145,8 +145,11 @@ A histogram shows the distribution of the sampled bond length deviations for eac
 Computational cost
 ------------------
 
-High: one MD simulation per molecule, each 1,000,000 steps. Faster inference can be achieved
-using the jax-accelerated simulations in MLIP Audit directly.
+High: 8 molecules of 4-13 atoms, one MD simulation each, of 1,000,000 steps, i.e. 1 ns at a
+1 fs timestep. The molecules are small, so the cost per step is dominated by per-call
+overhead rather than by system size, and tests are likely to take a couple of hours per
+model on GPU. Faster inference can be achieved using the jax-accelerated simulations in
+MLIP Audit directly.
 
 Data availability
 -----------------
--- a/ml_peg/calcs/molecular_dynamics/bond_length_distribution/calc_bond_length_distribution.py
+++ b/ml_peg/calcs/molecular_dynamics/bond_length_distribution/calc_bond_length_distribution.py
@@ -1,0 +1,67 @@
+"""
+Measure covalent bond length deviations during molecular dynamics.
+
+A molecular dynamics simulation is run for each of a set of small organic
+molecules, and the deviation of a tracked covalent bond from its reference
+equilibrium length is measured over the trajectory.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any
+from warnings import warn
+
+from mlipaudit.benchmarks.bond_length_distribution.bond_length_distribution import (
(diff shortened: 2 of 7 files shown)


docs/source/user_guide/benchmarks/molecular_dynamics.rst
```diff
@@ -76,3 +76,46 @@ Reference data:
...
+deviation from the reference bond length is averaged over the trajectory and across all
+molecules. A well behaved potential keeps bonds close to their reference length, so a lower
+deviation is better.
+
+A histogram shows the distribution of the sampled bond length deviations for each model.
+
+Computational cost
+------------------
+
+High: one MD simulation per molecule, each 1,000,000 steps. Faster inference can be achieved
```
joehart2001: gpu estimate?

joehart2001:
Thanks for the pr! same as #658 could you send the data + have a look at the element filtering?

docs: https://ddmms.github.io/ml-peg/developer_guide/filter.html

docs/source/user_guide/benchmarks/molecular_dynamics.rst
lwalew: There are 8 molecules of size 4-13 atoms, each doing 1M steps so for instance for UMA-S's throughput of 1.4ns/day/1000 atoms on an H100, we can expect at least 2h for instance. Although there will be some overhead associated with the simulation engine so for most models, a couple of hours is realistic.

lwalew: @joehart2001 I assume this benchmark (among others we're adding) should then be pytest-marked as slow or very slow. Is there any guidance as to how to choose between these two options?

joehart2001: i have a pr for guidance (not in yet sorry) which is broadly: slow = hours on a gpu, very slow = tens of hours on gpu

lwalew: Marked very slow in that case

joehart2001:
hey @lwalew, just tag/message us when we should take another look at the pr as i see you're making some changes. thanks!

lwalew:
> hey @lwalew, just tag/message us when we should take another look at the pr as i see you're making some changes. thanks!

Hi @joehart2001, I'm happy for another look on this one!

lwalew:
@joehart2001 I think we also need to upload the input data to your S3 bucket for this one - please tell me if you would need any assistance with this!

joehart2001:
> @joehart2001 I think we also need to upload the input data to your S3 bucket for this one - please tell me if you would need any assistance with this!

i just found it on hugging face and uploaded, will test

ml_peg/calcs/molecular_dynamics/bond_length_distribution/calc_bond_length_distribution.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    Benchmark covalent bond length deviations during MD.
+
+    Parameters
+    ----------
+    mlip
+        Name of model and model object to get calculator.
+    """
+    model_name, model = mlip
+    calc = model.get_calculator(precision="high")
```
joehart2001: do we want float64 here for long md? what do you think?

lwalew: Should be low I agree! I thought I had updated this

joehart2001:
@ElliottKasoar i'm happy to merge this, could you take a look?

ElliottKasoar (approved):
Looks good to me, thanks @lwalew!

PR #669, opened 2026-07-05
Add water-cl2 relaxation benchmark

kuryla:

## Summary

Physicality test probing the dissociation of a hydrated Cl2 molecule, with another water and 2xCl- cluster far away, and total charge being -2. The physical behaviour is the Cl2 molecule atoms staying together because the 2- charge is localised on the cluster with Cl- ions.

## Linked issue

Resolves #

## Progress

- [x] Calculations
- [x] Analysis
- [x] Application
- [x] Documentation

## Testing

MACE-MP-0a

### DIFF
--- a/ml_peg/calcs/physicality/water_cl2_relaxation/calc_water_cl2_relaxation.py
+++ b/ml_peg/calcs/physicality/water_cl2_relaxation/calc_water_cl2_relaxation.py
@@ -6,7 +6,7 @@ from pathlib import Path
 from typing import Any
 
 from ase import units
-from ase.io import read
+from ase.io import read, write
 from ase.optimize import LBFGS
 import pytest
 
@@ -57,3 +57,5 @@ def test_water_cl2(mlip: tuple[str, Any]) -> None:
     write_dir.mkdir(parents=True, exist_ok=True)
     opt = LBFGS(atoms, trajectory=str(write_dir / "relaxation.traj"))
     opt.run(fmax=0.01, steps=1000)
+    atoms_list = read(write_dir / "relaxation.traj", ":")
+    write(write_dir / "relaxation.xyz", atoms_list)
--- a/ml_peg/calcs/physicality/water_cl2_relaxation/calc_water_cl2_relaxation.py
+++ b/ml_peg/calcs/physicality/water_cl2_relaxation/calc_water_cl2_relaxation.py
@@ -6,13 +6,13 @@ from pathlib import Path
 from typing import Any
 
 from ase import units
-from ase.io import read, write
+from ase.io import read
 from ase.optimize import LBFGS
 import pytest
 
 from ml_peg.calcs.utils.utils import download_s3_data
+from ml_peg.models import current_models
 from ml_peg.models.get_models import load_models
-from ml_peg.models.models import current_models
 
 MODELS = load_models(current_models)
 
@@ -22,7 +22,7 @@ OUT_PATH = Path(__file__).parent / "outputs"
 
 
 @pytest.mark.parametrize("mlip", MODELS.items())
-def test_water_cl2(mlip: tuple[str, Any]) -> None:
+def test_water_cl2_relaxation(mlip: tuple[str, Any]) -> None:
     """
     Run WaterCl2 relaxation test.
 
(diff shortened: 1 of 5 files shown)


ElliottKasoar (dismissed):
I've made a few changes to tidy things up, but otherwise I think this is good to go. Thanks @kuryla!

@joehart2001, if you could take a quick look? One thing this struggles with a little is the visualisation with the point tracking, because it's so long. I think I saw somewhere that the way we do it is possibly redrawing the plot each time, rather than just moving the circle, so may be something to look into?

ml_peg/calcs/physicality/water_cl2_relaxation/calc_water_cl2_relaxation.py
```diff
@@ -0,0 +1,58 @@
...
+    atoms.calc = calc
+
+    write_dir = OUT_PATH / model_name
+    write_dir.mkdir(parents=True, exist_ok=True)
+    opt = LBFGS(
+        atoms,
+        trajectory=write_dir / "relaxation.traj",
+        logfile=write_dir / "relaxation.log",
+    )
+    opt.run(fmax=0.01, steps=1000)
```
joehart2001:
```suggestion
    opt.run(fmax=0.01, steps=2000)
```
should we bump this up to maybe 2K steps? e.g. mace-mp-0a finishes in ~800 for me so could be a little tight on some models finishing under the threshold?

joehart2001:
> I've made a few changes to tidy things up, but otherwise I think this is good to go. Thanks @kuryla!
> 
> @joehart2001, if you could take a quick look? One thing this struggles with a little is the visualisation with the point tracking, because it's so long. I think I saw somewhere that the way we do it is possibly redrawing the plot each time, rather than just moving the circle, so may be something to look into?

@ElliottKasoar what issues do you see for this? its super smooth for me. tested and am happy with rest

ElliottKasoar:
> > I've made a few changes to tidy things up, but otherwise I think this is good to go. Thanks @kuryla!
> > @joehart2001, if you could take a quick look? One thing this struggles with a little is the visualisation with the point tracking, because it's so long. I think I saw somewhere that the way we do it is possibly redrawing the plot each time, rather than just moving the circle, so may be something to look into?
> 
> @ElliottKasoar what issues do you see for this? its super smooth for me. tested and am happy with rest

Maybe it was my laptop just struggling, the highlighted point wasn't always moving with the frames, e.g. if I skipped along the trajectory using the WEAS slider

joehart2001:
> > > I've made a few changes to tidy things up, but otherwise I think this is good to go. Thanks @kuryla!
> > > @joehart2001, if you could take a quick look? One thing this struggles with a little is the visualisation with the point tracking, because it's so long. I think I saw somewhere that the way we do it is possibly redrawing the plot each time, rather than just moving the circle, so may be something to look into?
> > 
> > 
> > @ElliottKasoar what issues do you see for this? its super smooth for me. tested and am happy with rest
> 
> Maybe it was my laptop just struggling, the highlighted point wasn't always moving with the frames, e.g. if I skipped along the trajectory using the WEAS slider

hmm maybe too many tabs? i was actually quite surprised how smooth it was lol

ml_peg/calcs/physicality/water_cl2_relaxation/calc_water_cl2_relaxation.py
kuryla: That's fine!

joehart2001 (approved):

ElliottKasoar (approved):

ElliottKasoar:
> should we bump this up to maybe 2K steps? e.g. mace-mp-0a finishes in ~800 for me so could be a little tight on some models finishing under the threshold?

The downside to this is for models that explode, they'll end up doing 2000 potentially very slow steps... I think merge this, but can I suggest we look at a (general utility) function we attach to dynamics that interrupts if forces become say 10e6?

kuryla:
> > should we bump this up to maybe 2K steps? e.g. mace-mp-0a finishes in ~800 for me so could be a little tight on some models finishing under the threshold?
> 
> The downside to this is for models that explode, they'll end up doing 2000 potentially very slow steps... I think merge this, but can I suggest we look at a (general utility) function we attach to dynamics that interrupts if forces become say 10e6?

Good point, I haven't thought about this

PR #659, opened 2026-07-02
feat: add torsion net benchmark

lwehrhan:
## Pre-review checklist for PR author

PR author **must** check the checkboxes below when creating the PR.

- [x] I've confirmed the [contribution guidelines](https://github.com/ddmms/ml-peg/blob/main/contributing.md).

## Summary

The TorsionNet500 dataset consists of 500 drug-like organic molecules with systematically sampled dihedral angles. The target metric is the the barrier height of the dihedrals. The level of theory is ωB97M-D3(BJ)/def2-TZVPPD (re-computed by InstaDeep in-house).

## Linked issue

Resolves #642 

## Progress

- [x] Calculations
- [x] Analysis
- [x] Application
- [x] Documentation

Issue #642, opened 2026-06-24
feat: new benchmark torsion net 500 dihedral scan


lwehrhan:
### Summary

The TorsionNet500 dataset consists of 500 drug-like organic molecules with systematically sampled dihedral angles. The target metric is the the barrier height of the dihedrals. The level of theory is ωB97M-D3(BJ)/def2-TZVPPD (re-computed by InstaDeep in-house). 

### Interactive features

Parity plot.

### Category

Conformers

### Data availability

Geometries:

Brajesh K. Rai […] A. Bakken, Journal of Chemical Information and Modeling 2022 62 (4), 785-800. DOI: 10.1021/acs.jcim.1c01346

Reference energies: MLIP Audit Huggingface.

### Computational cost

Medium: 12,000 inference calls in total with small molecular systems.

PR #682, opened 2026-07-08
Add TorsionNet500CCSDT benchmark visualisation, MAE metric, and docs

### LINKED ISSUE #679: TorsionNet500CCSDT benchmark
### Summary

TorsionNet500CCSDT: evaluates predicted torsional energy profiles for 500 diverse organic molecular fragments against DLPNO-CCSD(T)/CBS reference torsion scans. Reports RMSE and MAE of the mean-centered relative energy profile, averaged across all fragments per model.

### Interactive features

Clicking a model's RMSE or MAE cell in the table shows that model's per-fragment scatter plot (fragment index vs metric value). Clicking a point on that scatter then shows the corresponding dihedral-angle-vs-energy curve for that fragment, comparing the CCSD(T) reference against the model's prediction.

### Category

Conformers

### Data availability

Input structures: B. K. Rai, V. Sresht, Q. Yang, R. Unwalla, M. Tu, A. M. Mathiowetz, and G. A. Bakken, TorsionNet: A Deep Neural Network to Rapidly Predict Small-Molecule Torsional Energy Profiles with the Accuracy of Quantum Mechanics, Journal of Chemical Information and Modeling 62 (2022), 785-800.

Reference data (DLPNO-CCSD(T)/CBS): J. L. Weber et al., Efficient Long-Range Machine Learning Force Fields for Liquid and Materials Properties, arXiv:2505.06462 (2025)

### Computational cost

Medium: hours per model on CPU, since each of the 500 fragments requires a single-point energy calculation across ~20 conformers in its torsion scan.

### Additional context

Deliberately named `TorsionNet500CCSDT` rather than `TorsionNet500` to coexist alongside the existing barrier-height implementation in #642.

ellalbolland:

## Summary

Adds per-fragment RMSE/MAE scatter plots and per-fragment torsion energy profile curves to the TorsionNet500CCSDT analysis, enabling table-cell → scatter → curve click-through navigation.

Named TorsionNet500CCSDT rather than TorsionNet500 to coexist alongside the existing barrier-height implementation in #642/#659, since this benchmark reports RMSE/MAE of the full relative energy profile against CCSD(T)/CBS reference data - a different metric and reference level of theory, not a duplicate.

## Linked issue

Resolves #679

## Progress

- [x] Calculations
- [x] Analysis
- [x] Application
- [x] Documentation

## Testing

Tested against 16 of the 21 registered models (full 500/500-fragment recompute completed for each). The uma family (uma-s-1p1-omat, uma-m-1p1-omat, uma-s-1p1-omol, uma-s-1p2-omol, uma-m-1p1-omol) could not be run due to lacking HuggingFace authorisation for the gated UMA model weights.
Confirmed working end-to-end by running the app standalone (`uv run python ml_peg/app/conformers/TorsionNet500CCSDT/app_TorsionNet500CCSDT.py`) and clicking through the table -> scatter -> curve interactivity manually.

## New decorators/callbacks

One new local callback, `_register_curve_callback` in `app_TorsionNet500CCSDT.py`. `plot_from_scatter` pre-loads every possible figure upfront (too expensive for 500 curves/model); `struct_from_scatter` lazy-loads but only for structures, not Plotly figures. This one lazy-loads a curve JSON from file per click instead. Flagged with a `TODO` to move it into `build_callbacks.py` if another benchmark needs the same pattern.

### DIFF
--- a/docs/source/user_guide/benchmarks/conformers.rst
+++ b/docs/source/user_guide/benchmarks/conformers.rst
@@ -93,3 +93,51 @@ Reference data:
 
 * Same as input data
 * :math:`DLPNO-CCSD(T)` level of theory: a local coupled-cluster method.
+
+
+TorsionNet500CCSDT
+===================
+
+Summary
+-------
+
+Performance in predicting torsional energy profiles for 500 diverse organic
+molecular fragments. Reference data from DLPNO-CCSD(T)/CBS calculations.
+
+Metrics
+-------
+
+1. RMSE of relative torsional energy profile
+2. MAE of relative torsional energy profile
+
+For each fragment, a torsion scan samples the energy at a series of dihedral
+angles. Both the reference and predicted energies are mean-centered per scan,
+so only the shape of the profile is compared rather than absolute energy
+offsets. RMSE and MAE are calculated between the reference and predicted
+profiles for each scan, then averaged across all 500 fragments.
+
+Computational cost
+------------------
+
+Medium: tests are likely to take hours per model to run on CPU, since each
+fragment requires a single-point energy calculation across ~20 conformers in
+its torsion scan, repeated for all 500 fragments.
+
+Data availability
+-----------------
+
+Input structures:
+
+* B. K. Rai, V. Sresht, Q. Yang, R. Unwalla, M. Tu, A. M. Mathiowetz, and
+  G. A. Bakken, TorsionNet: A Deep Neural Network to Rapidly Predict
+  Small-Molecule Torsional Energy Profiles with the Accuracy of Quantum
+  Mechanics, Journal of Chemical Information and Modeling 62 (2022), 785-800.
+  PMID: 35119861.
+
+Reference data:
+
+* J. L. Weber, R. D. Guha, G. Agarwal, Y. Wei, A. A. Fike, X. Xie,
+  J. Stevenson, B. Santra, R. A. Friesner, K. Leswing, M. D. Halls, R. Abel,
+  and L. D. Jacobson, Efficient Long-Range Machine Learning Force Fields for
+  Liquid and Materials Properties, arXiv:2505.06462 (2025).
+* DLPNO-CCSD(T)/CBS level of theory.
--- a/docs/source/user_guide/benchmarks/conformers.rst
+++ b/docs/source/user_guide/benchmarks/conformers.rst
@@ -119,9 +119,10 @@ profiles for each scan, then averaged across all 500 fragments.
 Computational cost
 ------------------
 
-Medium: tests are likely to take hours per model to run on CPU, since each
-fragment requires a single-point energy calculation across ~20 conformers in
-its torsion scan, repeated for all 500 fragments.
+Medium: tests are likely to take minutes to run on GPU, or less than an hour on
+CPU for each model, since each fragment requires a single-point energy
+calculation across ~20 conformers in its torsion scan, repeated for all 500
+fragments.
 
 Data availability
 -----------------
--- a/ml_peg/analysis/conformers/TorsionNet500CCSDT/analyse_TorsionNet500CCSDT.py
+++ b/ml_peg/analysis/conformers/TorsionNet500CCSDT/analyse_TorsionNet500CCSDT.py
@@ -10,7 +10,7 @@ import plotly.graph_objects as go
 import pytest
 
 from ml_peg.analysis.utils.decorators import build_table
-from ml_peg.analysis.utils.utils import load_metrics_config, mae, rmse
+from ml_peg.analysis.utils.utils import get_struct_info, load_metrics_config, mae, rmse
 from ml_peg.app import APP_ROOT
 from ml_peg.calcs import CALCS_ROOT
 from ml_peg.models import current_models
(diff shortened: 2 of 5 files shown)


docs/source/user_guide/benchmarks/conformers.rst
```diff
@@ -40,3 +40,50 @@ Reference data:
...
+For each fragment, a torsion scan samples the energy at a series of dihedral
+angles. Both the reference and predicted energies are mean-centered per scan,
+so only the shape of the profile is compared rather than absolute energy
+offsets. RMSE and MAE are calculated between the reference and predicted
+profiles for each scan, then averaged across all 500 fragments.
+
+Computational cost
+------------------
+
+Medium: tests are likely to take hours per model to run on CPU, since each
```
joehart2001: gpu estimate?

joehart2001:
Thanks @ellalbolland, from a brief scan its looking great! For the uma model access you can request it i think here: https://huggingface.co/facebook/UMA
it would be good to test. i think you then also need to install the hugging face cli etc. but that should be straightforward

- please could you also send us the input files for the calc so we can upload them and test it
- also add the element filtering too

ellalbolland:
Here's the input dataset used for the calc: [TorsionNet500_xyz (1).zip](https://github.com/user-attachments/files/29962757/TorsionNet500_xyz.1.zip)

joehart2001:
> Here's the input dataset used for the calc: [TorsionNet500_xyz (1).zip](https://github.com/user-attachments/files/29962757/TorsionNet500_xyz.1.zip)

thanks uploaded

ml_peg/calcs/conformers/TorsionNet500CCSDT/calc_TorsionNet500CCSDT.py
```diff
@@ -0,0 +1,65 @@
...
+    mlip
+        Name of model and model to get calculator.
+    """
+    model_name, model = mlip
+    try:
+        calc = model.get_calculator(precision="high")
+    except ModuleNotFoundError as exc:
+        pytest.skip(f"Skipping {model_name}: {exc}")
+
+    xyz_files = sorted(DATA_PATH.glob("*.xyz"))
```
joehart2001:
```suggestion
    data_path = (
        download_s3_data(
            filename="TorsionNet500CCSDT.zip",
            key="inputs/conformers/TorsionNet500CCSDT/TorsionNet500CCSDT.zip",
        )
        / "TorsionNet500CCSDT"
    )

    xyz_files = sorted(data_path.glob("*.xyz"))
```

ml_peg/calcs/conformers/TorsionNet500CCSDT/calc_TorsionNet500CCSDT.py
```diff
@@ -0,0 +1,65 @@
...
+from pathlib import Path
+from typing import Any
+from warnings import warn
+
+from ase import units
+from ase.io import read, write
+import numpy as np
+import pytest
+from tqdm import tqdm
+
```
joehart2001, line 14:
```suggestion

from ml_peg.calcs.utils.utils import download_s3_data
```

docs/source/user_guide/benchmarks/conformers.rst
joehart2001: this takes about 10 minutes for me locally on cpu, so lets change to less than an hour on cpu and minutes on GPU

joehart2001:
Hi @ellalbolland, everything is looking really nice so far!

One thing that would be great to add is visualisaiton from the torsion curve plot.  you could take a look at some of the NEB curves to see how they use the trajectory mode to visualise data points along the curve. thanks!

joehart2001:
the visualisaition is looking great! @ElliottKasoar could you take a look when you have time

ml_peg/analysis/conformers/TorsionNet500CCSDT/analyse_TorsionNet500CCSDT.py
```diff
@@ -0,0 +1,368 @@
...
+        labels = []
+        rmse_scans = []
+        mae_scans = []
+        profiles = []
+
+        if model_dir.exists():
+            xyz_files = sorted(model_dir.glob("*.xyz"))
+
+            for xyz_file in xyz_files:
+                atoms = read(xyz_file, ":")
```
ElliottKasoar, line 68: We may want to be a bit careful here, as there's nothing to check if some of the data is missing, as far as I can tell. For example, I interrupted one of my models after it only ran a couple of structures, and it got a score, presumably based on the average of those.

We do want (1) this to not break if data is missing and (2) to be flexible enough that when we do "full" filtering, we can return the full list of errors that were calculated, but we want to make sure a model has all of the errors that we expect, and give a None/NaN score in if not.

One way to do that is to use the labels/filename stems from `get_struct_info`, but any way to do this would be good.

In practice, I didn't see any models fail calculations, so the only real risk is if a calculation gets interrupted, but we should catch this.

ml_peg/app/conformers/TorsionNet500CCSDT/app_TorsionNet500CCSDT.py
```diff
@@ -0,0 +1,256 @@
...
+    scatter_id: str,
+    plot_id: str,
+    curve_dir: Path,
+    labels: list[str],
+    struct_plot_id: str,
+) -> None:
+    """
+    Attach callbacks that show a torsion curve and structure on point clicks.
+
+    Unlike `plot_from_scatter`, this reads the clicked fragment's curve JSON from
```
ElliottKasoar, line 46: Is the only difference the lazy loading? It might be worth just doing this for all plots from scatters, or at least combining these with a `lazy_load` flag

ElliottKasoar (commented):
This looks really nice, thanks @ellalbolland! I believe @lwehrhan proposed closing #659 and adding their data to this, so perhaps we should just rename the benchmark TorsionNet500?

Otherwise just a couple of minor suggestions/questions

I just rebased and fixed the conflict in the docs, hopefully that's ok

lwehrhan:
Hi @ellalbolland , if it is okay with you, I would like to close #659 and add our DFT labels to this PR (as Elliott suggested).

ellalbolland:
Hi @lwehrhan, apologies for the delayed reply, I was away on holiday. Yes, that sounds great to me. Thank you for this!

ml_peg/analysis/conformers/TorsionNet500CCSDT/analyse_TorsionNet500CCSDT.py
ElliottKasoar: As with #713, I'm less worried about this in retrospect, so let's just focus on adding the extra data and hopefully we can merge this

PR #683, opened 2026-07-08
autocomplete cli for tests and models

joehart2001:

## Linked issue

Resolves #

### DIFF
--- a/ml_peg/cli/cli.py
+++ b/ml_peg/cli/cli.py
@@ -111,12 +111,14 @@ def complete_test(ctx: Context, incomplete: str) -> list[str]:
     ]
 
 
-def complete_models(incomplete: str) -> list[str]:
+def complete_models(ctx: Context, incomplete: str) -> list[str]:
     """
     Complete comma-separated model names from ``models.yml``.
 
     Parameters
     ----------
+    ctx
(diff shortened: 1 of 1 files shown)


ml_peg/cli/cli.py
```diff
@@ -74,6 +74,69 @@ def get_tests(root: Path, script_prefix: str, category: str = "*") -> tuple[str,
...
+    Returns
+    -------
+    list[str]
+        Matching model names, preserving any already-typed comma-separated prefix.
+    """
+    import yaml
+
+    from ml_peg.models import MODELS_ROOT
+
+    with open(MODELS_ROOT / "models.yml", encoding="utf8") as handle:
```
ElliottKasoar: We probably want to do something closer to what we do in `list_models` where we use the `models_file` (so the models we autocomplete with is from the correct list) and `get_model_names`, (so we're not rewriting parsing, if we need to change something)

joehart2001: what do you think now?

ElliottKasoar (approved):
This is really nice, thanks!

ml_peg/cli/cli.py
ElliottKasoar: Works great!

PR #692, opened 2026-07-12
Add thermal conductivity benchmark kSRME

### LINKED ISSUE #92: Thermal Conductivity
### Summary

Anharmonic phonon properties and lattice thermal conductivity for binary crystals, to calculate κRMSE

### Interactive features

Bar chart for κRMSE?

### Category

Bulk crystal / materials

### Data availability

103-compound Wigner-transport benchmark

### Computational cost

Unknown (medium - high?)

### Additional context

_No response_

bpota:

## Summary

Adding k_SRME thermal conductivity benchmark.

Adding an adapted version of the Matbench-Discovery thermal conductivity script, with a FAST option to run a lower mesh and speed up the phono3py runtime, and reference data generation script. Please let me know where I could upload/send the computed reference data. The app seems to work except the plots, but the PR can have some other bugs too.

## Linked issue

Resolves #92 

## Progress

- [x] Calculations
- [x] Analysis
- [x] Application
- [x] Documentation

## Testing

No models have been tested within ml-peg. Please let me know what is the best way to arrange this. Prediction files should be compatible with Matbench-Discovery for convenience.

## New decorators/callbacks

Not yet.

### DIFF
--- a/ml_peg/analysis/bulk_crystal/thermal_conductivity/analyse_thermal_conductivity.py
+++ b/ml_peg/analysis/bulk_crystal/thermal_conductivity/analyse_thermal_conductivity.py
@@ -2,9 +2,9 @@
 
 from __future__ import annotations
 
-import os
 from pathlib import Path
 import traceback
+import warnings
 
 from ase.io import read
 import h5py
@@ -25,7 +25,7 @@ CALC_PATH = CALCS_ROOT / "bulk_crystal" / "thermal_conductivity" / "outputs"
 REF_PATH = CALCS_ROOT / "bulk_crystal" / "thermal_conductivity" / "data"
 OUT_PATH = APP_ROOT / "data" / "bulk_crystal" / "thermal_conductivity"
 
-METRICS_CONFIG_PATH = Path(__file__).with_name("metrics.yaml")
+METRICS_CONFIG_PATH = Path(__file__).with_name("metrics.yml")
 DEFAULT_THRESHOLDS, DEFAULT_TOOLTIPS, DEFAULT_WEIGHTS = load_metrics_config(
     METRICS_CONFIG_PATH
 )
@@ -166,7 +166,8 @@ def calc_kappa_srme_dataframes(
 
         # NOTE code below just until before return used to be wrapped in try/except in
         # which case SRME=2 was set for the failing material
-        if row_pred.get(tc.TCKeys.has_imag_ph_modes) is True:
+        has_imag_ph_modes = row_pred.get(tc.TCKeys.has_imag_ph_modes, False)
+        if pd.notna(has_imag_ph_modes) and bool(has_imag_ph_modes):
             srme_list.append(2)
             continue
         if relaxed_space_group_number := row_pred.get(tc.TCKeys.final_spg_num):
@@ -270,6 +271,42 @@ def calc_kappa_srme(kappas_pred: pd.Series, kappas_true: pd.Series) -> np.ndarra
     return 2 * microscopic_error / denominator
 
 
+def _add_missing_error_rows(
+    df: pd.DataFrame,
+    ref_df: pd.DataFrame,
+    model_name: str,
+    result_type: str = "",
+) -> pd.DataFrame:
+    """
+    Warn and add default-error rows for missing result files.
+
+    Parameters
+    ----------
+    df : pd.DataFrame
+        Prediction results.
+    ref_df : pd.DataFrame
+        Reference results defining the required index.
+    model_name : str
+        Model name used in the warning.
+    result_type : str
+        Optional result-type prefix used in the warning.
+
+    Returns
+    -------
+    pd.DataFrame
+        Prediction results reindexed to the reference when rows are missing.
+    """
+    missing = ref_df.index.difference(df.index)
+    if len(missing):
+        warnings.warn(
+            f"Missing {len(missing)} {result_type}thermal-conductivity result files "
+            f"for {model_name}; using default error rows.",
+            stacklevel=2,
+        )
+        return df.reindex(ref_df.index)
+    return df
+
+
 @pytest.fixture
 def kappa_stats() -> dict[str, pd.DataFrame]:
     """
@@ -312,18 +349,17 @@ def kappa_stats() -> dict[str, pd.DataFrame]:
             )
             pred_df = pd.DataFrame.from_dict(current_pred_dict, orient="index")
         elif (model_dir / "kappas.json.gz").exists():
-            pred_df = pd.read_json(model_dir / "kappas.json.gz", orient="index")
+            pred_df = pd.read_json(model_dir / "kappas.json.gz").set_index(
+                tc.TCKeys.mat_id
+            )
         else:
-            subdirs = sorted(os.listdir(model_dir))
-
-            if not any(not (Path(d) / "kappa.hdf5").exists() for d in subdirs):
-                try:
-                    current_pred_dict = tc.load_hdf5_subdir_dicts(
-                        model_dir, "kappa.hdf5"
-                    )
-                except Exception:
-                    print(f"Error loading kappas for {model_name}...")
-                    traceback.print_exc()
+            try:
+                current_pred_dict = tc.load_hdf5_subdir_dicts(model_dir, "kappa.hdf5")
+                if current_pred_dict:
+                    pred_df = pd.DataFrame.from_dict(current_pred_dict, orient="index")
+            except Exception:
+                print(f"Error loading kappas for {model_name}...")
+                traceback.print_exc()
 
         if (model_dir / "fast_kappas.hdf5").exists():
             current_pred_dict = tc.hdf5_to_dict(
@@ -331,20 +367,28 @@ def kappa_stats() -> dict[str, pd.DataFrame]:
             )
             fast_pred_df = pd.DataFrame.from_dict(current_pred_dict, orient="index")
         elif (model_dir / "fast_kappas.json.gz").exists():
-            fast_pred_df = pd.read_json(
-                model_dir / "fast_kappas.json.gz", orient="index"
+            fast_pred_df = pd.read_json(model_dir / "fast_kappas.json.gz").set_index(
+                tc.TCKeys.mat_id
             )
         else:
-            subdirs = sorted(os.listdir(model_dir))
-
-            if not any(not (Path(d) / "fast_kappa.hdf5").exists() for d in subdirs):
-                try:
-                    current_pred_dict = tc.load_hdf5_subdir_dicts(
-                        model_dir, "fast_kappa.hdf5"
+            try:
+                current_pred_dict = tc.load_hdf5_subdir_dicts(
+                    model_dir, "fast_kappa.hdf5"
+                )
+                if current_pred_dict:
+                    fast_pred_df = pd.DataFrame.from_dict(
+                        current_pred_dict, orient="index"
                     )
-                except Exception:
-                    print(f"Error loading fast_kappas for {model_name}...")
-                    traceback.print_exc()
+            except Exception:
+                print(f"Error loading fast_kappas for {model_name}...")
+                traceback.print_exc()
+
+        if pred_df is not None:
+            pred_df = _add_missing_error_rows(pred_df, ref_df, model_name)
+        if fast_pred_df is not None:
+            fast_pred_df = _add_missing_error_rows(
+                fast_pred_df, fast_ref_df, model_name, "fast "
+            )
 
         if pred_df is None and fast_pred_df is None:
             continue
@@ -411,6 +455,7 @@ def conductivity(kappa_stats: dict[str, pd.DataFrame]) -> dict[str, list]:
         df = kappa_stats[model_name]
         if tc.TCKeys.kappa_tot_avg not in df:
             continue
+        df = df.reindex(kappa_stats["ref"].index)
         conductivity_data[model_name] = df[tc.TCKeys.kappa_tot_avg].tolist()
     conductivity_data["ref"] = kappa_stats["ref"][tc.TCKeys.kappa_tot_avg].tolist()
     return conductivity_data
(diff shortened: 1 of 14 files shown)


ml_peg/analysis/bulk_crystal/thermal_conductivity/analyse_thermal_conductivity.py
```diff
@@ -0,0 +1,698 @@
...
+def test_thermal_conductivity(metrics: dict[str, dict]) -> None:
+    """
+    Run thermal conductivity benchmark tests.
+
+    Parameters
+    ----------
+    metrics : dict[str, dict]
+        Metric names and values for all models.
+    """
+    write_struct_info(data_path=STRUCTURE_FILE, out_path=OUT_PATH)
```
ElliottKasoar: Is there a reason to use this rather than `get_struct_info`, which also has a write info option?

This would save reading `STRUCTURE_FILE` an additional two times as part of `get_system_names` and `get_system_ids`.

ElliottKasoar:
Thanks for this, @bpota! We'll try to take a detailed look as soon as we can.

Regarding the data, if you want us to upload it to the S3 bucket, you can share the data with me or @joehart2001 in whatever way is easiest (here/OneDrive/Dropbox/email/Slack) and we can upload it for you.

joehart2001:
I think a good way to do the fast only and full tests would be to:
- have general running function
- mark the full test as `very_slow` like e.g. phonons
- this means when you run the test normally, on the fast one will run, and if you use the very slow cli flag then both will run

ml_peg/analysis/bulk_crystal/thermal_conductivity/analyse_thermal_conductivity.py
bpota: I don't think there should be any reason against changing. Thanks for the suggestion!

bpota:
> I think a good way to do the fast only and full tests would be to:
> 
> * have general running function
> * mark the full test as `very_slow` like e.g. phonons
> * this means when you run the test normally, on the fast one will run, and if you use the very slow cli flag then both will run

Thanks @joehart2001, I agree this should be easier. I will change it accordingly.

ml_peg/calcs/bulk_crystal/thermal_conductivity/data/collect_ref_kappas.py
```diff
@@ -0,0 +1,73 @@
...
+
+It loads the data from JSON and HDF5 files, processes it into pandas DataFrames,
+and saves it back to disk in both formats for easy access in future analyses.
+"""
+
+from __future__ import annotations
+
+import importlib
+from pathlib import Path
+
```
joehart2001, line 12:
```suggestion

import sys
```

ml_peg/calcs/bulk_crystal/thermal_conductivity/data/collect_ref_kappas.py
```diff
@@ -0,0 +1,73 @@
...
+import h5py
+import pandas as pd
+
+DATA_PATH = Path(__file__).parent
+
+tc = importlib.import_module(DATA_PATH.parent / "thermal_conductivity.py")
```
joehart2001:
```suggestion
sys.path.append(str(DATA_PATH.parent))
tc = importlib.import_module("thermal_conductivity")
```

ml_peg/calcs/bulk_crystal/thermal_conductivity/data/reference_generation.py
```diff
@@ -0,0 +1,360 @@
+"""Reference data generation for thermal conductivity calculations using phono3py."""
+
+from __future__ import annotations
+
```
joehart2001, line 4:
```suggestion

from concurrent.futures import ThreadPoolExecutor
```

ml_peg/calcs/bulk_crystal/thermal_conductivity/data/reference_generation.py
```diff
@@ -0,0 +1,360 @@
...
+import importlib
+from io import BytesIO
+import os
+from pathlib import Path
+import re
+import sys
+
```
joehart2001, line 13:
```suggestion
import time

```

ml_peg/calcs/bulk_crystal/thermal_conductivity/data/reference_generation.py
```diff
@@ -0,0 +1,360 @@
...
+    for link in tqdm(params_links, desc="Download"):
+        xz_bytes = requests.get(link, timeout=30).content
+
+        yaml_bytes = lzma.decompress(xz_bytes)
+
+        yaml_str = yaml_bytes.decode("utf-8")
+
+        yaml_list.append(yaml_str)
+
```
joehart2001:
```suggestion
    def _download_one(link: str) -> str:
        """
        Download and decompress one phono3py params YAML, retrying on failure.

        Parameters
        ----------
        link : str
            URL of the ``.yaml.xz`` file.

        Returns
        -------
        str
            Decompressed YAML content.
        """
        last_exc: Exception | None = None
        for attempt in range(3):
            try:
                xz_bytes = requests.get(link, timeout=30).content
                return lzma.decompress(xz_bytes).decode("utf-8")
            except Exception as exc:
                last_exc = exc
                time.sleep(2**attempt)
        raise RuntimeError(f"Failed to download {link} after 3 attempts") from last_exc

    # Downloads are I/O-bound, so thread them. map preserves order so yaml_list
    # stays aligned with params_links.
    with ThreadPoolExecutor(max_workers=8) as executor:
        yaml_list = list(
            tqdm(
                executor.map(_download_one, params_links),
                total=len(params_links),
                desc="Download",
            )
        )

```

ml_peg/calcs/bulk_crystal/thermal_conductivity/calc_thermal_conductivity.py
```diff
@@ -0,0 +1,376 @@
...
+        fast_results_path = out_dir / "fast_kappa.hdf5"
+        with h5py.File(fast_results_path, "w") as f:
+            tc.dict_to_hdf5(fast_results_dict, f)
+        fast_kappa_dicts[structure_id] = fast_results_dict
+
+    if PARALLEL_TASK_NUM == 1:
+        if not FAST_ONLY:
+            df = pd.DataFrame(kappa_dicts).T
+            df.index.name = tc.TCKeys.mat_id
+            df.reset_index().to_json(OUT_PATH / model_name / "kappas.json.gz")
```
joehart2001:
```suggestion
            # material_id is already a column, so drop the redundant index.
            df.reset_index(drop=True).to_json(OUT_PATH / model_name / "kappas.json.gz")
```

ml_peg/calcs/bulk_crystal/thermal_conductivity/calc_thermal_conductivity.py
```diff
@@ -0,0 +1,376 @@
...
+            with h5py.File(OUT_PATH / model_name / "kappas.hdf5", "w") as f:
+                tc.dict_to_hdf5(kappa_dicts, f)
+
+        fast_df = pd.DataFrame(fast_kappa_dicts).T
+        fast_df.index.name = tc.TCKeys.mat_id
+        fast_df.reset_index().to_json(OUT_PATH / model_name / "fast_kappas.json.gz")
```
joehart2001:
```suggestion
        fast_df.reset_index(drop=True).to_json(
            OUT_PATH / model_name / "fast_kappas.json.gz"
        )
```

ml_peg/calcs/bulk_crystal/thermal_conductivity/collect_kappas.py
```diff
@@ -0,0 +1,74 @@
...
+    output_name_no_ext : str
+        The base name to use for the output JSON and HDF5 files (without extension).
+    """
+    print(f"Loading {filename_no_ext} from {parent_dir} subdirectories...")
+    dicts = tc.load_hdf5_subdir_dicts(parent_dir, filename_no_ext + ".hdf5")
+    print("Loading finished.")
+
+    df = pd.DataFrame(dicts).T
+    df.index.name = tc.TCKeys.mat_id
+    df.reset_index().to_json(OUT_PATH / f"{output_name_no_ext}.json.gz")
```
joehart2001:
```suggestion
    # material_id is already a column, so drop the redundant index.
    df.reset_index(drop=True).to_json(OUT_PATH / f"{output_name_no_ext}.json.gz")
```

joehart2001:
The run failed at the very end (after all 103 materials had been computed and saved) when writing the combined `fast_kappas.json.gz` / `kappas.json.gz`: `ValueError: cannot insert material_id, already exists`

- each per-structure result dict already contains a material_id key, so pd.DataFrame(...).T has a material_id column. The code then set the index name to material_id and called reset_index(), which tries to promote that index into a second material_id column -> pandas rejects the duplicate

added code suggestions above for this

bpota:
Ideas for interactivity after discussing with @joehart2001:
- (Fast) kSRE: Predicted thermal conductivity against reference on a log-log scale.
- (Fast) kSRME: Same violin plot as in the case of phonons.
- Individual datapoints for both kSRE and kSRME: Predicted mode thermal conductivity against reference on a log-log scale.

joehart2001:
@ElliottKasoar i think we're happy with this now. can you take a look when you've got time?

ml_peg/analysis/bulk_crystal/thermal_conductivity/metrics.yml
ElliottKasoar, line 1: Apologies if this has been discussed elsewhere (probably with @joehart2001), but are these thresholds based on things from Matbench Discovery? Is there no known uncertainty in the errors, and is there no point before being pretty much entirely wrong where we'd consider these useless?

ml_peg/calcs/bulk_crystal/thermal_conductivity/calc_thermal_conductivity.py
```diff
@@ -0,0 +1,396 @@
...
+    "QuasiNewton",
+    "SciPyFminBFGS",
+    "BFGS",
+    "LBFGSLineSearch",
+    "SciPyFminCG",
+    "FIRE2",
+    "FIRE",
+    "LBFGS",
+] = "FIRE"  # "LBFGS"
+ase_filter: Literal["frechet", "exp"] = "frechet"
```
ElliottKasoar, line 68: Do we really need these listed? We only use one of each, and it's not even an input?

ml_peg/calcs/bulk_crystal/thermal_conductivity/calc_thermal_conductivity.py
```diff
@@ -0,0 +1,396 @@
...
+    except ValueError:
+        print(
+            "Could not parse parallel task arguments, running in serial.",
+            file=sys.stderr,
+        )
+        PARALLEL_TASK_ID = 0
+        PARALLEL_TASK_NUM = 1
+else:
+    PARALLEL_TASK_ID = 0
+    PARALLEL_TASK_NUM = 1
```
ElliottKasoar, line 52: I don't think this works with our CLI? We can use environment variables, or add custom CLI inputs if we want though?

ml_peg/calcs/bulk_crystal/thermal_conductivity/calc_thermal_conductivity.py
```diff
@@ -0,0 +1,396 @@
...
+    Run thermal conductivity test.
+
+    Parameters
+    ----------
+    mlip
+        Name of model use and model to get calculator.
+    """
+    model_name, model = mlip
+    model.default_dtype = default_dtype
+    calculator = model.get_calculator()
```
ElliottKasoar, line 102:
```suggestion

FAST_ONLY = False


@pytest.mark.very_slow
@pytest.mark.parametrize("mlip", MODELS.items())
def test_thermal_conductivity(mlip: tuple[str, Any]) -> None:
    """
    Run thermal conductivity test.

    Parameters
    ----------
    mlip
        Name of model use and model to get calculator.
    """
    model_name, model = mlip
    calculator = model.get_calculator(precision="high")
```
These are usually doing the same thing, but don't necessarily for all models

ml_peg/calcs/bulk_crystal/thermal_conductivity/calc_thermal_conductivity.py
```diff
@@ -0,0 +1,396 @@
...
+        "QuasiNewton": ase.optimize.BFGSLineSearch,
+        "SciPyFminBFGS": ase.optimize.sciopt.SciPyFminBFGS,
+        "BFGS": ase.optimize.BFGS,
+        "LBFGSLineSearch": ase.optimize.LBFGSLineSearch,
+        "SciPyFminCG": ase.optimize.sciopt.SciPyFminCG,
+        "FIRE2": ase.optimize.FIRE2,
+        "FIRE": ase.optimize.FIRE,
+        "LBFGS": ase.optimize.LBFGS,
+    }
+    optim_cls: type[Optimizer] = optimizer_dict[ase_optimizer]
```
ElliottKasoar, line 241: As above, I don't think we're really benefiting from this unless we want it to be customisable

ml_peg/calcs/bulk_crystal/thermal_conductivity/collect_kappas.py
ElliottKasoar, line 1: How is this supposed to be used? Normally this sort of thing would be called by the analysis script, not have its own standalone CLI

ml_peg/calcs/bulk_crystal/thermal_conductivity/data/reference_generation.py
```diff
@@ -0,0 +1,392 @@
...
+    phono3py_yaml_path : str, optional
+        Local directory to save the scraped YAML files, by default "data/DFT".
+
+    Returns
+    -------
+    tuple[list[str], list[str]]
+        A tuple containing a list of YAML strings and a corresponding list of material
+        names.
+    """
+    raw_url = "https://raw.githubusercontent.com/atztogo/phonondb/refs/heads/main/mdr/phono3py_103compounds_fd_PBE/README.md"
```
ElliottKasoar, line 183: Maybe we should use a tag for this, in case the README ever changes?

ElliottKasoar (commented):
Thanks for this, @bpota and @joehart2001! A few minor suggestions/questions, but otherwise it's looking really nice.

ml_peg/calcs/bulk_crystal/thermal_conductivity/calc_thermal_conductivity.py
ElliottKasoar: Out of interest, have you tested different optimisers? I know this can make a big difference in some cases (e.g. see SI from https://arxiv.org/pdf/2412.19330, which led to changes in #833 )

PR #699, opened 2026-07-16
Add weight presets

joehart2001:

## Summary
- add presets such as materials and molecules which weights specific categories as 0 and 1

## Linked issue

Resolves #

### DIFF
--- a/ml_peg/app/build_app.py
+++ b/ml_peg/app/build_app.py
@@ -3,15 +3,12 @@
 from __future__ import annotations
 
 from importlib import import_module
-import random
 import warnings
 
 from dash import (
-    ALL,
     Dash,
     Input,
     Output,
-    State,
     callback,
     clientside_callback,
     ctx,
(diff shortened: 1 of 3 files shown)


ml_peg/app/build_app.py
```diff
@@ -959,6 +998,65 @@ def build_nav(
         style={"marginBottom": "8px", "fontSize": "13px"},
     )
 
+    preset_names = [*WEIGHT_PRESETS, "Random"]
```
ElliottKasoar: Why do we want a random option?

ml_peg/app/build_app.py
ElliottKasoar, line 1: I think basically all of this should be in one or two of the utils files. This isn't a brand new problem, but in general `build_app` has become very bloated, it really ought to just be putting together the pieces of the app

ElliottKasoar (commented):
Seems to work great! Just a couple of questions/suggestions

ml_peg/app/build_app.py
joehart2001: @alinelena suggested for fun

ElliottKasoar: Ok sure

ml_peg/app/build_app.py
joehart2001: yep good idea, updated

ElliottKasoar (approved):
Looks good, thanks!

PR #713, opened 2026-07-21
Add Jacobian Symmetry physicality benchmark

### LINKED ISSUE #712: Jacobian Symmetry
### Summary

A physicality benchmark testing whether a model's predicted forces are conservative: i.e. the gradient of a single underlying energy. Following Bigi, Langer & Ceriotti's "The Dark Side of the Forces" ([arXiv:2412.11569](https://arxiv.org/abs/2412.11569)), the force Jacobian (∂fᵢ/∂rⱼ) is built by central finite differences and the antisymmetric fraction of its Frobenius norm, λ = ‖J_anti‖ / ‖J‖ where J_anti = (J − Jᵀ)/2, is reported. λ = 0 for a perfectly conservative model and grows towards 1 as forces become non-conservative. This probes what accuracy benchmarks miss. Non-conservative models predict forces directly rather than as gradients of an energy, which is cheaper but breaks energy conservation. The paper shows this causes ill-defined geometry optimization and unstable molecular dynamics, and unlike symmetry breaking, it is hard to detect and correct.

### Interactive features

Table 

### Category

Physicality

### Data availability

None required. All ten test structures (eight molecules + a metal + a 2D sheet) are generated in ASE.

### Computational cost

Low: a few minutes for the full model suite on CPU (well under a minute of compute per model). Cost scales with degrees of freedom, at 2 × 3N force evaluations per structure.

### Additional context

To validate the benchmark, we plan to register a known non-conservative model (an ORB-v3 direct variant) as a positive control, confirming that λ clearly detects non-conservative forces when present.

ellalbolland:

- [X] I have added human-written tests for the new logic.

## Summary

Adds a "Jacobian Symmetry" physicality benchmark that tests whether a model's predicted forces are conservative (derivatives of a single potential energy). The force Jacobian is built by central finite differences and the antisymmetric fraction of its Frobenius norm, lambda = ||J_anti|| / ||J||, is reported (mean and max over 10 diverse structures). lambda = 0 for perfectly conservative forces. Based on Bigi, Langer & Ceriotti, arXiv:2412.11569.

## Linked issue

Resolves #712

## Progress

- [X] Calculations
- [X] Analysis
- [X] Application
- [X] Documentation

## Testing

Tested on all 21 registered models

## New decorators/callbacks

None

## AI usage disclosure

I developed this benchmark with some assistance from Claude Code. I have personally reviewed, tested, and understand all of the code.


docs/source/user_guide/benchmarks/physicality.rst
```diff
@@ -213,3 +213,48 @@ Data availability
...
+-------
+
+Performance in predicting conservative forces, by measuring the antisymmetric fraction
+of the finite-difference force Jacobian [1].
+
+
+Metrics
+-------
+
+1. Mean/max lambda
```
joehart2001: would you be able to write these separately as 1 and 2 just to be consistent with the format

ml_peg/app/physicality/jacobian_symmetry/app_jacobian_symmetry.py
joehart2001, line 1: it would be great to have some form of interactivity here, do you have any ideas what you'd like to do? one idea would be a bar chart (bar for each molecule), like in the water slab dipoles, which you can click on to see the structures

docs/source/user_guide/benchmarks/physicality.rst
ellalbolland: Yes of course.

ml_peg/calcs/physicality/jacobian_symmetry/calc_jacobian_symmetry.py
```diff
@@ -0,0 +1,127 @@
...
+import os
+
+# ORB models compile their forward pass with torch.compile/dynamo. Unlike
+# other benchmarks, this one cycles through structures with many different
+# atom counts and immediately hammers each freshly-compiled graph with up to
+# 2*3N forward calls, which most likely triggers a low-level SIGABRT inside
+# torchinductor's dlopen'd kernels. Disabling compilation avoids it.
+# torch._dynamo.config reads this env var at import time, so it must be set
+# before torch (or anything importing it) is loaded anywhere in the process.
+os.environ["TORCH_COMPILE_DISABLE"] = "1"
```
joehart2001: ive managed to get around this issue by changing `frame.calc = copy(calc)` to `frame.calc = calc`. you can then do `frame.calc = None` before appending the frame, like in S30L/plf547 benchmarks

joehart2001:
Thanks for the PR! just left a few comments for now, mostly minor

ml_peg/calcs/physicality/jacobian_symmetry/calc_jacobian_symmetry.py
joehart2001: opinions on this @ElliottKasoar, i think this is safe to do?

ml_peg/app/physicality/jacobian_symmetry/app_jacobian_symmetry.py
ellalbolland: Went with the bar chart idea: click a model's cell → bar chart of λ per structure, then click a bar → a heatmap of that structure's antisymmetric Jacobian showing where the symmetry breaks. Also added N/A markers for structures a model can't run.

ml_peg/calcs/physicality/jacobian_symmetry/calc_jacobian_symmetry.py
ElliottKasoar: Sorry, I missed this.

Generally you just have to be careful about how results are stored, although I think in the past calculator caching has got in the way in unexpected situations...

I believe if you set values to the info e.g. `struct.info["energy"] = struct.get_potential_energy()`, which we do in some examples, then it should be fine to reuse the calculator without `copy`, as long as you also set `struct.calc =  None` before `ase.io.write` (else it tries to write out the same key twice).

It is also at least usually quite obvious if it's not working.

ElliottKasoar: I think the current handling is basically this (with forces, so storing to `struct.array`), so should be fine

ml_peg/calcs/physicality/jacobian_symmetry/calc_jacobian_symmetry.py
```diff
@@ -0,0 +1,123 @@
...
+    mlip
+        Name of model use and model to get calculator.
+    """
+    model_name, model = mlip
+    calc = model.get_calculator(precision="high")
+
+    write_dir = OUT_PATH / model_name
+    write_dir.mkdir(parents=True, exist_ok=True)
+
+    for struct_name, struct in STRUCTURES.items():
```
ElliottKasoar: It might be nice to have tqdm here since it's fast enough to run locally, but could still be a few minutes for larger models.

ml_peg/analysis/physicality/jacobian_symmetry/analyse_jacobian_symmetry.py
ElliottKasoar, line 1: One thing to be careful of here is, similar to #682, if a model failed to a calculation for one of the structures, it should not receive a score, but currently does.

The current handling with the plots etc is nice though and most of the way there

ml_peg/app/physicality/jacobian_symmetry/app_jacobian_symmetry.py
ElliottKasoar: This looks nice to me! I'm a little confused by the μ appearing in a few places though?

ElliottKasoar (commented):
A few more minor suggestions/questions, but this is looking great!

ellalbolland:
@ElliottKasoar There's no μ in the code, plotly just formats small ticks with SI prefixes, so 2e-6 renders as "2μ". Will change it to proper scientific notation.

ml_peg/analysis/physicality/jacobian_symmetry/analyse_jacobian_symmetry.py
ElliottKasoar: Actually I think this is ok for now - we catch errors during the calculation, and it's a very quick test, so this shouldn't really be a problem (and no more so than many other tests...).

(With #45 we may aim for the greater level of robustness)

ml_peg/calcs/physicality/jacobian_symmetry/calc_jacobian_symmetry.py
ellalbolland: Added, thanks!

ml_peg/analysis/physicality/jacobian_symmetry/analyse_jacobian_symmetry.py
ellalbolland: Fixed it anyway in the latest push, models failing any structure now get no score, and the per-structure plots still show N/A for the failed ones.

joehart2001:
thanks @ellalbolland i think this is looking really nice now. One thing i would add is maybe a few words to the description which shows on the app in more layman's terms. e.g. "Fraction of the force Jacobian's antisymmetric component, across 10 diverse structures which shows us...", imagine someone does not know what this means and what the implications are

ElliottKasoar (approved):
Thanks for this and the updates, it looks great!

PR #717, opened 2026-07-22
Add materials discovery evaluation metrics

janosh:
## Summary
- port Matbench Discovery classification, regression, and subset metrics
- add typed schemas and local CSV/JSONL artifact helpers
- require explicit unrounded prevalence for leaderboard-compatible DAF scoring

## Test plan
- [x] `pytest -q tests/metrics/test_discovery_metrics.py` (50 passed)
- [x] Ruff and diff checks


joehart2001:
I've pushed:
- the analysis file needed to build ml-peg tables
- app file wit table and density plots

In terms of data, the suggestion from @janosh was to work with local files for now (which we can download from matbench figshare online) and then later add a downloader/cache for a pinned matbench reference release so runs stay reproducible. 

@ElliottKasoar, i've tested and seems to be working well, what do you think about data?

PR #730, opened 2026-07-23
feat: add protein folding stability benchmark

### LINKED ISSUE #725: New benchmark: protein folding
### Summary

This benchmark evaluates the ability of a machine-learned interatomic potential (MLIP) to maintain the structural integrity of experimentally determined protein conformations during molecular dynamics (MD) simulations.

### Interactive features

_No response_

### Category

Molecular Dynamics

### Data availability

Input structures available in the MLIP Audit benchmark suite (or Huggingface page).

### Computational cost

_No response_

### Additional context

_No response_

lwalew:

## Summary

Adds the **protein folding stability** benchmark, migrated from the MLIP Audit suite. For a small set of proteins (chignolin, tryptophan cage, and an orexin/hypocretin fragment) an NVT molecular dynamics simulation is run at 300 K starting from the native folded reference conformation, and the model's ability to keep each protein folded is measured along the trajectory.

Reported metrics:

- **RMSD** — C-alpha RMSD from the native reference structure, averaged over the trajectory and across proteins (lower is better).
- **TM score** — TM score of each frame against the native reference, averaged over the trajectory and across proteins (closer to 1 is better).
- **Rgyr deviation** — maximum absolute deviation of the radius of gyration from the initial folded state, across proteins (lower is better).

The interactive app shows a line plot of the RMSD from the reference structure along the trajectory, averaged across proteins, for each model.

This benchmark introduces a new top-level **`biomolecules`** category (app tab, category YAML, and docs page). It also adds the shared **mlipaudit** wiring (optional extra, `grace` conflict entry, and `[tool.uv.sources]` git pin to the `mlpeg-migration` branch) plus the `mlip_audit` framework entry. This wiring is shared with other in-flight MLIP Audit migration PRs (e.g. ring planarity, water RDF, protein sampling), so overlap on `pyproject.toml`, `frameworks.yml`, and the `biomolecules` category files is expected and will be resolved at merge.

## Linked issue

Resolves #725

## Progress

- [x] Calculations
- [x] Analysis
- [x] Application
- [x] Documentation

## Testing

End-to-end model runs are pending the upload of the input data to S3 (`inputs/biomolecules/protein_folding_stability/protein_folding_stability.zip`). All `pre-commit` hooks (ruff check, ruff format, numpydoc-validation, file checks) pass locally, and the new modules compile cleanly. The implementation follows the established MLIP Audit migration pattern (see the ring planarity and water RDF benchmarks).

## New decorators/callbacks

None. The RMSD trajectory line plot reuses the existing `@plot_scatter` decorator (`show_line=True`, `show_markers=False`) and the `plot_from_table_column` callback.

### DIFF
--- a/docs/source/user_guide/benchmarks/biomolecules.rst
+++ b/docs/source/user_guide/benchmarks/biomolecules.rst
@@ -1,0 +1,58 @@
+============
+Biomolecules
+============
+
+Protein folding stability
+=========================
+
+Summary
+-------
+
+Performance in keeping small proteins folded during molecular dynamics. For each
+protein, an NVT molecular dynamics simulation is run at 300 K starting from the
+native (folded) reference conformation, and the ability of the model to retain the
+fold is measured along the trajectory. The benchmark uses a small set of well
+characterised proteins (chignolin, tryptophan cage, and an orexin/hypocretin
+fragment) with experimental reference structures.
+
+Metrics
+-------
+
+1. RMSD
+
+The root mean square deviation of the C-alpha atoms from the native reference
+structure is computed for each frame of the trajectory, then averaged over the
+trajectory and across all proteins. A lower RMSD indicates the fold is retained.
+
+2. TM score
+
+The TM score of each trajectory frame against the native reference structure is
+computed, then averaged over the trajectory and across all proteins. A TM score
+closer to 1 indicates that the global fold is preserved.
+
+3. Radius of gyration deviation
+
+The maximum absolute deviation of the radius of gyration from the initial folded
+state along the trajectory, taken across all proteins. A lower value indicates the
+protein does not unfold or collapse.
+
+A line plot shows the RMSD from the reference structure along the trajectory,
+averaged across the proteins, for each model.
+
+Computational cost
+------------------
+
+High: one MD simulation per protein. Faster inference can be achieved using the
+jax-accelerated simulations in MLIP Audit directly.
+
+Data availability
+-----------------
+
+Input structures:
+
+* MLIP Audit benchmark suite, InstaDeep. Native reference structures taken from the
+  Protein Data Bank (chignolin 1UAO, tryptophan cage 2JOF, orexin-B 1CQ0).
+
+Reference data:
+
+* Experimental reference structures (X-ray and NMR) from the Protein Data Bank.
--- a/ml_peg/analysis/biomolecules/protein_folding_stability/metrics.yml
+++ b/ml_peg/analysis/biomolecules/protein_folding_stability/metrics.yml
@@ -1,0 +1,22 @@
+metrics:
+  RMSD:
+    good: 0.0
+    bad: 2.0
+    unit: Å
+    weight: 1
+    tooltip: Root mean square deviation of the C-alpha atoms from the native reference structure, averaged over the trajectory and across all proteins. Lower is better.
+    level_of_theory: Experiment
+  TM Score:
+    good: 1.0
+    bad: 0.5
+    unit: null
+    weight: 1
+    tooltip: TM score of each trajectory frame against the native reference structure, averaged over the trajectory and across all proteins. Closer to 1 indicates the fold is retained.
+    level_of_theory: Experiment
+  Rgyr Deviation:
+    good: 0.0
+    bad: 2.0
+    unit: Å
+    weight: 1
+    tooltip: Maximum absolute deviation of the radius of gyration from the initial folded state along the trajectory, taken across all proteins. Lower is better.
+    level_of_theory: Experiment
--- a/ml_peg/calcs/biomolecules/protein_folding_stability/calc_protein_folding_stability.py
+++ b/ml_peg/calcs/biomolecules/protein_folding_stability/calc_protein_folding_stability.py
@@ -1,0 +1,76 @@
+"""
+Assess protein folding stability during molecular dynamics.
+
+A molecular dynamics simulation is run for each of a set of small proteins
+starting from their native (folded) conformation, and the ability of the model
+to keep each protein folded is measured along the trajectory via the RMSD,
+TM score, and radius of gyration relative to the reference structure.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any
+from warnings import warn
+
+from mlipaudit.benchmarks.folding_stability.folding_stability import (
+    STRUCTURE_NAMES,
+    FoldingStabilityModelOutput,
+)
+from mlipaudit.io import write_model_output_to_disk
+import pytest
+
+from ml_peg.calcs.utils.mlipaudit import MlPegFoldingStabilityBenchmark
+from ml_peg.calcs.utils.utils import download_s3_data
+from ml_peg.models import current_models
+from ml_peg.models.get_models import load_models
+
+MODELS = load_models(current_models)
+
+OUT_PATH = Path(__file__).parent / "outputs"
+
+
+@pytest.mark.parametrize("mlip", MODELS.items())
+def test_protein_folding_stability(mlip: tuple[str, Any]) -> None:
+    """
+    Benchmark protein folding stability during MD.
+
+    Parameters
+    ----------
+    mlip
+        Name of model and model object to get calculator.
+    """
+    model_name, model = mlip
+    calc = model.get_calculator()
+    calc = model.add_d3_calculator(calc)
+
+    data_input_dir = download_s3_data(
+        key="inputs/biomolecules/protein_folding_stability/protein_folding_stability.zip",
+        filename="protein_folding_stability.zip",
+    )
+
+    benchmark = MlPegFoldingStabilityBenchmark(
+        force_field=calc,
+        data_input_dir=data_input_dir,
+        run_mode="standard",
+    )
+    try:
+        benchmark.run_model()
+    except Exception as exc:
+        warn(
+            f"Error running protein folding stability benchmark for "
+            f"{model_name}: {exc}",
+            stacklevel=2,
+        )
+        # Structures with a ``None`` simulation state are treated as failed by
+        # analyze(), so this yields a failed result for every structure.
+        benchmark.model_output = FoldingStabilityModelOutput(
+            structure_names=list(STRUCTURE_NAMES),
+            simulation_states=[None] * len(STRUCTURE_NAMES),
+        )
+
+    write_model_output_to_disk(
+        MlPegFoldingStabilityBenchmark.name,
+        benchmark.model_output,
+        OUT_PATH / model_name,
+    )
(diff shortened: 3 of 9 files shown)


lwalew:
Hi @joehart2001 , as we have some benchmarks that reuse the same calculations, is this something that you now support? It says in the docs that this will soon be possible using fixtures, is this already the case?

joehart2001:
> Hi @joehart2001 , as we have some benchmarks that reuse the same calculations, is this something that you now support? It says in the docs that this will soon be possible using fixtures, is this already the case?

hi @lwalew, could you elaborate slightly more, do you mean the calculation output or the calculation procedure?

lwalew:
> > Hi @joehart2001 , as we have some benchmarks that reuse the same calculations, is this something that you now support? It says in the docs that this will soon be possible using fixtures, is this already the case?
> 
> hi @lwalew, could you elaborate slightly more, do you mean the calculation output or the calculation procedure?

I'm referring to from the [docs](https://arc.net/l/quote/wknusjzw):

> Future examples will also demonstrate the use of fixtures, which allow calculations such as relaxation to be reused by multiple tests within the module.

In this specific case, we have both benchmarks `protein folding stability` #730 and `protein sampling` #729 which both reuse the same input systems/calculations for their analysis.

joehart2001:
> > > Hi @joehart2001 , as we have some benchmarks that reuse the same calculations, is this something that you now support? It says in the docs that this will soon be possible using fixtures, is this already the case?
> > 
> > 
> > hi @lwalew, could you elaborate slightly more, do you mean the calculation output or the calculation procedure?
> 
> I'm referring to from the [docs](https://arc.net/l/quote/wknusjzw):
> 
> > Future examples will also demonstrate the use of fixtures, which allow calculations such as relaxation to be reused by multiple tests within the module.
> 
> In this specific case, we have both benchmarks `protein folding stability` #730 and `protein sampling` #729 which both reuse the same input systems/calculations for their analysis.

so we dont have anything in place for this yet, but you could point both analysis files to a single calc/output. thoughts @ElliottKasoar ?

ElliottKasoar:
> > > > Hi @joehart2001 , as we have some benchmarks that reuse the same calculations, is this something that you now support? It says in the docs that this will soon be possible using fixtures, is this already the case?
> > > 
> > > 
> > > hi @lwalew, could you elaborate slightly more, do you mean the calculation output or the calculation procedure?
> > 
> > 
> > I'm referring to from the [docs](https://arc.net/l/quote/wknusjzw):
> > > Future examples will also demonstrate the use of fixtures, which allow calculations such as relaxation to be reused by multiple tests within the module.
> > 
> > 
> > In this specific case, we have both benchmarks `protein folding stability` #730 and `protein sampling` #729 which both reuse the same input systems/calculations for their analysis.
> 
> so we dont have anything in place for this yet, but you could point both analysis files to a single calc/output. thoughts @ElliottKasoar ?

There are a couple of options depending on exactly what you want to do. There's nothing stopping you using fixtures in the way the docs mention if it makes sense.

This is already effectively what happens in the case of the [Li diffusion benchmarks in NEBs](https://github.com/ddmms/ml-peg/blob/main/ml_peg/calcs/nebs/li_diffusion/calc_li_diffusion.py#L25). It's a bit hidden since the "different tests" are in the same file, but the two different NEB paths each share an endpoint, which is relaxed before both.

The same would work between two different files, but you only see benefits when you run the benchmarks as part of a single `ml_peg calc` call, so its uses are quite limited. The Li diffusion example is also not the best since it currently interlinks the models more than I'd like, since it means we don't complete the combined geometry optimisation+NEB sequentially for each model.

Checking for the existence of files is probably the simplest and maybe best immediate solution for most cases? I'm not sure we have any examples of this currently across different benchmarks, but it's what we do for the [phonon reference data](https://github.com/ddmms/ml-peg/blob/main/ml_peg/calcs/bulk_crystal/phonons/calc_phonons.py#L196). I think you could either pick one as the source of truth, or if they're relatively small files, perhaps check and copy over the data if necessary?

The other related thing is we do have various tests (PLF547/PLA15, GSCDB138 benchmarks where we set up shared utility functions, since the tests are essentially the same in these cases, but the inputs for these differ, so I think it's a slightly different case.

ElliottKasoar:
Also just to add, this now has a small conflict with the mlipaudit utility, if that can be fixed?

docs/source/user_guide/benchmarks/biomolecules.rst
```diff
@@ -0,0 +1,59 @@
...
+
+The root mean square deviation of the C-alpha atoms from the native reference
+structure is computed for each frame of the trajectory, then averaged over the
+trajectory and across all proteins. A lower RMSD indicates the fold is retained.
+
+2. TM score
+
+The TM score of each trajectory frame against the native reference structure is
+computed, then averaged over the trajectory and across all proteins. A TM score
+closer to 1 indicates that the global fold is preserved.
```
ElliottKasoar, line 32:
```suggestion
The template modeling (TM) score of each trajectory frame against the native reference
structure is computed, then averaged over the trajectory and across all proteins. A TM
score closer to 1 indicates that the global fold is preserved.
```
This isn't necessarily something people are familiar with (I hadn't come across it, anyway)

docs/source/user_guide/benchmarks/biomolecules.rst
```diff
@@ -0,0 +1,59 @@
...
+state along the trajectory, taken across all proteins. A lower value indicates the
+protein does not unfold or collapse.
+
+A line plot shows the RMSD from the reference structure along the trajectory,
+averaged across the proteins, for each model.
+
+Computational cost
+------------------
+
+High: one MD simulation per protein. Faster inference can be achieved using the
```
ElliottKasoar, line 46: If you have an estimate for GPU hours for mace-mp-0a, this would be useful here. If not, we can come back to it

ml_peg/analysis/biomolecules/protein_folding_stability/metrics.yml
```diff
@@ -0,0 +1,22 @@
...
+    weight: 1
+    tooltip: TM score of each trajectory frame against the native reference structure, averaged over the trajectory and across all proteins. Closer to 1 indicates the fold is retained.
+    level_of_theory: Experiment
+  Rgyr Deviation:
+    good: 0.0
+    bad: 2.0
+    unit: Å
+    weight: 1
+    tooltip: Maximum absolute deviation of the radius of gyration from the initial folded state along the trajectory, taken across all proteins. Lower is better.
+    level_of_theory: Experiment
```
ElliottKasoar, line 22:
```suggestion
    level_of_theory: Experimental
  TM Score:
    good: 1.0
    bad: 0.5
    unit: null
    weight: 1
    tooltip: TM score of each trajectory frame against the native reference structure, averaged over the trajectory and across all proteins. Closer to 1 indicates the fold is retained.
    level_of_theory: Experimental
  Rgyr Deviation:
    good: 0.0
    bad: 2.0
    unit: Å
    weight: 1
    tooltip: Maximum absolute deviation of the radius of gyration from the initial folded state along the trajectory, taken across all proteins. Lower is better.
    level_of_theory: Experimental
```
We need to document these better still, dory. `Experimental` is effectively a keyword

ml_peg/calcs/biomolecules/protein_folding_stability/calc_protein_folding_stability.py
```diff
@@ -0,0 +1,99 @@
...
+    Benchmark protein folding stability during MD.
+
+    Parameters
+    ----------
+    mlip
+        Name of model and model object to get calculator.
+    """
+    model_name, model = mlip
+    calc = model.get_calculator(precision="low")
+    calc = model.add_d3_calculator(calc)
```
ElliottKasoar, line 59: Just to confirm, do we want to keep this? I would imagine they are somewhat important, so probably, but then we may need to recalculate a few models e.g. mace-mp-0a, if I understood correctly?

ElliottKasoar (commented):
Thanks for this, @lwalew!

I'm doing a quick check of the analysis with the data you've shared, but otherwise I are just have a few very minor suggestions/questions, and I think we should be good to merge.

ElliottKasoar:
Hi @lwalew, just comparing our results to what it currently live on the mlipaudit leaderboard, it seems that the TM score and RMSD match, but there's a substantial difference in the maximum radius of gyration. Do you know why this might be? I also wondered if there is a reason not to include the DSSP match too?

image

image

ml_peg/analysis/biomolecules/protein_folding_stability/metrics.yml
mcarobene-id: Dropped `TM score` and `Rgyr Deviation`. Changed `RMSD` to use `Experimental` rather than `Experiment` ✅

docs/source/user_guide/benchmarks/biomolecules.rst
mcarobene-id: Have dropped `TM score` so removed this metric description

ml_peg/calcs/biomolecules/protein_folding_stability/calc_protein_folding_stability.py
mcarobene-id: Yes, as these systems are all small proteins in water. And yes, this means that we need to rerun with mace-mp with this term.

As discussed on our call, we are finding that all models which require an explicit D3 term will not complete all 3 simulations within our internal timeout of 7 days, whereas the models we're running without a D3 term will all complete. Hoping to look into why this is and if the D3 term can be made faster

mcarobene-id:
Hi @ElliottKasoar , thanks for highlighting the discrepancy between the metrics shown in ml-peg and those on the MLIPAudit public leaderboard. The reason for the Rg discrepancy was just that the aggregation over the 3 systems didn't agree (mean / max). As discussed on our call on Monday, only the RMSD is used to compute the score reported for each model. We decided to keep scoring-only metrics for this benchmark so have removed the TM and Rg deviation metrics.

This PR is ready for another review whenever you're ready 🙏

docs/source/user_guide/benchmarks/biomolecules.rst
mcarobene-id: Due to the dispersion term and the sizes of these systems, we are finding that all models we're running with a dispersion term are set to take >7 days (our internal timeout) to run all 3 systems on an 80GB H100 GPU, including `mace-mp-0a`.

I will need to go find runtimes for the models we ran previously (`mace-off23-m`, `uma-s-1p1-omol`, `mace-polar-1-m`) but here are the runtimes we are getting for recent runs:

* `orb-v3-consv-omol`: 40hrs
* `mace-polar-1-s`: 76hrs
* `mace-mh-1-omol`: 84hrs
* `mace-omol`: 82hrs (note: only ran the first 2 systems - the 3rd failed immediately due to an OOM error)

Happy to include any info like this if helpful

PR #751, opened 2026-07-26
Add framework summary tables on each page + overall framework summary

### LINKED ISSUE #733: Add summary tables for frameworks
### Summary

Similarly to categories, we should have summary tables per framework, and a summary table of frameworks.

### Example

_No response_

### Alternatives

_No response_

### Additional context

_No response_

joehart2001:
… on main page

## Summary

- framework table on each framework page which collects all the framework tests
- summary framework table on main page, collecting all frameworks
- have tested updating weights and thresholds and score propagation seems to be working as expected
- dedicated build framework file to reduce bloat again
- claude code for review, fixing precommit and bug finding

## Linked issue

Resolves #

### DIFF
--- a/ml_peg/app/utils/build_frameworks.py
+++ b/ml_peg/app/utils/build_frameworks.py
@@ -33,19 +33,12 @@ class CategoryView(TypedDict):
     tests: list[CategoryTest]
 
 
-class CategoryGroup(TypedDict):
-    """Benchmarks of one category, grouped for display on a framework page."""
-
-    category: str
-    tests: list[Div]
-
-
 class FrameworkView(TypedDict):
     """Data needed to build a single framework page."""
 
     framework_id: str
     label: str
-    category_groups: list[CategoryGroup]
+    benchmarks_by_category: dict[str, list[Div]]
     summary_table: NotRequired[DataTable]
     weight_components: NotRequired[Div]
 
@@ -74,7 +67,7 @@ def build_framework_views(
         if framework_id == "ml_peg":
             continue
 
-        category_groups = []
+        benchmarks_by_category: dict[str, list[Div]] = {}
         for category_name, category_view in category_views.items():
             tests = [
                 test["layout"]
@@ -82,13 +75,13 @@ def build_framework_views(
                 if framework_id in test["framework_ids"]
             ]
             if tests:
-                category_groups.append({"category": category_name, "tests": tests})
+                benchmarks_by_category[category_name] = tests
 
-        if category_groups:
+        if benchmarks_by_category:
             framework_views[framework_id] = {
                 "framework_id": framework_id,
                 "label": get_framework_config(framework_id)["label"],
-                "category_groups": category_groups,
+                "benchmarks_by_category": benchmarks_by_category,
             }
 
     return framework_views
@@ -109,8 +102,8 @@ def build_framework_summary_tables(
     all_frameworks
         Framework ids for each benchmark, grouped by category.
     framework_views
-        Framework page metadata (keys are the frameworks to build tables for;
-        ``ml_peg`` and empty frameworks are already excluded).
+        Framework page metadata, keyed by the frameworks to build tables for.
+        ``ml_peg`` and empty frameworks are already excluded.
 
     Returns
     -------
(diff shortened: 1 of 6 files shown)


ml_peg/app/utils/build_frameworks.py
```diff
@@ -0,0 +1,177 @@
...
+        Category metadata including benchmark layout components.
+    framework_ids
+        All framework IDs discovered from benchmark apps.
+
+    Returns
+    -------
+    dict[str, dict[str, object]]
+        Mapping of framework ID to grouped benchmark layouts by category.
+    """
+    framework_views: dict[str, dict[str, object]] = {}
```
ElliottKasoar:
```suggestion
    framework_views: dict[str, dict[str, Any]] = {}
```
I wouldn't normally use `object` for a type hint. `Any` isn't great either, so if we can be more specific, that would be better

ml_peg/app/utils/build_frameworks.py
```diff
@@ -0,0 +1,177 @@
...
+
+    Parameters
+    ----------
+    all_tables
+        Benchmark tables grouped by category.
+    all_frameworks
+        Framework ids for each benchmark, grouped by category.
+    framework_views
+        Framework page metadata (keys are the frameworks to build tables for;
+        ``ml_peg`` and empty frameworks are already excluded).
```
ElliottKasoar: Values are..? `object` again is unhelpful

ml_peg/app/utils/build_frameworks.py
```diff
@@ -0,0 +1,177 @@
...
+            ]
+            if tests:
+                category_groups.append({"category": category_name, "tests": tests})
+
+        if category_groups:
+            framework_views[framework_id] = {
+                "framework_id": framework_id,
+                "label": get_framework_config(framework_id)["label"],
+                "category_groups": category_groups,
+            }
```
ElliottKasoar: So `category_groups` is a list of {category: [list of tests in category]} ? Two questions: 

(1) Why not just a straight dict, rather than the list of dicts? (If it's relevant, I may have misled you in the past, as while order of dicts wasn't preserved in the past, any Python versions we use should preserve insertion order, but I think we'll sort things later anyway)

(2) Why do we still need things grouped by category? Presumably not for the tables, but is it so we know which weights to update?

ml_peg/app/utils/build_frameworks.py
```diff
@@ -0,0 +1,177 @@
...
+        ``ml_peg`` and empty frameworks are already excluded).
+
+    Returns
+    -------
+    tuple[dict[str, DataTable], dict[str, dict[str, DataTable]]]
+        Per-framework summary tables keyed by framework id, and the
+        {framework_id: {benchmark_name: benchmark_table}} grouping.
+    """
+    # Local import avoids a circular import (build_app imports this module).
+    from ml_peg.app.build_app import build_summary_table
```
ElliottKasoar: If we move that to a separate module too it might remove the need for this?

ml_peg/app/utils/build_frameworks.py
ElliottKasoar: Ah for (2) you also want to order the tests by category name, not just test name?

joehart2001: (1) yep good call here, done

(2) grouping isnt for weights, it's just used only to lay out the framework page: one heading per category with that framework's benchmarks beneath it. ive renmaed as it was confusing

ml_peg/app/utils/build_frameworks.py
joehart2001: have added now

ml_peg/app/utils/build_frameworks.py
```diff
@@ -0,0 +1,205 @@
+"""Framework page and summary-table builders for the ML-PEG app."""
+
+from __future__ import annotations
+
+from typing import NotRequired, TypedDict
```
ElliottKasoar: I think `NotRequired` may only have been added in 3.11 so we may not be able to do this while supporting 3.10

joehart2001: would this not be caught by the ci tests or not?

ElliottKasoar: Ideally yes, but I don't think currently - none of the CI currently attempts to build the app. I just tried with 3.10 and I do get:

```
ImportError: cannot import name 'NotRequired' from 'typing' (/Users/elliottkasoar/.local/share/uv/python/cpython-3.10.16-macos-aarch64-none/lib/python3.10/typing.py)
```

joehart2001:
@ElliottKasoar think its all good now

ElliottKasoar (approved):
Looks great, thanks!

PR #757, opened 2026-07-28
Enable float64 precision for FairChem (UMA) calculators

### LINKED ISSUE #756: FairChemCalc ignores precision; UMA models run in float32
### Summary

Silent correctness bug, no traceback. The output below is the observed vs expected result.

`FairChemCalc.get_calculator` ignores the `precision` argument the rest of ML-PEG passes, so UMA/FairChem models always run in FairChem's default float32, even with `precision="high"`. Sibling calculators (`MaceCalc`, `PetMadCalc`, `OrbCalc`) already honour `precision` via a `precision_map`; `FairChemCalc` does not.

This matters for finite-difference benchmarks: float32 force noise is amplified by the finite-difference step, inflating metrics like the Jacobian antisymmetry (lambda) for otherwise-conservative models.

Reproducible example:

    import numpy as np
    from ase.build import molecule

    model = load_models(current_models)["uma-s-1p1-omat"]
    w = molecule("H2O"); w.info["charge"] = 0; w.info["spin"] = 1

    w.calc = model.get_calculator(precision="high"); f_high = w.get_forces()
    w.calc = model.get_calculator(precision="low");  f_low = w.get_forces()
    print(np.abs(f_high - f_low).max())   # -> 0.0

Expected: `f_high` (float64) differs from `f_low` (float32).
Actual: identical, both float32.

### Platform

macOS (Darwin 25.5.0), arm64

### Python version

Python 3.11.0

ellalbolland:

- [ ] I have added human-written tests for the new logic.

## Summary

FairChemCalc.get_calculator` ignored the `precision` argument that the rest of ML-PEG passes, so UMA/FairChem models always ran in FairChem's default float32, even with `precision="high"`. 

This makes `FairChemCalc` honour `precision` the same way the sibling calculators (`MaceCalc`, `PetMadCalc`, `OrbCalc`) already do: `precision` is mapped to a dtype (`high → float64`, `low → float32`) and applied by replacing `base_precision_dtype` on FairChem's default `InferenceSettings`, which is passed to `get_predict_unit`. `default_dtype` becomes an optional per-model override instead of a hardcoded float32.

## Linked issue

Resolves #756 

## Testing

Verified manually: for uma-s-1p1 on H₂O, precision="high" and precision="low" now give different forces (max abs diff ~1.9e-6), whereas before the fix they were bit-identical (both float32).

### DIFF
--- a/ml_peg/models/models.py
+++ b/ml_peg/models/models.py
@@ -216,17 +216,19 @@ class FairChemCalc(SumCalc):
     model_name: str
     task_name: str
     device: Device | str = "cpu"
-    default_dtype: str = "float32"
+    default_dtype: str | None = None
     overrides: dict = dataclasses.field(default_factory=dict)
 
-    def get_calculator(self, **kwargs) -> Calculator:
+    def get_calculator(self, precision="high", **kwargs) -> Calculator:
         """
         Prepare and load the calculator.
 
         Parameters
         ----------
+        precision
+            Level of precision to evaluate the model.
         **kwargs
-            Unused additional keyword arguments.
+            Any additional keyword arguments.
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


ElliottKasoar (approved):
This look great, thanks @ellalbolland!

PR #761, opened 2026-07-28
fix weas highlight follow bug for density plots 

joehart2001:

## Summary

- issue: When you click a density-scatter point that reveals a list of structures and step through them in WEAS, the highlighted scatter point incorrectly moves with each frame. The frame-follow handler assumed a one-to-one mapping between trajectory frames and scatter points—valid for a shared NEB path, but not for the structures grouped behind a density-cell point.

- Fix: Keep the highlight fixed on the originally clicked density point while stepping through its structures, while retaining frame-following for shared NEB trajectory files.

## Linked issue

Resolves #

## Testing
elasticity density

### DIFF
--- a/ml_peg/app/utils/build_callbacks.py
+++ b/ml_peg/app/utils/build_callbacks.py
@@ -174,7 +174,10 @@ def plot_from_table_cell(
         return Div(TABLE_HINT, style=INSTRUCTION_STYLE), None
 
 
-_HIGHLIGHTED_SCATTERS: dict[str, bool] = {}
+# A scatter can be passed to more than one helper, but its highlight callback
+# should only be registered once. Store the first frame-follow setting so later
+# registrations can be skipped or rejected when they disagree.
+registered_highlights: dict[str, bool] = {}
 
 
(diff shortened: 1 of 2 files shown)


ml_peg/app/utils/build_callbacks.py
```diff
@@ -174,10 +174,10 @@ def show_plot(active_cell, current_table_data) -> Div:
         return Div(TABLE_HINT, style=INSTRUCTION_STYLE), None
 
 
-_HIGHLIGHTED_SCATTERS: set[str] = set()
+_HIGHLIGHTED_SCATTERS: dict[str, bool] = {}
```
ElliottKasoar: I know this isn't new, but why is this set up as a constant? It looks like it's only used within one function, and it's also modified?

joehart2001: Yep, you’re right that it isn’t a constant. It needs to be module-level because several helpers may register highlighting for the same scatter. I’ve renamed it and added a comment explaining this.

ElliottKasoar (approved):

PR #765, opened 2026-07-29
Enable float64 precision for MatterSim calculators

### LINKED ISSUE #764: Enable float64 precision for MatterSim calculators
### Summary

Currently MatterSim always runs in float32, even when `precision="high"` is requested. It is the only backend that cannot run in double precision. 

The cause is the pinned version: `mattersim==1.2.2` has no dtype/float64 support. float64 was added in MatterSim 1.2.4 (released 11 May 2026), which requires Python >= 3.12.

This matters for finite-difference physicality benchmarks (e.g. Jacobian antisymmetry): in float32, otherwise-conservative models are noise-limited, so the metric reflects precision noise rather than true model behaviour.

Desired behaviour: MatterSim honours `precision="high"` and runs in float64, consistent with the other models.

### Example

With the proposed change, get_calculator maps precision to MatterSim's dtype:

    model.get_calculator(precision="high")   # currently float32; would become float64

Verified locally on a branch, for mattersim-v1.0.0-5M on H2O:
- With the current pinned 1.2.2: `precision="high"` and `precision="low"` forces are
  identical (max abs diff 0.0) -> float32 regardless of precision.
- With 1.2.4 + the wrapper: they differ (max abs diff 1.4e-5) -> float64 engages.

### Alternatives

_No response_

### Additional context

_No response_

ellalbolland:

- [ ] I have added human-written tests for the new logic.

## Summary

Enables float64 for MatterSim (previously float32-only; it ignored `precision="high"`).

- Bump `mattersim==1.2.2` -> `1.2.4; python_version >= '3.12'` 
- Add a `MatterSimCalc` wrapper mirroring `PetMadCalc` (maps precision to `dtype`)

## Linked issue

Resolves #764 

## Testing

Tested in an isolated Python 3.12 venv with mattersim 1.2.4:

- `uv lock` resolves cleanly with the bumped pin (no dependency conflicts).
- `pre-commit run --all-files` passes (ruff, numpydoc).
- End-to-end via `load_models`: `MatterSimCalculator` routes to the new `MatterSimCalc`, and `precision="high"` runs in float64.
- Measured on mattersim-v1.0.0-5M / H2O: with the current 1.2.2, `high` and `low` forces are identical (max abs diff 0.0, float32 only); with 1.2.4 and the wrapper they differ (max abs diff 1.4e-5), confirming float64 engages.


ml_peg/models/models.py
```diff
@@ -126,6 +126,35 @@ def get_calculator(self, precision="high", **kwargs) -> Calculator:
...
+
+        Returns
+        -------
+        Calculator
+            Loaded ASE Calculator.
+        """
+        precision_map = {"low": "float32", "high": "float64"}
+        kwargs["dtype"] = precision_map[precision]
+
+        if self.default_dtype is not None:
```
ElliottKasoar, line 152: Just to note, if someone were to try to install mattersim separately to ml-peg with Python <= 3.11 (e.g. `pip install mattersim`), this would still run, but not set anything meaningful.

I think really this comes into something we want to store as part of #45 (e.g. the module version), so I think this is probably fine, since we've declared the requirement, but I wouldn't be surprised if it catches people out every now and then.

ElliottKasoar (approved):
Looks great, thanks @ellalbolland!

Issue #771, opened 2026-07-30
MatterSimCalculator.__setstate__ ignores saved model architecture, crashes on copy() for non-default checkpoints


ellalbolland:
### Summary

While testing MatterSim's new float64 precision support ([#764](https://github.com/ellalbolland/ml-peg/issues/764)/[#765](https://github.com/ellalbolland/ml-peg/issues/765)), copying a loaded MatterSimCalculator (via Python's copy.copy()) crashes for any non-default checkpoint (e.g. mattersim-5M). This affects any benchmark that reuses a loaded calculator across multiple structures via copy(calc) - a pattern used in ~25 calc scripts across this repo.

Minimal reproduction:
```
from copy import copy
from mattersim.forcefield import MatterSimCalculator

calc = MatterSimCalculator(load_path="mattersim-v1.0.0-5m", device="cpu")
copy(calc)

```

Full error output:
Traceback (most recent call last):
  File "<string>", line 5, in <module>
  File ".../copy.py", line 97, in copy
    return _reconstruct(x, None, *rv)
  File ".../copy.py", line 261, in _reconstruct
    y.__setstate__(state)
  File ".../mattersim/forcefield/potential.py", line 1270, in __setstate__
    self.potential.model.load_state_dict(model_state_dict)
  File ".../torch/nn/modules/module.py", line 2624, in load_state_dict
    raise RuntimeError(
RuntimeError: Error(s) in loading state_dict for M3Gnet:
[...]

PR #775, opened 2026-07-31
Fix MatterSimCalculator copy() crash on non-default checkpoints

ellalbolland:
## Pre-review checklist for PR author

PR author **must** check the checkboxes below when creating the PR.

- [X] I've confirmed the [contribution guidelines](https://github.com/ddmms/ml-peg/blob/main/contributing.md).
- [X] I have [reviewed and understand](https://ddmms.github.io/ml-peg/developer_guide/vibecoding.html#contributing-with-ai-the-vibe-coding-guide) all AI-generated code in this PR.
- [X] I have added human-written tests for the new logic.
- [X] I have properly cited any upstream algorithms or libraries the AI utilized.
- [X] I have disclosed significant AI tool usage in the PR description.

## Summary

Patches MatterSimCalculator.__setstate__ (mattersim's own code) to rebuild the model from its saved model_args on copy.copy(), instead of always defaulting to the wrong (1M/128-unit) architecture regardless of which checkpoint was actually loaded. Fixes a crash affecting any benchmark using the standard copy(calc) pattern with mattersim-5M (or any non-default MatterSim checkpoint).

## Linked issue

Resolves #771 

## Testing

Standalone (no ml-peg): confirmed copy.copy(MatterSimCalculator(load_path="mattersim-v1.0.0-5m")) crashes on plain mattersim==1.2.4, and that mattersim's own __setstate__ source never uses the model_args it saves (confirming root cause before fixing).
After the patch: copied calculator's parameters and predictions are bitwise-identical to the original (not just crash-free).
Verified before/after on three independent benchmarks - each fails identically without the fix, passes with it: translational_symmetry, extensivity, SBH17 

## AI tool usage disclosure
This investigation, root-cause diagnosis, and fix were developed with assistance from Claude (Anthropic)

PR #788, opened 2026-08-03
Fix silent float32 precision truncation in MatterSim graph construction

### LINKED ISSUE #786: Fix silent float32 precision truncation in MatterSim graph construction
### Summary

GraphConverter.convert (mattersim's own code) hardcodes torch.FloatTensor for atom_pos and cell, always float32, regardless of the calculator's requested dtype. calculate() upcasts afterward, but that is too late: precision already lost to float32 rounding cannot be recovered. This silently violates translational symmetry even at dtype="float64".

Minimal example (periodic structure, so the separate copy() bug in #771 cannot apply):

import numpy as np
from ase.build import graphene
from mattersim.forcefield import MatterSimCalculator

DIRECTION = np.array([37.2, -18.9, 24.6])
DIRECTION = DIRECTION / np.linalg.norm(DIRECTION)
calc = MatterSimCalculator(load_path="mattersim-v1.0.0-5m", device="cpu", dtype="float64")

original = graphene(formula="C2", a=2.46, thickness=0.0, vacuum=10.0)
original.calc = calc
e0, f0 = original.get_potential_energy(), original.get_forces()

translated = original.copy()
translated.translate(DIRECTION * 40.0)
translated.wrap()
translated.calc = calc
e1, f1 = translated.get_potential_energy(), translated.get_forces()

print(abs(e1 - e0) / len(original) * 1000, "meV/atom")
print(np.abs(f1 - f0).max() * 1000, "meV/A")

Output:
0.0 meV/atom
0.007912 meV/A

Expected: both exactly 0. Confirmed cause by rebuilding only atom_pos/cell at float64 instead of float32: delta_F drops to exactly 0.000000.

### Platform

macOS 26.5 arm64 (Darwin 25.5.0)

### Python version

Python 3.12.13

ellalbolland:

- [X] I have added human-written tests for the new logic.

## Summary

Patches GraphConverter.convert to rebuild atom_pos/cell at the calculator's actual requested dtype, instead of mattersim's hardcoded float32. The existing upcast in calculate() happens too late to recover lost precision, silently breaking translational symmetry even at dtype="float64".
Delegates to mattersim's own unmodified convert() for all the real logic, only rebuilding the two fields that lose precision. pbc_offsets is untouched, it's small integers, exact in float32 either way.

## Linked issue

Resolves #786 

## Testing

- Standalone: confirmed torch.FloatTensor ignores torch.set_default_dtype
- Controlled proof: graphene translated 40 A (periodic, so the separate copy() bug in #771 cannot apply). delta_F goes from ~0.008 meV/A (without fix) to exactly 0.000000 (with fix), verified both standalone and through the real ml-peg pipeline.
- Verified across five benchmarks with mattersim-1M (unaffected by #771): translational_symmetry, extensivity, SBH17, elemental_tm_vacancies, lattice_constants (11+ step relaxation, confirming correctness under repeated sequential calls, not just one).

## AI tool usage disclosure
This investigation, root-cause diagnosis, and fix were developed with assistance from Claude (Anthropic)


ElliottKasoar:
Thanks for this, @ellalbolland!

Would you be able to rebase/resolve the conflict now that #775 is merged, and also open an issue for this with MatterSim (it'll be useful for other people, and to discuss with their devs, but also, patches are always tricky to maintain for long)?

ellalbolland:
I actually don't think we should merge this. The fix only lives in ml-peg's models.py monkeypatch, not in MatterSim itself so our table numbers reflect a MatterSim that's better-behaved than what anyone gets from pip install MatterSim. Someone reproducing our test against a plain install would likely see a bigger float32-truncation error.

I will open an issue for this with MatterSim maintainers about the float32 truncation so it's tracked upstream too. Happy to rebase and resolve the conflict too, but wanted to flag the above before we merge. Let me know what you think.

ElliottKasoar:
> I actually don't think we should merge this. The fix only lives in ml-peg's models.py monkeypatch, not in MatterSim itself so our table numbers reflect a MatterSim that's better-behaved than what anyone gets from pip install MatterSim. Someone reproducing our test against a plain install would likely see a bigger float32-truncation error.
> 
> I will open an issue for this with MatterSim maintainers about the float32 truncation so it's tracked upstream too. Happy to rebase and resolve the conflict too, but wanted to flag the above before we merge. Let me know what you think.

I'm also a bit torn, and so far we've followed a fairly similar principle of taking packages at face value.

I'd certainly agree if it were a fix that changed the underlying performance of the model, but this is just a bug in the ASE wrapper, which feels like a bit more of a grey area.

Let's see how responsive they are once you open an issue. If it's fixed within a week then it's one less thing to decide.

joehart2001:
> > I actually don't think we should merge this. The fix only lives in ml-peg's models.py monkeypatch, not in MatterSim itself so our table numbers reflect a MatterSim that's better-behaved than what anyone gets from pip install MatterSim. Someone reproducing our test against a plain install would likely see a bigger float32-truncation error.
> > I will open an issue for this with MatterSim maintainers about the float32 truncation so it's tracked upstream too. Happy to rebase and resolve the conflict too, but wanted to flag the above before we merge. Let me know what you think.
> 
> I'm also a bit torn, and so far we've followed a fairly similar principle of taking packages at face value.
> 
> I'd certainly agree if it were a fix that changed the underlying performance of the model, but this is just a bug in the ASE wrapper, which feels like a bit more of a grey area.
> 
> Let's see how responsive they are once you open an issue. If it's fixed within a week then it's one less thing to decide.

agreed

ellalbolland:
Merged in main and resolved the conflict. Also opened an issue with MatterSim: [microsoft/mattersim#171](https://github.com/microsoft/mattersim/issues/171)

PR #793, opened 2026-08-04
Speed markers

joehart2001:

## Summary

- Add test speed markers to each test
- add a test speed summary on the main page, as well on framework pages
- more thorough marking of test speeds, where all tests are marked with fast, medium, slow, very slow, multi-day
- add ability to time tests
- Run fast and medium benchmarks by default

- requires **rerun of analysis**

## Linked issue

Resolves #

## Testing

Screenshot 2026-08-04 at 18 17 17

Screenshot 2026-08-04 at 18 17 39


ElliottKasoar:
From a UI perspective, the numbers (2 hours, 2 hours, 3 min etc.) aren't entirely obvious in terms of what they represent.

I'm particularly confused by the 3 min for the slow tests?

joehart2001:
> From a UI perspective, the numbers (2 hours, 2 hours, 3 min etc.) aren't entirely obvious in terms of what they represent.
> 
> I'm particularly confused by the 3 min for the slow tests?

so the NEB tests were marked as slow but they only took 1 min each on gpu, so have now moved to fast. i have also moved elasticity and phonons to slow category now as they are under 10h

ElliottKasoar:
> > From a UI perspective, the numbers (2 hours, 2 hours, 3 min etc.) aren't entirely obvious in terms of what they represent.
> > I'm particularly confused by the 3 min for the slow tests?
> 
> so the NEB tests were marked as slow but they only took 1 min each on gpu, so have now moved to fast. i have also moved elasticity and phonons to slow category now as they are under 10h

Some of these are little tricky since whenever there's an optimisation, there's sometimes more than two orders of magnitude difference in runtime between models, depending on how many steps they need (on top of the inference difference).

I agree the Li diffusion is probably still "fast" as is, but ideally we could take this into account a little more somehow, since it's quite different to singlepoint/MD tests where they're a bit more predictable.

joehart2001:
> > > From a UI perspective, the numbers (2 hours, 2 hours, 3 min etc.) aren't entirely obvious in terms of what they represent.
> > > I'm particularly confused by the 3 min for the slow tests?
> > 
> > 
> > so the NEB tests were marked as slow but they only took 1 min each on gpu, so have now moved to fast. i have also moved elasticity and phonons to slow category now as they are under 10h
> 
> Some of these are little tricky since whenever there's an optimisation, there's sometimes more than two orders of magnitude difference in runtime between models, depending on how many steps they need (on top of the inference difference).
> 
> I agree the Li diffusion is probably still "fast" as is, but ideally we could take this into account a little more somehow, since it's quite different to singlepoint/MD tests where they're a bit more predictable.

yeah i agree with this, but i think we just need to stay true to our baseline. its only rough guidance anyway and in the case of the NEB test, it took under a minute with mace-mp-0a, so the difference would only be a minute or so with different models. i know this can compound, but i think we just have to say "these are the results for X model on Y hardware, hopefully this is useful guidance but take it with a pinch of salt". im also sure we will come up with better ideas down the line, maybe when we have improved logging

joehart2001:
Screenshot 2026-08-04 at 22 02 26


heres it updated. for the 3 very slow tests, im running them now to see how long (low dimensional relaxation test etc)

i also think this format is much more clear about what everything means

joehart2001:
@ElliottKasoar any idea why these tests dont want to work?

ElliottKasoar:
> @ElliottKasoar any idea why these tests dont want to work?

image
Looks like it was probably some sort of GitHub issue. Looks ok now.

I'll try to take a look at this tomorrow, sorry the delay in reviewing.

joehart2001:
> > @ElliottKasoar any idea why these tests dont want to work?
> 
> image Looks like it was probably some sort of GitHub issue. Looks ok now.
> I'll try to take a look at this tomorrow, sorry the delay in reviewing.

thanks! no worries

ElliottKasoar:
@joehart2001 I'm having a look now but in the meantime could you rebase and resolve the conflicts, and add any missing markers from new benchmarks since you created this?

joehart2001:
> @joehart2001 I'm having a look now but in the meantime could you rebase and resolve the conflicts, and add any missing markers from new benchmarks since you created this?

done. also note, you need to run the analysis to get the labels

joehart2001:
@ElliottKasoar ok no more edits from me

conftest.py
```diff
@@ -6,11 +6,23 @@
...
 from ml_peg import models
 
 
+def _round_runtime_minutes(minutes: float) -> int | float:
+    """Round a measured runtime while retaining a one-minute overhead floor."""
+    if minutes < 1:
+        return 1
+    if minutes < 10:
+        return round(minutes, 1)
+    return round(minutes)
```
ElliottKasoar: Do you think we need to worry about 5.3 minutes vs 5 given all the factors that will influence this etc. (especially if we're already ignoring 0-1 min).

If not, we can just do `min(1, round(minutes))` for all values (and then probably just get rid of the function)

conftest.py
```diff
@@ -46,31 +64,68 @@ def pytest_addoption(parser):
             "comma-separated list of framework ids. Default is all tests."
         ),
     )
+    for level in ("fast", "medium"):
+        parser.addoption(
+            f"--{level}-only",
```
ElliottKasoar: I'm not sure we need (pytest) options for this. You can specify markers that must match with `-m` (see https://docs.pytest.org/en/stable/example/markers.html), so if we do want this option, I think we could just make this more user-friendly via the typer CLI inputs

conftest.py
```diff
@@ -46,31 +64,68 @@ def pytest_addoption(parser):
...
+            action="store_true",
+            help=f"Run only benchmarks marked {level}",
+        )
+    parser.addoption(
```
ElliottKasoar: Why are we defining this here, and not in the typer CLI?

conftest.py
```diff
@@ -90,3 +145,45 @@ def pytest_collection_modifyitems(config, items):
     if deselected:
         config.hook.pytest_deselected(items=deselected)
     items[:] = selected
+
+
+def pytest_terminal_summary(terminalreporter, exitstatus, config):
```
ElliottKasoar: Is there not a way to use `--durations` for timings?

docs/source/user_guide/benchmarks/bulk_crystal.rst
```diff
@@ -459,8 +459,8 @@ Same as (11), for the (110) cleavage plane. Reference: 30.0 GPa.
 Computational cost
 ------------------
 
-Medium: tests are likely to take minutes to run on GPU, or hours on CPU for each model.
-The benchmark is marked as slow and excluded from default test runs.
+Medium: tests are likely to take tens of minutes to run on GPU, or hours on CPU
+for each model. The benchmark is marked ``medium`` and runs by default.
```
ElliottKasoar: If it takes hours on CPU per model, we may not want to run by default?

I also probably wouldn't state whether they run by default here, since then we ought to for every single benchmark, so just saying the speed marking and having that clear should be enough?

docs/source/user_guide/benchmarks/conformers.rst
```diff
@@ -76,7 +76,8 @@ for which the model fails to produce an energy profile score 0.
 Computational cost
 ------------------
 
-Medium: Minutes on GPU. Minutes to tens of minutes on CPU.
+Low: minutes on GPU, and minutes to tens of minutes on CPU. The benchmark is
```
ElliottKasoar: Elsewhere you've replaced the equivalent of "low" with the marker?

ml_peg/app/build_app.py
```diff
@@ -379,6 +380,7 @@ def get_all_tests(
             )
             test_app = test_module.get_app()
             apps[test_name] = test_app
+            test_app.table.benchmark_key = f"{category_name}/{test_name}"
```
ElliottKasoar: Currently, test names have to be unique, so I'm not sure we need both for a key, or is it useful?

ml_peg/calcs/bulk_crystal/iron_properties/calc_iron_properties.py
```diff
@@ -10,7 +10,7 @@
 - Generalized stacking fault energy curves (110, 112)
 - Traction-separation curves (100, 110)
 
-This benchmark is computationally expensive and marked with @pytest.mark.slow.
+This benchmark is computationally expensive and marked with @pytest.mark.medium.
```
ElliottKasoar: We can probably just drop this since it doesn't say anything beyond what the code says

tests/test_speed_markers.py
```diff
@@ -0,0 +1,621 @@
...
+def test_single_marker_is_returned(tmp_path):
+    """A benchmark with one speed marker reports that marker."""
+    calc_dir = write_calc(
+        tmp_path,
+        "import pytest\n\n@pytest.mark.slow\ndef test_a():\n    pass\n",
+    )
+    assert get_benchmark_speed(calc_dir) == "slow"
+
+
+def test_multiple_markers_return_the_slowest(tmp_path):
```
ElliottKasoar: Do we actually want to allow what are essentially conflicting markers? I'd be inclined to make them mutually exclusive, and then test that having (any pair of) two raises an error.

tests/test_speed_markers.py
```diff
@@ -0,0 +1,621 @@
...
+        "@pytest.mark.slow\ndef test_a():\n    pass\n\n"
+        "@pytest.mark.very_slow\ndef test_b():\n    pass\n\n"
+        "@pytest.mark.multi_day\ndef test_c():\n    pass\n",
+    )
+    assert get_benchmark_speed(calc_dir) == "multi_day"
+
+
+def test_very_slow_is_not_matched_as_slow(tmp_path):
+    """very_slow is matched exactly, not as a substring of slow."""
+    calc_dir = write_calc(
```
ElliottKasoar, line 95: Why not just have the marker(s) as the input to `write_calc` since the rest is always necessary?

Probably also more readable to do a multi-line string (`"""`)?

docs/source/user_guide/benchmarks/bulk_crystal.rst
```diff
@@ -106,7 +106,7 @@ Symmetry checks are applied to components on the diagonal with a relative tolera
 Computational cost
 ------------------
 
-High: tests are likely to take hours-days to run on GPU.
+Slow: around 10 hours per model on an A100 GPU.
```
ElliottKasoar: I guess this is borderline with very slow? Is it slightly under 10 hours, or you just rounded down?

joehart2001: this is actually more like 12 ish hours now with frasers additional metric, will update

conftest.py
joehart2001: agreed

conftest.py
joehart2001: agreed

conftest.py
joehart2001: good idea

tests/test_speed_markers.py
joehart2001: agreed

tests/test_speed_markers.py
joehart2001: partially: we want to reject multiple speed markers on the same test, but allow separate tests within one benchmark to have different speeds e.g. phonons the reference calc and actual calc have differnet speeds. so will update

conftest.py
joehart2001: removed

docs/source/user_guide/benchmarks/bulk_crystal.rst
joehart2001: agreed

ml_peg/app/build_app.py
joehart2001: yep agreed

PR #797, opened 2026-08-05
Add GRACE

ElliottKasoar:

## Summary

Updates GRACE in models.yml to use the foundational model interface, allowing models to be downloaded by name, and updates the GRACE dependency.

One problem seems to be that based on their [docs](https://gracemaker.readthedocs.io/en/latest/gracemaker/foundation/), float32 is the default and float64 seems to only be accessible via different model downloads. I've modified the get_calculator function to try to deal with this, which seems to work from testing the two models I've added (one of which doesn't have a fp64 variant).
 
## Model Reference
https://github.com/ICAMS/grace-tensorpotential

## Progress

- [x] Model configuration
- [x] Documentation

## Testing

Tested against X23


ml_peg/models/models.py
```diff
@@ -474,3 +475,46 @@ def get_calculator(self, **kwargs) -> Calculator:
...
+        Calculator
+            Loaded GRACE ASE calculator.
+        """
+        from tensorpotential.calculator.foundation_models import MODELS_NAME_LIST
+
+        precision_map = {"low": "", "high": "-fp64"}
+        suffix = precision_map[precision]
+
+        if self.default_dtype is not None:
+            suffix = self.default_dtype
```
joehart2001, line 509: looks good apart from this. i think thsi would result in e.g. float32 rather than fp32 due to the precision map in the generic ase calc.

ElliottKasoar: I may be missing something but why? `default_dtype` defaults to `None` for `GenericASECalc`, and the `precision_map` is only defined within the `get_calculator` functions, which GRACE has its own implementation of.

I've admittedly forgotten all of the discussions about how `default_dtype` works, but I believe the intent is it's only usually set by an `overwrite_dtype`, which comes from the model-specific part of the yaml (I just fixed passing this to GRACE).

joehart2001:
@ElliottKasoar i think this is ok, happy to merge

joehart2001 (approved):

ElliottKasoar:
> @ElliottKasoar i think this is ok, happy to merge

Thanks @joehart2001! You're happy with the `default_dtype` bit then?

joehart2001:
> > @ElliottKasoar i think this is ok, happy to merge
> 
> Thanks @joehart2001! You're happy with the `default_dtype` bit then?

yep

PR #807, opened 2026-08-10
Frontend overhaul: responsive redesign, dark mode, settings, faster framework pages

DhruvSkyy:

## Summary

This redesigns the app frontend: it's now responsive to screen sizes, has dark mode, a settings panel (theme, colour scheme, table zoom, expand all, clear cache), and click-to-pin table tooltips. The overall design is cleaner, with card-based pages, a sticky sidebar, and the model-name column staying pinned when wide tables scroll. Preferences like theme and zoom persist between visits. Framework pages are also much faster, since benchmark cards now only render when expanded:

| Framework page | main | this PR |
|---|---|---|
| mace-polar-1 | ~62 s | ~4 s |
| mace-multihead | ~11 s | ~3.5 s |

No model changes, models.yml matches main. I also wrote 49 Playwright UI tests (smoke, interactions, responsive, settings and theme) and a workflow to run them in CI.

## Testing

pre-commit, the full pytest suite (59 passed, including the 49 UI tests) and the docs build all pass locally. I also manually tested the app throughout (different screen sizes, dark/light, compared against the live site).

## AI disclosure

Significant AI assistance (Claude Code) was used for the implementation and tests.


joehart2001:
Thanks for this @DhruvSkyy! Will take a look soon. In the meantime could you share some screenshots of the site?

ml_peg/app/utils/load.py
```diff
@@ -25,6 +24,24 @@
 from ml_peg.models.get_models import get_model_names, load_model_configs
 
 
+def _read_json(path: str | Path) -> object:
+    """
+    Read a JSON file quickly via orjson (faster than the stdlib parser).
```
joehart2001, line 29: we need a fallback to stdlib for this as the analysis stage sometimes writes NaN values which orjson breaks on

ml_peg/app/data/fonts/inter-variable.woff2
joehart2001, line 1: maybe could have some options for this?

joehart2001:
i think its looking nice and much faster which is amazing.

critical issues:
- when a plot is displayed and i click on a datapoint to view the structure, the structure vis replaces the plot instead of being shown below
     - this seems to effect most tables, but doenst effect e.g. elasicity tensor MAE (but does effect bulk and shear moduli)
- when adjusting a weight on a benchmark page, it updates the table nicely and propagates to the summary table. however, when switching back to the benchmark page, the score is reset to the original value.
     - i seemed to fix this by changing to `prevent_initial_call="initial_duplicate",` which is what we used originally. theres a slight cost penalty to this, but otherwise theres a race condition between the original table and the updated table

other things:
- maybe lets extend N tables by default instead of collapsing all. i think this could be a sweet spot as you do want to see some data at a glace instead of having to click for all of it
- i think its quite important to show benchmark weights and threholds by default too. also for me the coloured threhsold box seems to be cropped to the RHS of it
- try to make sure the pre-commit passes
- table zoom seems to cut off col titles

joehart2001:
Hey @DhruvSkyy, i think we should actually break this PR down into separate PRs: things that dont touch the callbacks and things that do, as the parts which touch the callbacks needs a lot of testing and will delay getting in the rest of the changes. 

e.g. could split and make a PR for the cards and collapsable benchmarks? basically bite sized chunks are best so we can isolate each change and test it properly.

joehart2001:
> Hey @DhruvSkyy, i think we should actually break this PR down into separate PRs: things that dont touch the callbacks and things that do, as the parts which touch the callbacks needs a lot of testing and will delay getting in the rest of the changes.
> 
> e.g. could split and make a PR for the cards and collapsable benchmarks? basically bite sized chunks are best so we can isolate each change and test it properly.

actually lets keep the collapsed tests how they are for now

DhruvSkyy:
Split at the callback boundary as suggested, so closing this in favour of #820 (no callback changes: theming, settings, responsive layout) and #821 (the callback half, stacked on #820 and a draft until it merges: lazy-mount cards, faster framework pages, and the fixes from the review). Both are green on pre-commit and the full suites.

PR #777, opened 2026-07-31
clear cache button rename -> hard reset

joehart2001:
Please click the `Preview` tab and select a PR template:

- [New benchmark for existing models](?expand=1&template=new_benchmark.md&labels=new+benchmark)
- [Benchmark a new model](?expand=1&template=new_model.md&labels=new+model)
- [All other proposed changes](?expand=1&template=custom.md)

PR #820, opened 2026-08-19
Frontend part 1: theming, settings and responsive shell

DhruvSkyy:

- [x] I have added human-written tests for the new logic. <sub>(AI-drafted; I reviewed and ran every test, and manually tested the app.)</sub>

## Summary

First half of #807, split so that this part touches no callbacks. Cards, collapsible benchmarks and the callback fixes follow in a second PR, stacked on this one.

- Dark mode with a persisted preference, a `--mlpeg-*` token layer, and theme-aware Plotly figures.
- Settings popover: theme, font, colour scheme, table zoom, expand-all, clear cache. Preferences persist between visits.
- Responsive layout from phone to ultrawide, with a sticky sidebar and an off-canvas mobile nav.
- Onboarding tidy-up and click-to-pin table tooltips.
- New `ui-tests` CI job: 25 Playwright tests against the real app.

Two fixes reported on #807:

- The coloured threshold boxes were cropped on the right. Dash 4 puts the component `style` (our width) on a `.dash-input-container` wrapper that hides overflow, while the inner `<input>` that carries the colour keeps its wider intrinsic width. The field now fills its wrapper; the regression test is in part 2, where the scoring grid exists.
- `BaseApp.load_info` warned that `info.json` was missing and then opened it anyway. The published data tarball has none, so every benchmark failed to load under the new CI job.

Notes:

- `ui-tests` is scoped with a `paths:` filter so unrelated PRs skip it, and reports coverage under its own Coveralls flag, since `ci.yml` now skips `tests/app`. No test data is committed: the suite runs against a ~800 KB fixture published once to the bucket as `tests/ui-fixture-v1.zip`, which `conftest` downloads, checks against a pinned SHA-256 and caches (`actions/cache` keeps the bucket off the critical path in CI). The key is never overwritten — a changed fixture is a `-v2` object plus a constant bump; recipe in the developer guide. (The pinned digest is of the object as uploaded, `b0fab469…` — it differs from the `518b33…` quoted in the comments because the archive was re-zipped on upload, adding `__MACOSX/` entries; contents verified identical, and extraction skips those entries.)
- The tarball restores its own `onboarding/sidebar_hover.css`, and Dash loads root assets before subfolder ones, so the theme rule is scoped to `#sidebar-nav` and `sidebar_hover.css` is left as it is on main.
- `app/data/data.tar.gz` is from 2026-05-29 and contains no `info.json`, so element filtering is inert in CI. Could someone with upload rights re-publish it?

There is also a pass over the new code here: preference persistence, dark-mode contrast, keyboard focus and the data handling in the new CI job.

No model changes; `models.yml` matches main.

## Linked issue

Resolves #

## Testing

pre-commit, the full pytest suite, the 25 UI tests and the docs build pass locally, against both a local data directory and a fresh extract of the published `app/data/data.tar.gz`.

## AI disclosure

Significant AI assistance (Claude Code) was used for the implementation and tests.

## Screenshots

| Summary page | Settings |
|---|---|
|  |  |


joehart2001:
hey @DhruvSkyy thanks for this!

a few things:
- i think we should add outlines to tables? its not too obvious in the light version
- light version set as default
- the loading wheel stays on 0% for most of the time and then jumps to 92%

- we seem to have an additional box here. i think we have the square outlies on the live site but not the inner box too. could just hide this:
Screenshot 2026-08-19 at 15 01 31

- in your last version you had vertical dividers between the weight boxes which separated out the cols which i think was nice, would be good to add back

- also apply dark theme to weight and threshold area:
Screenshot 2026-08-19 at 15 01 57

- the top bar with tutorial etc seems to be a set width, would be good to maintain it whilst horizontally scrolling
Screenshot 2026-08-19 at 14 59 44

.github/workflows/ui-tests.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      - name: Install dependencies
+        run: uv sync
+
+      - name: Install Playwright browser
+        run: uv run playwright install --with-deps chromium
+
+      - name: Download and unpack app data
+        run: |
+          uv run ml_peg download --key app/data/data.tar.gz --filename data.tar.gz
```
ElliottKasoar: Downloading and unpacking 2GB isn't insignificant, but more importantly, it's going to keep growing (in fact, I wouldn't be surprised if the data that's actually live is several factors larger, as the S3 version is a bit out of date for various reasons)

I don't think we want to use the real data, or certainly not all of it, for tests.

ml_peg/app/run_app.py
```diff
@@ -48,6 +49,54 @@ def _build_full_app(app: Dash, category: str, test: str):
...
+        # apply on phones (without this, mobile browsers assume a ~980px page).
+        {"name": "viewport", "content": "width=device-width, initial-scale=1"},
+    ],
+)
+
+# Apply the saved theme before first paint so a dark-mode reload doesn't flash
+# white. Reads the persisted dcc.Store value straight from localStorage — Dash
+# stores local data under the component id ("theme-store", JSON-encoded; see
+# ml_peg.app.utils.settings). Falls back to the OS preference, then light.
+app.index_string = app.index_string.replace(
```
ElliottKasoar, line 64: This is something we probably should import - it's not fundamental to controlling running the app

.gitignore
```diff
@@ -41,4 +41,14 @@ __pycache__/
 ml_peg/app/data/*
 !ml_peg/app/data/onboarding/
 !ml_peg/app/data/table_download/
+!ml_peg/app/data/theme.css
+!ml_peg/app/data/theme_plots.js
+!ml_peg/app/data/settings_popover.js
+!ml_peg/app/data/tooltip_portal.js
```
ElliottKasoar: These could probably be in their own directory to simplify this, but even if not, if these are already git tracked, they don't need to be excluded from the ignore?

ElliottKasoar:
Hi @DhruvSkyy, the live data is slightly messy at the moment. I'd hoped to tidy things up a bit before doing an upload, but I don't think I'll be able to this week, so I've just pushed the current version, which should include all the info.json files.

As I mentioned above, I don't think this is sensible to download as part of the CI though.

joehart2001:
alos we changed clear cache to hard reset recently, maybe scan over recent commits to check there isnt anything else like this

DhruvSkyy:
Thanks both — all of this is addressed.

Four of the UI points were already fixed in the second PR, which is why they looked broken here: the CSS landed in this half but the markup that wears `mlpeg-scoring-grid` and `mlpeg-table-scroll` was in the other, so those rules were dead selectors on this branch alone. Rather than leave that until #821, the classNames and tokens are backported, so this PR now stands on its own. That covers the inner box on the weight inputs, the column dividers, the dark weights/thresholds panel, and the top bar during horizontal scroll — wide tables now scroll inside their card, so the page never scrolls sideways at all.

The other three:

- Tables are outlined. `--mlpeg-border` was `#dcdcdc` on white, about 1.3:1, and the table had no perimeter at all. New `--mlpeg-table-border` / `--mlpeg-table-outline` tokens drive the cell rules, the scoring-grid dividers and a new outer border, so the card and top-bar borders keep their lighter weight.
- Light is now the unconditional default. The pre-paint script fell back to `prefers-color-scheme`, so the app looked different depending on the visitor's OS.
- The loading ring was a real bug. `--mlpeg-pct` was a registered custom property animated by CSS into a conic-gradient and a counter; both need a repaint every frame, so the animation ran on the main thread, which is blocked solid while Dash hydrates. No intermediate frame ever painted, so you saw 0% and then the finished 92%. It is driven by `requestAnimationFrame` now, capping how much a single frame may contribute so a blocked stretch cannot produce a jump. That also fixes browsers without `@property` support, which snapped straight from 0 to 92, and `prefers-reduced-motion`, which pinned it at 92 from the first frame.

On the CI data — agreed, and it is gone. The suite only ever built a couple of benchmarks, so the download was almost entirely waste. `tests/data/app` now holds a 2.4 MB fixture covering four benchmarks across three categories, chosen for coverage rather than size: a parity plot with a structure viewer, a twelve-metric table with dropdown-driven figures, a physicality benchmark (which is also what makes `/framework/mace-multihead` exist), and one with experimental levels of theory. `conftest` copies it in only where real data is absent, so a local checkout with the real tree is unaffected. The job runs in about 30 seconds instead of five and a half minutes, and it now tests a fixed input rather than whatever was last published.

Building with `category="*"` means the fixture directory is the single place coverage is declared — adding a benchmark to `tests/data/app` extends the suite with no code change — and the session asserts none of the declared benchmarks silently failed to load.

Two things that fell out of this worth flagging:

- `surfaces/graphene_wetting_under_strain` is skipped even with the full download. It opens `ml_peg/calcs/surfaces/graphene_wetting_under_strain/outputs/database_info.yml` at import time and that file is in neither the repo nor the tarball.
- The next tarball upload should drop the hand-authored CSS/JS it currently ships (`theme.css`, `loading.css`, `settings_popover.js`, `theme_plots.js`, `tooltip_portal.js`, `utils/`, `table_download/`, `onboarding/`, `icons/`, `fonts/`). They live under `data/ui/` now, so an old tarball no longer overwrites them, but Dash would register both copies. The scripts guard against running twice in the meantime.

The head scripts move to `app/utils/head_scripts.py`.

On the `.gitignore`: you were right, and it was worse than it looked. `!ml_peg/app/data/icons/` never worked — a trailing-slash negation only lets git descend into the directory, and the files inside still matched the global `*.svg`, so those icons were tracked only because they had been force-added. Five more tracked files had no exception at all. Everything hand-authored now lives in `ml_peg/app/data/ui/`, so it is one exception, the `*.json`/`*.svg` interaction stops mattering, and the tarball unpacks alongside tracked files instead of over them — which also removes the `git checkout -- ml_peg/app/data` step the job needed.

Re-enabled the page-overflow and sticky-column assertions in `test_responsive` that were parked waiting for part 2. 26 UI tests here, 55 on the stacked branch, both green against the fixture alone.

#821 is rebased on this and stays in draft until this merges.

DhruvSkyy:
Thanks @joehart2001! All addressed in the latest push:

- **Table outlines**: tables now get a visible perimeter outline plus heavier inter-cell borders in light mode (dedicated tokens, so dark mode keeps its own values).
- **Light default**: light is the default on first visit; dark is strictly opt-in via the settings toggle.
- **Loading wheel**: root-caused — the percentage was animated purely in CSS, and Dash hydration blocks the main thread, so no intermediate frame ever painted (hence 0% → 92%). It's now driven by requestAnimationFrame, advances only on frames that actually paint, and snaps to 100% when the app is ready.
- **Extra inner box**: gone — the Dash input wrapper was drawing its own border inside the outer one; the inner input keeps a single box.
- **Vertical dividers** between the weight/threshold columns are restored (and line up with the table columns above).
- **Weights & thresholds panel** now follows dark mode (it had hardcoded light colours), and the "Show normalised scores" label got an explicit themed colour.
- **Top bar**: fixed by removing page-level horizontal scroll entirely — wide tables scroll inside their own container, so the bar always spans the page. There's a browser test asserting no page-level horizontal overflow at several viewports.
- **Recent-commit drift**: scanned main since the branch point — the button is "Hard Reset" per #777, and I didn't find other renames/regressions.

While re-checking dark mode I also fixed a few things the toggle path was masking:

- Figures mounted after first paint (clicking a table cell, or any callback that swaps a figure) never picked up the theme until you happened to toggle it — plotly adds its class after Dash inserts the wrapper, so the re-theming observer never fired. Fixed, with a regression test that loads dark and mounts a figure without touching the toggle.
- The `--Dash-*` token overrides (checklist/dropdown label colours) sat in a selector that ties with Dash's runtime-injected rule and can lose on injection order — raised so they always win; that's what made "Show normalised scores" nearly invisible in your screenshot.
- FAQ and benchmark-title headings had hardcoded black; the 🔗 link column is now a masked SVG icon that follows the theme (a colour-emoji glyph ignores CSS `color` entirely); and re-weighting the summary table no longer flips the link cells back to a white background in dark mode.

.github/workflows/ui-tests.yml
DhruvSkyy: Agreed — the job no longer downloads anything. The UI tests now run against a small committed fixture (tests/data/app, ~2.4 MB, four benchmarks picked for coverage per KB), so CI is independent of the published data and stays fixed as the real data grows.

ml_peg/app/run_app.py
DhruvSkyy: Moved out — the script is now built by `ml_peg/app/utils/head_scripts.py` and imported into run_app.

.gitignore
DhruvSkyy: The hand-authored assets now live in their own directory (`ml_peg/app/data/ui/`), but the exceptions are still needed: `data/*` has to stay ignored for the generated benchmark tree that unpacks there, and without the un-ignore lines git would never pick up *new* files added under `ui/` (the directory-level rule blocks descent). Added a comment in the file explaining this — happy to restructure if you'd prefer a different split.

joehart2001:
@DhruvSkyy  i think lets upload the data you need to the s3 bucket to keep it clean

DhruvSkyy:
Hi, I think a few of these comments were already resolved in the second PR. Some features didn't split cleanly, but this PR should now work independently.

The second PR is stacked on this one, so its diff still includes both commits. I think the comments are addressed, with just the S3 point still open.

I can't upload it myself because ml_peg upload requires credentials. Could you publish [ui-fixture-v1.zip](https://github.com/user-attachments/files/31336360/ui-fixture-v1.zip) (654 KB)? 

```bash
ml_peg upload --key tests/ui-fixture-v1.zip --filename ui-fixture-v1.zip --credentials credentials.json
```

Expected SHA256: `518b3313476aff69ae2be1e0af4860cde8fac247d3d5c4ff2980e22c11aaa37f`

joehart2001:
> Hi, I think a few of these comments were already resolved in the second PR. Some features didn't split cleanly, but this PR should now work independently.
> 
> The second PR is stacked on this one, so its diff still includes both commits. I think the comments are addressed, with just the S3 point still open.
> 
> I can't upload it myself because ml_peg upload requires credentials. Could you publish [ui-fixture-v1.zip](https://github.com/user-attachments/files/31336360/ui-fixture-v1.zip) (654 KB)?
> 
> ```shell
> ml_peg upload --key tests/ui-fixture-v1.zip --filename ui-fixture-v1.zip --credentials credentials.json
> ```
> 
> Expected SHA256: `518b3313476aff69ae2be1e0af4860cde8fac247d3d5c4ff2980e22c11aaa37f`

thanks @DhruvSkyy ive uplaoded it, is it working for you?

DhruvSkyy:
hi @joehart2001, it should work now.

ml_peg/calcs/bulk_crystal/high_pressure_relaxation/data/P000.json.bz2
joehart2001, line 1: shouldnt need any of these

.gitignore
```diff
@@ -38,7 +38,20 @@ build/
 dist/
 pip-wheel-metadata/
 __pycache__/
+# ml_peg/app/data is the Dash assets folder, so it holds two unrelated things:
+# the generated benchmark tree (never committed) and the hand-authored front-end
+# assets. The latter all live in ui/, so a single exception covers them and the
+# data tarball unpacks alongside rather than on top of tracked files.
+# The `**` line matters: un-ignoring the directory only lets git descend into it,
+# and the files inside would still match the *.json/*.svg/*.csv rules above.
```
joehart2001, line 46: probably dont need all these comments

joehart2001:
@DhruvSkyy thanks this is looking great. attatched below is the only bug ive found: when the page is resized after we have horizontally scrolled with the table, the table disappears


https://github.com/user-attachments/assets/7a679c11-02ca-4f63-a700-faa8e9524b15

joehart2001:
> @DhruvSkyy thanks this is looking great. attatched below is the only bug ive found: when the page is resized after we have horizontally scrolled with the table, the table disappears
> 
>  stretch-bug.mov

btw can you resolve commnets if they have been addressed.

otherwise im happy to merge. will get elliott to take a scan over when hes back from holiday (soon)

joehart2001:
@DhruvSkyy ive managed to fix this, see the commit

.github/workflows/ci.yml
```diff
@@ -26,7 +26,10 @@ jobs:
         env:
           # show timings of tests
           PYTEST_ADDOPTS: "--durations=0"
-        run: uv run pytest --cov ml_peg --cov-append . --cov-report xml
+        # tests/app is the browser suite, run separately by ui-tests.yml (which
+        # installs Chromium and fetches the ~800 KB test fixture from the
+        # bucket); skip it here.
```
ElliottKasoar, line 31:
```suggestion
        # tests/app is the browser suite, run separately by ui-tests.yml
```
This is enough...

.github/workflows/ui-tests.yml
```diff
@@ -0,0 +1,81 @@
+name: ui-tests
+
+# Browser-driven tests for the Dash app (tests/app). These boot the real app on
+# an ephemeral port and drive it with Playwright, so they need a headless
+# Chromium. The benchmark data is a small fixture (~800 KB) that conftest fetches
+# from the ml-peg-data bucket, checks against a pinned digest and caches under
+# ~/.cache/ml_peg; the full app data (~9 GB) is never downloaded.
```
ElliottKasoar, line 7: Maybe we want slightly more generic sizes of data? Neither of these values are likely to be true for long, but I imagine the test data will remain ~MB, and the full data will be at least ~10GB

ml_peg/app/utils/weight_presets.yml
```diff
@@ -6,7 +6,7 @@ Materials:
   - Surfaces
   - NEBs
   - Physicality
-  - Defect
+  - Defects
```
ElliottKasoar, line 9: This needs fixing but isn't related to this PR. I know it's trivial but can you do a new PR for this (or I can) and remove it from this?

It makes it much easier to keep track of changes etc.

ml_peg/calcs/bulk_crystal/high_pressure_relaxation/data/P025.json.bz2
ElliottKasoar, line 1: As above, please remove these files

PR #835, opened 2026-08-31
feat: install mlipaudit 0.1.6 from PyPI and fix resolution conflict

lwalew:

## Summary

Pins `mlipaudit` to `0.1.6`, installed from **PyPI** instead of a git source (the release is now
published), so `[tool.uv.sources]` loses its entry and installs no longer clone the repo.

Two follow-on dependency fixes were needed to make this resolve:

- **`requires-python` capped at `<3.14`.** mlipaudit 0.1.6 pulls `mlip 0.2.4 -> e3j`, which pins
  `pandas<2.3`, while `janus-core -> codecarbon` requires `pandas>=2.3.3` on 3.14. Marker-gating
  the extra is not enough — `e3j`'s cap has no Python marker, so uv still explores the 3.14 split.
  Excluding 3.14 to be able to resolve the dependencies. This matches the classifiers and the 3.10-3.13 CI matrix; it can be reverted once `e3j` relaxes its cap.
- **`jax`/`jaxlib` `<0.11` cap dropped.** mlip 0.2.4 no longer pulls `dm-haiku`, which is what
  needed it. 

Also repoints the MLIP Audit logo in `frameworks.yml` at the `v0.1.6` tag so it survives deletion
of the `mlpeg-migration` branch.

⚠️ Side effect: `e3j`'s cap is not confined to the extra, so **`pandas` moves 2.3.3 -> 2.2.3 for the
whole resolution**, including a bare `uv sync`.

## Testing

Successfully ran `mlipaudit` benchmarks.


lwalew:
Upstream bug is that `mlipaudit` depends on `mlip` which depends on `e3j` which declares `pandas~=2.2.2`. This `pandas` dependency will be moved to `dev` in a future release. Once this is done, we should be able to relax `requires-python = ">=3.10,<3.14" -> ">=3.10"`, although in its current state this would reflex the python versions advertised in the README and the ci.

pyproject.toml
```diff
@@ -9,7 +9,7 @@ authors = [
     { name = "Alin M. Elena" },
     { name = "Gábor Csányi" },
 ]
-requires-python = ">=3.10"
+requires-python = ">=3.10,<3.14"
```
joehart2001: im unsure this global exclusion is needed. we can just exclude for mlip audit, see code suggestion below

pyproject.toml
```diff
@@ -66,11 +66,7 @@ mattersim = [
     "mattersim==1.2.4; python_version >= '3.12'",
 ]
 mlipaudit = [
-    "mlipaudit; python_version >= '3.11'",
-    # dm-haiku 0.0.16 (pulled via mlip -> e3nn-jax -> haiku) still uses
-    # jax.core.DropVar, removed in JAX 0.11. Cap jax below 0.11.
-    "jax<0.11; python_version >= '3.11'",
-    "jaxlib<0.11; python_version >= '3.11'",
+    "mlipaudit == 0.1.6; python_version >= '3.11'",
```
joehart2001:
```suggestion
    "mlipaudit == 0.1.6; python_version >= '3.11' and python_version < '3.14'",
```

uv.lock
joehart2001, line 1: Could you regenerate uv.lock without --upgrade to avoid the unrelated dependency updates?

lwalew:
@joehart2001 I've applied your suggested changes

ElliottKasoar (approved):
Looks good to me, thanks!

PR #843, opened 2026-09-08
Hide models with no data

ElliottKasoar:

## Summary

Hide missing models from tables. Needs more testing. I'm also seeing a weird bug where NaN values aren't stopping the total score being calculated, but I'm not sure if this is introduced here or not...

## Linked issue

Resolves #

### DIFF
--- a/ml_peg/app/utils/load.py
+++ b/ml_peg/app/utils/load.py
@@ -18,6 +18,7 @@ from ml_peg.app.utils.utils import (
     clean_table_data,
     clean_thresholds,
     clean_weights,
+    drop_empty_model_rows,
     is_numeric_column,
     none_to_nan,
     sig_fig_format,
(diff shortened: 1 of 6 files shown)


joehart2001:
things me and claude found + fixes ive made:
 - Uncomputed and failed results were treated alike -> filter uncomputed rows before converting nulls to NaNs
 - Partial data could receive a score -> preserve missing benchmarks as NaNs
 - Zero-weight missing benchmarks blocked valid scores -> ignore them during scoring
 - Fully empty tables could crash -> safely handle empty table styling

joehart2001:
scenario: model A only has data for 1 test
- this means the model does not have a row in the summary table
- but if we put all the weights to zero except the 1 test for that category, then the model will have a category score and should appear in the summary table

so we need this to be part of a callback

ml_peg/app/utils/load.py
```diff
@@ -55,22 +55,29 @@ def rebuild_table(
         table_json = json.load(f)
 
     data = table_json["data"]
+    columns = table_json["columns"]
+    metric_columns = [
+        column["id"]
+        for column in columns
+        if column.get("id") not in {"MLIP", "Score", "id", "link"}
+    ]
+    # A row containing only null metrics was never computed. Keep explicit NaNs,
```
ElliottKasoar: @joehart2001 unless I'm misunderstanding, this isn't true. Currently the analysis has no way to distinguish between data that's missing because of failed calculations, or models we haven't tried yet, and this won't be possible in a completely consistent way until we have better logging (hence the `none_to_nan` below).

I'm not sure I understand why we'd want to do this rather than just filter these tables by whether there are any `NaN`s after `none_to_nan(data)` is run, since if a benchmark table has no metric results, that will never change even with weight changes

joehart2001: ok yep maybe this was a bit forward looking, but without the proper infrastructure to do it properly. the last sentence convinced me

joehart2001:
think this is ok?

ElliottKasoar:
If you've tested and you're happy, can you approve, @joehart2001? I don't think it includes any of the changes you'd originally pushed

joehart2001:
> If you've tested and you're happy, can you approve, @joehart2001? I don't think it includes any of the changes you'd originally pushed

no i think youve fixed it, will approve

joehart2001 (approved):

PR #855, opened 2026-09-14
Add carbon benchmarking from GAP-20 paper (Part 1)

patrickwrowe:
A carbon category porting the elemental-carbon test suite that validated GAP-20 (Rowe et al., J. Chem. Phys. 153, 034702, 2020) this PR includes three essential benchmarks, all scored against archived optB88-vdW VASP data from that work. 

Lattice parameters and bond lengths:
Assesses models on lattice constants, fullerene bond lengths, and bond lengths for different nanotube indices. 

Formation Energies:
Models are assessed on their prediction of the formation energies of 7 key carbon allotropes relative to their graphite reference.

Surface Energies: 
Assessment of surface energies for key surfaces.

Reference data is published separately at patrickwrowe/Carbon_GAP under CC BY 4.0 and fetched with the existing download_github_data helper. The reference data used in this PR are the exact reference calculations used in the original publication, but the code has been updated and modernized to python 3 and to fit with the current ML-PEG project conventions.

Thresholds for good/bad were tuned for this work based on the performance of 5 MACE models + the reported ref values for GAP-20 from the paper (not recomputed here). 

One or two intentional changes are made here versus the paper which are all for the better. Lonsdaelite now checked against both an and c lattice parameters. Nanotube strain energy now reported relative to flat graphene sheet instead of the isolated atom, matching both the science and the formation energies relative to graphite better.

Some tests are intentionally missing entirely, this PR was getting too big, so I'll follow up with more when this work merges.

---

AI use disclosure: The original tests and associated VASP-calculated reference data were all hand-written/performed without any AI assistance (pre- 2020). The process of updating this code and porting it to python 3 and the current project conventions were AI assisted. Values computed in this PR have been validated to match those in the existing publication.


patrickwrowe:
Hi @ElliottKasoar and @joehart2001 - I can't request reviewers on this repo but you'd probably be good candidates. 

This is the first of a couple of carbon-focused benchmarks I'll try to contribute in the coming weeks (when I can find the time! This took me a while to get into a good state). 

Some more to come for pure carbon, and hopefully some more exploratory C,H,O / combustion related work following a conversation with Gábor a little while ago.

docs/source/user_guide/benchmarks/carbon.rst
```diff
@@ -0,0 +1,165 @@
...
+
+computed identically for model and reference. Graphite is 0 by construction and
+excluded; the other six systems contribute. Reference values span 72-450 meV/atom, so
+a percentage error would be dominated by the smallest denominator and this is
+reported as a mean absolute error in meV/atom.
+
+Computational cost
+------------------
+
+Medium: relaxations of up to 288 atoms, per model variant.
```
joehart2001, line 66: lets add a gpu estimate for e.g. mace-mp-0a?

ml_peg/analysis/utils/utils.py
```diff
@@ -166,6 +170,27 @@ def mae(ref: list, prediction: list) -> float:
     return mean_absolute_error(ref, prediction)
 
 
+def mape(ref: list, prediction: list) -> float:
```
joehart2001, line 173: we are planning on adding a feature where in the app you can select e.g. rmse, mae, male etc so so far we have been doing all as mae to be consistent. however, if this is for specific reproduction of your paper then we can leave this for now. just for your info

ml_peg/app/carbon/carbon.yml
```diff
@@ -0,0 +1,3 @@
+title: Carbon
+description: Structural and energetic properties of carbon allotropes from the GAP-20 test suite
```
joehart2001, line 2: idea: take a look at the `frameworks.yml` and tag your tests with the GAP-20 paper like e.g. for the multihead paper. then we could make this category carbon more generally instead of just gap-20 tests and you can have your own gap-20 page in addition. see the live site for this (btw mace polar tab will almost crash your browser (we are working on it!) so have a look at the mutlihead tab and tags for an example)

joehart2001:
Hey @patrickwrowe, thank you for this, its looking great! i will take a more thorough look soon. excited to have these tests in!

ml_peg/app/carbon/carbon.yml
patrickwrowe: Nice - thanks @joehart2001 - I've incorporated this.

ml_peg/analysis/utils/utils.py
patrickwrowe: Awesome - I would say it makes sense eventually to merge this with the project's conventions, but maybe that can wait for a follow-up if that's not a feature that's added yet?

docs/source/user_guide/benchmarks/carbon.rst
patrickwrowe: Sure, how carefully should this estimate be made? 

One thing to note here is that some potentials with non-conservative forces (e.g. ORB) actually don't converge with the current settings. So the cost is one element, but in some sense the potential also just "fails" the test. What's the convention for this in the project at the moment? For the time-being I just let them hit max-iterations.

PR #856, opened 2026-09-15
add dpa3 and 4 models and model download

joehart2001:

## Summary

Adds DPA models:
- dpa-3p3-1M-omat
- dpa-4-nano-omat
- dpa-4-neo-omat
- dpa-4-plus-omat

note: dpa checkpoints are fixed at float32

## Linked issue

Resolves #


docs/source/user_guide/models.rst
```diff
@@ -204,6 +204,94 @@ mace-polar-1-l
...
+
+.. code-block:: yaml
+
+   dpa-3p3-1M-omat:
+     module: deepmd.calculator
+     class_name: DP
+     datasets: [OpenLAM-v1, OMAT, MPtrj, OC20, OC22, ODAC23, SPICE2]
+     trained_on_dispersion: false
+     level_of_theory: PBE
+     overwrite_dtype: float32
```
ElliottKasoar: `overwrite_dtype` isn't something that should be in the models.yml file (or here)

pyproject.toml
```diff
@@ -9,7 +9,7 @@ authors = [
     { name = "Alin M. Elena" },
     { name = "Gábor Csányi" },
 ]
-requires-python = ">=3.10"
+requires-python = ">=3.10,<3.14"
```
ElliottKasoar, line 12: I think this is already changed since #350 was merged

docs/source/developer_guide/add_models.rst
```diff
@@ -99,6 +99,12 @@ The common fields are:
    benchmarks request either ``precision="high"`` or ``precision="low"``; this
    field forces a specific dtype regardless of that request.
 
+``download``
```
ElliottKasoar: Two questions regarding this:

If they've chosen not to add certain models to their [registry](https://github.com/deepmodeling/deepmd-kit/blob/master/deepmd/pretrained/registry.py#L8), might it not be better to either suggest this be updated or wait? 

If we do still choose to download models that aren't added to the registery, we may still want to use more of their [download](https://github.com/deepmodeling/deepmd-kit/blob/master/deepmd/pretrained/download.py) functions/cache, else you may end up storing duplicates if they are added later.

I'm open to discussion on either point, but think it's worth discussing

joehart2001: ive opened an issue and will remove the model and utility for now so it doenst slow down getting the other mdoels in

ElliottKasoar (commented):
From setting up DPA4 in janus, I think we need to update the dependencies a little - it has various dependencies that aren't specified, including a version of e3nn that clashes with the version mace uses.

See: https://github.com/stfc/janus-core/pull/723

joehart2001:
> From setting up DPA4 in janus, I think we need to update the dependencies a little - it has various dependencies that aren't specified, including a version of e3nn that clashes with the version mace uses.
> 
> See: [stfc/janus-core#723](https://github.com/stfc/janus-core/pull/723)

@ElliottKasoar i think its looking good now: i have added the same `override-dependencies` as janus for windows

PR #859, opened 2026-09-16
Move conftest into source code

### LINKED ISSUE #316: conftest.py missing from pip installed package
### Summary

Since conftest.py lives in the top level of the repo, it's not included when you pip install `ml-peg`.

This means commands such as `ml_peg calc` fail, as our custom options are not included, and also means that our custom markers and `current_models` won't work.

We may be able to move `conftest.py` to `ml_peg`, but we'll need to ensure this is set up correctly.

### Platform

All

### Python version

_No response_

ElliottKasoar:

## Summary

Moves conftest.py into `ml_peg`, and some of it into `ml_peg/calcs`, which I believe means they'll be pip installable.

## Linked issue

Resolves #316

## Testing

Adds some (AI-assisted) tests, which are broken down into two parts. Tests of the ml_peg CLI via `runner.invoke`, and tests of how `pytest` handles the inputs that are parsed to it. These look sensible, but they're not fully end-to-end, as that would invoke globs and mutate various global parameters that make it hard to test.


ml_peg/conftest.py
```diff
@@ -65,16 +65,17 @@ def pytest_configure(config):
         models.models_file = model_file
 
 
-def pytest_collection_modifyitems(config, items):
-    """Skip slow tests and deselect tests outside the requested framework(s)."""
-    skip_slow = pytest.mark.skip(reason="need --run-slow option to run")
-    skip_very_slow = pytest.mark.skip(reason="need --run-very-slow option to run")
-    for item in items:
-        if "very_slow" in item.keywords and not config.getoption("--run-very-slow"):
```
joehart2001, line 73: is removing these then changing to use `-m run-very-slow`?

ElliottKasoar: They shouldn't actually be removed, just moved to the calculation-specific conftest.py, since they're not used by the analysis.

ElliottKasoar: That is something to consider though

joehart2001: ok cool, can do later anyway

joehart2001 (approved):

